PR #12, opened 2015-10-06
DISPATCH-178 - Added code to route proton trace messages to the dispa…

ganeshmurthy:
…tch router log file instead of console

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -64,6 +64,15 @@ static qd_thread_t *thread(qd_server_t *qd_server, int id)
 
 qd_error_t qd_entity_update_connection(qd_entity_t* entity, void *impl);
 
+/**
+ * This function is set as the pn_transport->tracer and is invoked when proton tries to write the log message to pn_transport->tracer
+ */
+static void qd_log_tracer(pn_transport_t *transport, const char *message)
+{
+    qd_connection_t *ctx = (qd_connection_t*) pn_transport_get_context(transport);
+    qd_log(ctx->server->log_source, QD_LOG_TRACE, "[%d]:%s", ctx->uid, message);
+}
+
 static qd_error_t connection_entity_update_host(qd_entity_t* entity, qd_connection_t *conn)
 {
     const qd_server_config_t *config;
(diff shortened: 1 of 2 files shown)


src/server.c
```diff
@@ -252,6 +262,8 @@ static void thread_process_listeners_LH(qd_server_t *qd_server)
         ctx->user_context = 0;
         ctx->link_context = 0;
         ctx->ufd          = 0;
+        ctx->uid          = qd_server->next_connection_id; // Increment the connection id so the next connection can use it
+        qd_server->next_connection_id++;
```
ted-ross: You can combine these into one statement:  ctx->uid = qd_server->next_connection_id++;

src/server.c
```diff
@@ -282,6 +294,16 @@ static void thread_process_listeners_LH(qd_server_t *qd_server)
         pn_transport_set_max_frame(tport, config->max_frame_size);
         pn_transport_set_idle_timeout(tport, config->idle_timeout_seconds * 1000);
 
+        //
+        // Proton pushes out its trace to qd_log_tracer() which in turn writes a trace message to the qdrouter log
+        // If trace level logging is enabled on the router set PN_TRACE_FRM on the proton transport
+        //
+        if (qd_log_enabled(qd_server->log_source, QD_LOG_TRACE)) {
+            pn_transport_set_context(tport, ctx);
```
ted-ross: We should set the context unconditionally on all transports.

src/server.c
```diff
@@ -799,6 +822,9 @@ static void cxtr_try_open(void *context)
     // qdpn_connector is not thread safe
     //
     sys_mutex_lock(ct->server->lock);
+    // Increment the connection id so the next connection can use it
+    ctx->uid          = ct->server->next_connection_id;
+    ct->server->next_connection_id++;
```
ted-ross: Same comment as above.

src/server_private.h
```diff
@@ -98,6 +98,7 @@ struct qd_connection_t {
     void             *user_context;
     void             *link_context; // Context shared by this connection's links
     qd_user_fd_t     *ufd;
+    uint64_t         uid; // A unique identifier for the qd_connection_t. The underlying pn_connection already has one but it is long and clunky.
```
ted-ross: This should be called "connection_id".

src/server.c
```diff
@@ -64,6 +64,15 @@ static qd_thread_t *thread(qd_server_t *qd_server, int id)
 
 qd_error_t qd_entity_update_connection(qd_entity_t* entity, void *impl);
 
+/**
+ * This function is set as the pn_transport->tracer and is invoked when proton tries to write the log message to pn_transport->tracer
+ */
+static void qd_log_tracer(pn_transport_t *transport, const char *message)
```
ted-ross: It would be more accurate to call this function "qd_transport_tracer".

PR #14, opened 2015-10-06
NO-JIRA - Adding system test to test linkRoutePattern attribute of di…

ganeshmurthy:
…spatch router

### DIFF
--- a/tests/system_tests_link_routes.py
+++ b/tests/system_tests_link_routes.py
@@ -1,0 +1,282 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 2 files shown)


tests/system_tests_link_routes.py
```diff
@@ -0,0 +1,282 @@
...
+        b_listener_port = cls.tester.get_port()
+        c_listener_port = cls.tester.get_port()
+
+        router('A',
+               [
+                   ('listener', {'role': 'normal', 'addr': '0.0.0.0', 'port': a_listener_port, 'saslMechanisms': 'ANONYMOUS'}),
+               ])
+        router('B',
+               [
+                   ('listener', {'role': 'inter-router', 'addr': '0.0.0.0', 'port': b_listener_port, 'saslMechanisms': 'ANONYMOUS'}),
```
ted-ross: Router B shouldn't need two inter-router connections.  One will do.

ted-ross:
This should not be a NO-JIRA commit.  You should open a Jira for this feature.

PR #19, opened 2015-10-16
DISPATCH-163 - Added username and password for authentication of conn…

ganeshmurthy:
…ectors. Added system test to verify SASL PLAIN authentication


ChugR:
How do we manage these credentials internally? Or when viewed through the management interface?
Will 'qdmanage query' show the credentials in plain text?

ganeshmurthy:
You can use - sasldblistusers2 -f qdrouterd.sasldb - to see the user names (passwords will not be displayed)

saslpasswd2 can be used to add.modify users. 

I am not aware of other credential management tools.

PR #21, opened 2015-10-20
DISPATCH-163 - Added username and password for authentication of conn…

ganeshmurthy:
…ectors. Added system test to verify SASL PLAIN authentication


ganeshmurthy:
I have introduced an important change to the tests/CMakeLists.txt. The python system tests are copied into the build/tests folder and executed from there. I made sure that all tests pass when I ran "ctest -VV" from the build folder.

ganeshmurthy:
Does the CI system execute "ctest -VV" to run the tests?

tests/CMakeLists.txt
```diff
@@ -61,20 +73,22 @@ add_test(unit_tests_size_3     ${TEST_WRAP} --vg unit_tests_size 3)
...
-add_system_test(system_tests_broker)
-add_system_test(system_tests_management)
-add_system_test(system_tests_one_router)
-add_system_test(system_tests_qdmanage)
-add_system_test(system_tests_qdstat)
-add_system_test(system_tests_two_routers)
-add_system_test(system_tests_link_routes)
+# Add all sytem_tests* using add_test
+add_test(router_tests             ${TEST_WRAP} -s ${CMAKE_CURRENT_SOURCE_DIR}/router_engine_test.py -v)
+add_test(system_tests_broker      ${TEST_WRAP} -s ${CMAKE_CURRENT_BINARY_DIR}/system_tests_broker.py -v)
```
ted-ross: Why do you move the tests to the build folder when you could simply run them from the source folder here?  Then you could run the "configured" tests from the build folder and the non-configured ones from the source folder.

src/server.c
```diff
@@ -838,6 +840,13 @@ static void cxtr_try_open(void *context)
         return;
     }
 
+    // Set the sasl user name and password on the proton connection object. This has to be done before the call to qdpn_connector_transport() which
+    // binds the transport to the connection
+    if(config->sasl_username && config->sasl_password) {
+        pn_connection_set_user(ctx->pn_conn, config->sasl_username);
+        pn_connection_set_password(ctx->pn_conn, config->sasl_password);
```
ted-ross: I think you should treat the username and password independently.  You shouldn't assume that there will always be both present.

ganeshmurthy:
I was talking about this to crolke yesterday and the thought was that running all tests from the same folder (in this case build/tests) will make it look uniform instead of running some tests from source and some tests from build

src/server.c
ganeshmurthy: I will go ahead and make this fix.

ganeshmurthy:
ted-ross, do you still want me to run the "configured" tests from the build folder and the non-configured ones from the source folder?

ChugR:
I tried this new branch and I think it's better for tests (*.py) all to end up in the <build>/tests folder.
At first the test failed. The problem was that the *.pyc files are all in the <source>/tests folder and there was a stale system_tests_sasl_plain.pyc that was running.
If all the .py files are in the build folder then the source is pristine and the build artifacts from one build are not interfering with another build.
Either way the new test passes for me on Fedora 21.

ganeshmurthy:
ted-ross, I will close this PR and create a new PR which will include the server.c change that you pointed out earlier.

PR #30, opened 2015-11-09
NO-JIRA - Added some detail to the error message that is printed when…

ganeshmurthy:
… qdstat is invoked without parameters

### DIFF
--- a/tools/qdstat
+++ b/tools/qdstat
@@ -54,7 +54,15 @@ def parse_args(argv):
     opts, args = parser.parse_args(args=argv)
 
     if not opts.show:
-        parser.error("You must specify one of these options: -g, -c, -l, -n, -a, -m, -h or --log.")
+        parser.error("You must specify one of these options: "
+                     "-g (Show General Router Stats), "
+                     "-c (Show Connections) , "
+                     "-l (Show Router Links), "
+                     "-n (Show Router Nodes), "
+                     "-a (Show Router Addresses), "
+                     "-m (Show Broker Memory Stats), "
+                     "-v (Show verbose maximum detail), "
+                     "--log (Show recent log entries).")
 
     return opts, args
 


tools/qdstat
```diff
@@ -54,7 +54,15 @@ def parse_args(argv):
...
-        parser.error("You must specify one of these options: -g, -c, -l, -n, -a, -m, -h or --log.")
+        parser.error("You must specify one of these options: "
+                     "-g (Show General Router Stats), "
+                     "-c (Show Connections) , "
+                     "-l (Show Router Links), "
+                     "-n (Show Router Nodes), "
+                     "-a (Show Router Addresses), "
+                     "-m (Show Broker Memory Stats), "
+                     "-v (Show verbose maximum detail), "
+                     "--log (Show recent log entries).")
```
ted-ross: The new information is misleading (-v should not be in there and -h should).
Rather than reproducing a subset of the already existing help, why not just add a line that says "use --help for more information"?

ganeshmurthy:
Any idea where the code to handle the -h option is? 

Also, if -v is not available, the following line must be removed in tools/qdstat - 

parser.add_option("-v", "--verbose", help="Show maximum detail",        action="store_true", dest="verbose")

ted-ross:
On 11/09/2015 01:58 PM, Ganesh Murthy wrote:
> Any idea where the code to handle the -h option is?

The --help option is handled by the optparse module.

>
> Also, if -v is not available, the following line must be removed in tools/qdstat -
>
> parser.add_option("-v", "--verbose", help="Show maximum detail",        action="store_true", dest="verbose")

The -v option is valid as a modifier to other options.  "qdstat -v" is 
not a valid command line as -v is not a standalone option.  "qdstat -n 
-v" is valid and will show more detail than "qdstat -n".

PR #34, opened 2015-11-18
DISPATCH-179 - Added boilerplate code to cover delete and create mana…

ganeshmurthy:
…gement operations

### DIFF
--- a/include/qpid/dispatch/router_core.h
+++ b/include/qpid/dispatch/router_core.h
@@ -262,7 +262,7 @@ void qdr_manage_create(qdr_core_t *core, void *context, qd_router_entity_type_t 
  * @param identity The identity supplied with the request (or 0 if the name was supplied)
  */
 void qdr_manage_delete(qdr_core_t *core, void *context, qd_router_entity_type_t type,
-                       qd_field_iterator_t *name, qd_field_iterator_t *identity);
+                       qd_field_iterator_t *name, qd_field_iterator_t *identity, qd_composed_field_t *body);
 
 /**
  * qdr_manage_read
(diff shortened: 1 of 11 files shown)


include/qpid/dispatch/router_core.h
```diff
@@ -262,7 +262,7 @@ void qdr_manage_create(qdr_core_t *core, void *context, qd_router_entity_type_t
  * @param identity The identity supplied with the request (or 0 if the name was supplied)
  */
 void qdr_manage_delete(qdr_core_t *core, void *context, qd_router_entity_type_t type,
-                       qd_field_iterator_t *name, qd_field_iterator_t *identity);
+                       qd_field_iterator_t *name, qd_field_iterator_t *identity, qd_composed_field_t *body);
```
ted-ross: Why do you need this?  The body is always an empty map.  I don't see any need to pass it into the core.

src/router_core/agent_waypoint.c
```diff
@@ -0,0 +1,175 @@
...
+ * under the License.
+ */
+
+#include "agent_waypoint.h"
+
+const char *address   = "address";
+const char *connector = "connector";
+const char *inPhase   = "inPhase";
+const char *outPhase  = "outPhase";
+const char *mode      = "mode";
```
ted-ross: These constant strings are redundant with the ones below in the columns array.

src/router_core/agent_waypoint.c
```diff
@@ -0,0 +1,175 @@
...
+
+    // TODO - Add more logic here.
+    if (success) {
+        // If the request was successful then the statusCode MUST be 204 (No Content).
+        query->status = &QD_AMQP_NO_CONTENT;
+    }
+
+    // The body of the message MUST consist of an amqp-value section containing a Map with zero entries.
+    qd_compose_start_map(query->body);
+    qd_compose_end_map(query->body);
```
ted-ross: This should be done outside the core.  The core isn't adding any value here.

PR #36, opened 2015-11-23
DISPATCH-179 - Added code to handle the management update operation. …

ganeshmurthy:
…Also added code to check the parameters to create and update operations before control reaches core

### DIFF
--- a/include/qpid/dispatch/router_core.h
+++ b/include/qpid/dispatch/router_core.h
@@ -322,5 +322,5 @@ void qdr_query_free(qdr_query_t *query);
 
 typedef void (*qdr_manage_response_t) (void *context, const qd_amqp_error_t *status, bool more);
 void qdr_manage_handler(qdr_core_t *core, qdr_manage_response_t response_handler);
-
+bool qdr_is_set_manage_handler(qdr_core_t *core);
 #endif
--- a/src/router_core/management_agent.c
+++ b/src/router_core/management_agent.c
@@ -27,6 +27,7 @@
 #include "management_agent_private.h"
 #include "router_core_private.h"
 #include "dispatch_private.h"
+#include "agent_waypoint.h"
 #include "alloc.h"
 
 const char *entity_type_key = "entityType";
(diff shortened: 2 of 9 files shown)


include/qpid/dispatch/router_core.h
```diff
@@ -322,5 +322,5 @@ void qdr_query_free(qdr_query_t *query);
 
 typedef void (*qdr_manage_response_t) (void *context, const qd_amqp_error_t *status, bool more);
 void qdr_manage_handler(qdr_core_t *core, qdr_manage_response_t response_handler);
-
+bool qdr_is_set_manage_handler(qdr_core_t *core);
```
ted-ross: I don't think this is needed.  The "outer" agent should register its callbacks once at initialization time.  It is not safe to register new handlers for each request that comes in and it's not necessary to repeatedly register the same handler.

src/router_core/management_agent_private.h
```diff
@@ -20,6 +20,6 @@
  * under the License.
  */
 
-void management_agent_handler(void *context, qd_message_t *msg, int link_id);
+void qd_management_agent_handler(void *context, qd_message_t *msg, int link_id);
```
ted-ross: I think a more descriptive name is needed here.  Something like qd_router_agent_on_message.

src/router_core/management_agent.c
```diff
@@ -268,6 +269,62 @@ static void qd_core_agent_read_handler(qd_dispatch_t              *qd,
...
+        return false;
+
+    return true;
+}
+
+
+/**
+ * The body map containing any attributes that are not applicable for the entity being updated MUST result in a failure response with a statusCode of 400 (Bad Request).
+ */
+static bool qd_check_body(qd_router_entity_type_t  entity_type,
```
ted-ross: This function's signature suggests that it is simply checking the message body.  In fact, it also has the side effect of composing the error message for the response.  I think it would be better to stick with the simple check.  The calling logic will then be clearer to understand as it will either go forward with the request or create an error response based on the outcome of this check.

src/router_core/management_agent.c
```diff
@@ -268,6 +269,62 @@ static void qd_core_agent_read_handler(qd_dispatch_t              *qd,
...
+                          qd_parsed_field_t       *in_body,
+                          qd_composed_field_t     *out_body,
+                          qd_management_context_t *ctx)
+{
+    if (entity_type == QD_ROUTER_WAYPOINT) {
+        if(!qd_is_valid_waypoint_keys(in_body)){
```
ted-ross: Rather than have a proliferation of nearly identical type-specific validation functions, why not do a more general check on the appropriate attribute-name array?  This way, you can write the logic once for all types.

PR #38, opened 2015-12-02
DISPATCH-190 - Exposed the protocol family for the configuration of l…

ganeshmurthy:
…isteners and connectors. Modified relevant unit tests to use protocol family

### DIFF
--- a/include/qpid/dispatch/driver.h
+++ b/include/qpid/dispatch/driver.h
@@ -142,12 +142,16 @@ void qdpn_driver_free(qdpn_driver_t *driver);
  * @param[in] driver driver that will 'own' this listener
  * @param[in] host local host address to listen on
  * @param[in] port local port to listen on
+ * @param[in] protocol family to use (IPv4 or IPv6)
  * @param[in] context application-supplied, can be accessed via
  *                    qdpn_listener_context()
  * @return a new listener on the given host:port, NULL if error
  */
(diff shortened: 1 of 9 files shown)


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -633,7 +633,13 @@
                     "default": "both",
                     "description": "['in', 'out', 'both', 'no'] in: Strip the dispatch router specific annotations only on ingress; out: Strip the dispatch router specific annotations only on egress; both: Strip the dispatch router specific annotations on both ingress and egress; no - do not strip dispatch router specific annotations",
                     "create": true
-                }
+                },
+                "protocolFamily": {
```
ted-ross: Rather than duplicate this entry for both listener and connector, you could put this in the "addrPort" annotation.  Like addr and port, protocolFamily is common to both entity types.

You have a default of IPv4 in the connector.  That is inconsistent with the document string.

include/qpid/dispatch/driver.h
```diff
@@ -142,12 +142,16 @@ void qdpn_driver_free(qdpn_driver_t *driver);
  * @param[in] driver driver that will 'own' this listener
  * @param[in] host local host address to listen on
  * @param[in] port local port to listen on
+ * @param[in] protocol family to use (IPv4 or IPv6)
```
ted-ross: You should include 0 as a valid option as well.  It has a specific behavior.  Likewise on connector.

PR #40, opened 2015-12-14
NO-JIRA: Added dockerfiles that will help launch the dispatch router …

ganeshmurthy:
…inside docker containers

### DIFF
--- a/dockerfiles/Dockerfile-fedora
+++ b/dockerfiles/Dockerfile-fedora
@@ -1,0 +1,56 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# Downloads and installs apache proton and qpid dispatch on the latest version of Fedora and starts the dispatch router
+# /main is the top level folder under which proton (/main/qpid-proton/) and dispatch (/main/qpid-dispatch) source code is downloaded from github
+# /usr/local/lib64 is the folder in which the proton artifacts are installed.
+# /usr/local/sbin is the folder in which dispatch executable installed
+# Copy this docker file to your local folder. Build the docker file like this - sudo docker build -t <username>/dispatch --file=Dockerfile-fedora . (don't miss the dot at the end)
+# To run it - sudo docker run -i -t <username>/dispatch (this will launch the dispatch router)
+
+# Gets the latest Fedora from dockerhub
+FROM fedora:latest
+
+#Uncomment the following line to specify a maintainer
+#MAINTAINER John Doe "jdoe@apache.org"
+
+# Install all the required packages. Some in this list were picked off from proton's INSTALL.md (https://github.com/apache/qpid-proton/blob/master/INSTALL.md) and the rest are from dispatch (https://github.com/apache/qpid-dispatch/blob/master/README)
+RUN dnf -y install gcc cmake libuuid-devel openssl-devel cyrus-sasl-devel cyrus-sasl-plain cyrus-sasl-gssapi cyrus-sasl-md5 swig python-devel ruby-devel php-devel perl-devel java-1.8.0-openjdk-devel git make doxygen valgrind emacs
+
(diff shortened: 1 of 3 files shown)


dockerfiles/Dockerfile-fedora
```diff
@@ -0,0 +1,56 @@
...
+# /usr/local/lib64 is the folder in which the proton artifacts are installed.
+# /usr/local/sbin is the folder in which dispatch executable installed
+# Copy this docker file to your local folder. Build the docker file like this - sudo docker build -t <username>/dispatch --file=Dockerfile-fedora . (don't miss the dot at the end)
+# To run it - sudo docker run -i -t <username>/dispatch (this will launch the dispatch router)
+
+# Gets the latest Fedora from dockerhub
+FROM fedora:latest
+
+#Uncomment the following line to specify a maintainer
+#MAINTAINER John Doe "jdoe@apache.org"
```
ted-ross: Why not just set the maintainer to dev@qpid.apache.org ?

dockerfiles/Dockerfile-fedora
```diff
@@ -0,0 +1,56 @@
...
+# make and install proton
+RUN cmake .. -DSYSINSTALL_BINDINGS=ON && make install
+
+# Clone the qpid-dispatch git repo
+RUN cd /main && git clone https://github.com/apache/qpid-dispatch.git && mkdir /main/qpid-dispatch/build
+
+WORKDIR /main/qpid-dispatch/build
+RUN cmake .. -DCMAKE_INSTALL_PREFIX=/usr && make install
+
+ENV LD_LIBRARY_PATH=/usr/local/lib64:$LD_LIBRARY_PATH
```
ted-ross: This would not be necessary if you installed proton in /usr like you did with dispatch.

dockerfiles/README.md
```diff
@@ -0,0 +1,35 @@
...
+The two docker files, Dockerfile-fedora and Dockerfile-ubuntu provide quick start
+docker containers with an running instance of qpid dispatch router.
+
+The dockerfiles follow these steps before launching the dispatch router
+
+* Downloads all the fedora or ubuntu specific dependencies needed for building apache qpid-proton.
+* Downloads the source code of qpid-proton to /main/qpid-proton
+* Builds and installs qpid-proton using gcc. /usr/local/lib64 is the folder in which the qpid-proton artifacts are installed.
+* Downloads the source code of qpid-dispatch to /main/qpid-dispatch.
+* Builds and installs qpid-dispatch using gcc.  /usr/local/sbin is the folder in which qpid-dispatch executable is installed
```
ted-ross: This is inconsistent with the dockerfile text.  qpid-dispatch is installed in /usr, not /usr/local (and proton should be as well).

dockerfiles/README.md
```diff
@@ -0,0 +1,35 @@
...
+* Launch qpid-dispatch router
+
+####Building and running Dockerfiles
+* To build the fedora docker file
+ * sudo docker build -t username/dispatch-fedora:latest --file=Dockerfile-fedora  .  (substitute username with your username e.g. johndoe)
```
ted-ross: I don't think the username is required here.  If a username is used, it must be the username for the Docker Index subscription (not the apache or local username).  It's ok to omit the username in the image title if there's no plan to push the image to the Docker Index.

dockerfiles/README.md
```diff
@@ -0,0 +1,35 @@
...
+ * sudo docker build -t username/dispatch-ubuntu:latest --file=Dockerfile-ubuntu  .  (substitute username with your username e.g. johndoe)
+* To run the fedora docker file
+ * sudo docker run -i -t username/dispatch-fedora:latest (substitute username with your username e.g. johndoe)
+* To run the ubuntu docker file
+ * sudo docker run -i -t username/dispatch-ubuntu:latest (substitute username with your username e.g. johndoe)
+* After the docker run command is invoked, execute the ___sudo docker ps___ command to check if your container name shows up in the list of docker processes.
+
+####Customizing Docker files
+* If you want to change the router configuration by modifying the contents of qdrouterd.conf (/etc/qpid-dispatch/qdrouterd.conf), replace the last line in the Dockerfile that launches the dispatch router (CMD ["qdrouterd"]) with RUN ["/bin/bash"] and build and run the container. This will put you inside the container from where you can modify the contents of qdrouterd.conf and relaunch the dispatch router to use the modified config.
+* Uncomment the line RUN ctest -VV to run dispatch unit and system tests
```
ted-ross: This might be a bit too much information for the readme.  We don't want to write a Docker tutorial.
This section, "Customizing Docker files", is a bit mis-named.  What is described does not customize the Dockerfile.  It simply changes the image to launch a shell so the user can make non-persistent modifications to the running image.  Note that any modifications that are made will be lost once the container is shut down.
To make a customized image is a bit more involved and is certainly beyond the scope of a Dispatch Router readme.

ted-ross:
I think you should raise a Jira for this, so this addition can be mentioned in the next release notes.

scholzj:
I think the idea is good for Dispatch development and testing. But the images are IMHO not really sufficient for some "more serious" use (because they are always based on the Git master, because they contain all the compilers which are not necessary to run the image etc.). Maybe the README.md should be more clear about that, so that nobody takes image built using these docker files and uses it in production?

PR #43, opened 2016-02-05
DISPATCH-207 - Allow empty message bodies. This will prevent dispatch…

ganeshmurthy:
… router from displaying the 'Invalid body value' error

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -318,8 +318,10 @@ static int qd_check_and_advance(qd_buffer_t         **buffer,
     int pre_consume = 1;  // Count the already extracted tag
     int consume     = 0;
     unsigned char tag = next_octet(&test_cursor, &test_buffer);
-
-    switch (tag & 0xF0) {
+    int tag_bit = tag & 0xF0;
+    if (!test_cursor && tag_bit != 0x40)
+        return 0;
+    switch (tag_bit) {
     case 0x40:               break;
     case 0x50: consume = 1;  break;
     case 0x60: consume = 2;  break;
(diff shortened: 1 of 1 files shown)


src/message.c
```diff
@@ -318,7 +318,7 @@ static int qd_check_and_advance(qd_buffer_t         **buffer,
     int pre_consume = 1;  // Count the already extracted tag
     int consume     = 0;
     unsigned char tag = next_octet(&test_cursor, &test_buffer);
-    if (!test_cursor) return 0;
+
```
ted-ross: Taking away this test removes the protection it provided and means that test_cursor may be referenced-as-null in the body of the following switch statement.  Here's a test case that I think may cause a problem with this change:  The body (from the tag forward) contains only E0 with no data following.  This is a malformed field, but it would cause next_octet to be called with a null cursor.
A better solution would be to leave the test in place, but allow it to pass only if the tag is a 4x-class tag (i.e. zero length).

src/message.c
```diff
@@ -318,8 +318,10 @@ static int qd_check_and_advance(qd_buffer_t         **buffer,
     int pre_consume = 1;  // Count the already extracted tag
     int consume     = 0;
     unsigned char tag = next_octet(&test_cursor, &test_buffer);
-    if (!test_cursor) return 0;
-    switch (tag & 0xF0) {
+    int tag_bit = tag & 0xF0;
```
ted-ross: This looks ok.  One minor change that's needed:  tag is an unsigned char, so tag_bit must be of the same type.  Also, 0xF0 represents 4 bits so tag_bit might not be the best name.  The spec calls it a subcategory.  Maybe use that.

PR #55, opened 2016-03-08
DISPATCH-232 - Add capability to delete listeners and connectors via …

ganeshmurthy:
…the qdmanage DELETE operation


ted-ross:
Please don't merge this patch.  The use of "label" as the name is going to cause a problem when DISPATCH-179 is merged.

PR #61, opened 2016-03-29
DISPATCH-200 - Add a flexible mapping from x.509 certificate fields t…

ganeshmurthy:
…o an identity


src/container.c
```diff
@@ -21,6 +21,7 @@
 #include <string.h>
 #include "dispatch_private.h"
 #include "connection_manager_private.h"
+#include "server_private.h"
```
ted-ross: It would be better to not add this private linkage between container and server.  Rather, put the signature for qd_connection_set_user into the public server.h file.

tests/management/qdrouter.py
```diff
@@ -82,22 +81,11 @@ def do_test_qdrouter_parse(self, text):
...
         content = conf._parse(text.split("\n"))
         self.maxDiff = None
-        expect = [
-            [u"router", {u"mode":u"standalone"}],
-            [u"sslProfile", {u"name":u"test-profile", u"password":u"secret"}],
-            [u"listener", {u"name":u"l0", u"saslMechanisms":u"ANONYMOUS", u"sslProfile":u"test-profile"}],
-            [u"listener", {u"identity":u"l1", u"saslMechanisms":u"ANONYMOUS", u"port":u"1234"}],
-            [u"listener", {u"saslMechanisms":u"ANONYMOUS", u"port":u"4567"}]
-        ]
+        expect = [[u'router', {u'mode': u'standalone'}], [u'sslProfile', {u'password': u'secret', u'name': u'test-profile'}], [u'listener', {u'sslProfile': u'test-profile', u'name': u'l0', u'saslMechanisms': u'ANONYMOUS'}], [u'listener', {u'saslMechanisms': u'ANONYMOUS', u'identity': u'l1', u'port': u'1234'}], [u'listener', {u'saslMechanisms': u'ANONYMOUS', u'port': u'4567'}]]
```
ted-ross: Why did you change these?  They were so much more readable before.

PR #62, opened 2016-04-01
DISPATCH-235 - Added code so the router respects the receiver's send …

ganeshmurthy:
…settle mode


ted-ross:
This looks good, feel free to commit it.

PR #65, opened 2016-04-18
DISPATCH-261 - Added a new field called state to qd_config_listener_t…

ganeshmurthy:
… so it can be checked before activating listeners


ted-ross:
This looks good to me.

PR #70, opened 2016-05-05
DISPATCH-306 - Deprecated addr and routerId and introduce host and id…

ganeshmurthy:
… instead


ted-ross:
This looks good.  Please re-base to master and commit.

PR #73, opened 2016-05-11
DISPATCH-320 - Added a new connector property called verifyHostName w…

ganeshmurthy:
…hich will verify host name on SSL connections


ted-ross:
+1  Please re-base to master and commit.

PR #74, opened 2016-05-13
DISPATCH-335 - Added code to balance link routes over remote routers

ganeshmurthy:


src/router_core/forwarder.c
```diff
@@ -559,20 +559,34 @@ bool qdr_forward_link_balanced_CT(qdr_core_t     *core,
...
+        if (addr->cost_epoch != core->cost_epoch) {
+            addr->next_remote = -1;
+            addr->cost_epoch  = core->cost_epoch;
+        }
 
-                if (next_node && next_node->peer_data_link)
-                    conn = next_node->peer_data_link->conn;
-            }
+        if (addr->next_remote < 0) {
+            qd_bitmask_first_set(addr->rnodes, &addr->next_remote);
```
ted-ross: You removed the check of the return value of qd_bitmask_first_set.  If there is no bit set in the mask, bad things will happen.

ganeshmurthy: How about I check addr->next_remote >= 0 like below

        qdr_node_t *next_node;

        if (addr->cost_epoch != core->cost_epoch) {
            addr->next_remote = -1;
            addr->cost_epoch  = core->cost_epoch;
        }

        if (addr->next_remote < 0) {
            qd_bitmask_first_set(addr->rnodes, &addr->next_remote);
        }
        
        if (addr->next_remote >= 0) {

            qdr_node_t *rnode = core->routers_by_mask_bit[addr->next_remote];
    
            if (rnode) {
                //
                // Advance the addr->next_remote so there will be link balance across containers
                //
                _qdbm_next(addr->rnodes, &addr->next_remote);
                if (addr->next_remote == -1)
                    qd_bitmask_first_set(addr->rnodes, &addr->next_remote);
    
                if (rnode->next_hop)
                    next_node = rnode->next_hop;
                else
                    next_node = rnode;
    
                if (next_node && next_node->peer_data_link)
                    conn = next_node->peer_data_link->conn;
            }
        }

PR #76, opened 2016-05-20
DISPATCH-341 - Added drain support for message routes and link routes

ganeshmurthy:


src/router_core/connections.c
```diff
@@ -188,7 +188,10 @@ int qdr_connection_process(qdr_connection_t *conn)
         sys_mutex_unlock(conn->work_lock);
 
         if (link) {
-            core->flow_handler(core->user_context, link, link->incremental_credit);
+            if (link->drain_mode)
+                core->drained_handler(core->user_context, link);
+            else
```
ted-ross: What happens if there is credit _and_ drain mode?  Does the credit not get issued?

ganeshmurthy: I removed the else and made it 2 if statements

PR #79, opened 2016-06-01
DISPATCH-357 - Added a terminus_addr field to qdr_link_t which holds …

ganeshmurthy:
…the complete address of the link


ted-ross:
Two points on this change:
  - There are no tests included
  - Does this change work in all cases?  Specifically, does the proper visibility occur on the ingress router as well as the other routers in the link's path?  Also, does this work for both sender and receiver links?

PR #82, opened 2016-06-14
DISPATCH-184 - Resolve port name to int when applying configuration

jdanekrh:
Error in router is caused by a call to getaddrinfo() that uses the 'amqp' string for port In function qdpn_listener_t *qdpn_listener(qdpn_driver_t *driver, ...). Same call to getaddrinfo() also appears in qdpn_connector().

There ia a Python class Url.Port() that converts a string port to int and can handle missing /etc/services. Class used to be in dispatch in proton_future package, it was recently removed and now it is only in Proton.

As a solution, I had the following ideas:

1. Add a Port type to python/qpid_dispatch_internal/management/schema.py, do the conversion to int as part of validation.
2. Convert port to int in python/qpid_dispatch_internal/management/agent.py in ListenerEntity and ConnectorEntity, either in validate, create,  or init method.
3. Catch the error code in qdpn_listener and qdpn_connector and do a if strcmp(port, "amqp") ... there, Or call into Url.Port() in Python to avoid code duplication.

I decided to do a PR with the 2nd one.

### DIFF
--- a/python/qpid_dispatch_internal/management/agent.py
+++ b/python/qpid_dispatch_internal/management/agent.py
@@ -64,7 +64,7 @@ data that may be updated in other threads.
 Temporary solution is to lock the entire dispatch router lock during full refresh.
 Better solution coming soon...
 """
-
+import socket
 import traceback, json, pstats
 from itertools import ifilter, chain
 from traceback import format_exc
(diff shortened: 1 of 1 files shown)


jdanekrh:
@ganeshmurthy What do you think about my

1. Add a Port type to python/qpid_dispatch_internal/management/schema.py, do the conversion to int as part of validation.
2. Convert port to int in python/qpid_dispatch_internal/management/agent.py in ListenerEntity and ConnectorEntity, either in validate, create, or init method.
3. Catch the error code in qdpn_listener and qdpn_connector and do a if strcmp(port, "amqp") ... there, Or call into Url.Port() in Python to avoid code duplication.

dilemma? Is one of these preferable? I went with 2, but other may make more sense. Or surely even other options could be suggested.

PR #83, opened 2016-06-22
DISPATCH-401 - Verified host name by default and added a --no-verify-…

ganeshmurthy:
…host-name to disable host name verification

### DIFF
--- a/python/qpid_dispatch_internal/tools/command.py
+++ b/python/qpid_dispatch_internal/tools/command.py
@@ -83,6 +83,11 @@ def connection_options(options, title="Connection Options"):
                      help="Trusted Certificate Authority Database file (PEM Format)")
     group.add_option("--ssl-password", action="store", type="string", metavar="PASSWORD",
                      help="Certificate password, will be prompted if not specifed.")
+    group.add_option("--no-verify-host-name", action="store_true", default=False,
+                     help="Does not ensure that when initiating an SSL connection (as a client) the host name in "
+                          "the URL to which this connector connects to matches the host name in the digital "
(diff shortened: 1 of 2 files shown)


python/qpid_dispatch_internal/tools/command.py
```diff
@@ -83,6 +83,11 @@ def connection_options(options, title="Connection Options"):
                      help="Trusted Certificate Authority Database file (PEM Format)")
     group.add_option("--ssl-password", action="store", type="string", metavar="PASSWORD",
                      help="Certificate password, will be prompted if not specifed.")
+    group.add_option("--no-verify-host-name", action="store_true", default=False,
```
alanconway: I'd suggest: --verify-peer-name type=bool, default=true. Bool options with negative names are confusing and "host" is a bit vague - peer is more clearly the _other_ host. Also the help text is very long, maybe "Verify the peer host name matches the certificate. Default true, setting to false is insecure ."

grs: Maybe add 'ssl' to the option also.

ganeshmurthy:
alanconway, 

@alanconway,  if I understood correctly, to disable hostname verification, one must do the following -

qdstat -c --verify-peer-name=false --ssl-trustfile=/home/gmurthy/opensource/dispatch/tests/config-2/ca-certificate.pem --ssl-certificate=/home/gmurthy/opensource/dispatch/tests/config-2/client-certificate.pem --ssl-key=/home/gmurthy/opensource/dispatch/tests/config-2/client-private-key.pem --ssl-password=client-password

PR #84, opened 2016-06-23
DISPATCH-401 - Made qdstat and qdmanage verify host name by default. …

ganeshmurthy:
…Added --ssl-allow-peer-name-mismatch to allow peer name mismatch


python/qpid_dispatch_internal/tools/command.py
```diff
@@ -83,6 +83,9 @@ def connection_options(options, title="Connection Options"):
                      help="Trusted Certificate Authority Database file (PEM Format)")
     group.add_option("--ssl-password", action="store", type="string", metavar="PASSWORD",
                      help="Certificate password, will be prompted if not specifed.")
+    group.add_option("--ssl-allow-peer-name-mismatch", action="store_true", default=False,
+                     help="Verify the peer host name matches the certificate. Default true, "
+                          "setting to false is insecure .")
     return group
```
ted-ross: The name of this option seems inverted.  Allowing a mismatch is insecure, no?  How about "--ssl-verify-hostname"?

PR #88, opened 2016-07-01
DISPATCH-441 Support password environment variable in qdrouter config

eric2a:
password attribute can take an env variable or literal value.

password: env:<string>
or 
password: literal:<string>


eric2a:
password attribute can take an env variable or literal value.
password: env:string
or
password: literal:string

PR #96, opened 2016-08-23
 Use atomic ops for ref_counts

lulf:


src/connection_manager.c
```diff
@@ -254,9 +252,8 @@ qd_config_ssl_profile_t *qd_dispatch_configure_ssl_profile(qd_dispatch_t *qd, qd
     ssl_profile->ssl_trusted_certificates   = qd_entity_opt_string(entity, "trustedCerts", 0); CHECK();
     ssl_profile->ssl_uid_format             = qd_entity_opt_string(entity, "uidFormat", 0); CHECK();
     ssl_profile->ssl_display_name_file      = qd_entity_opt_string(entity, "displayNameFile", 0); CHECK();
-    sys_mutex_lock(qd->connection_manager->ssl_profile_lock);
-    ssl_profile->ref_count                  = 0;
-    sys_mutex_unlock(qd->connection_manager->ssl_profile_lock);
+
+    __sync_and_and_fetch(&ssl_profile->ref_count, 0);
```
grs: Typo? Should this be __sync_add_and_fetch?

lulf: No, i didn't find an atomic assign, but come to think of it, i think just regular assignment should work. This was just a workaround I did without thinking.

grs: Ah, I get it now! Sorry for the noise!

ted-ross:
I like this change very much (I had planned to do it myself eventually).  It would be better I think if rather than sprinkle the directives throughout the code, you could make a portable abstraction (similar to the threading module).  Even if the abstraction is made up of macros, that would be ok.  I assume that other compilers might use different syntax for atomic operations, and some might not support it at all and might need to revert to mutexes.
Perhaps there should be a portable ref_count abstraction with increment and decrement-test-for-zero operations.

lulf:
@ted-ross Good idea! I've updated the PR with a header for atomic operations that supports c11, gnu/clang builtins and uses mutexes as a fallback for other compilers. It might also be that the flags checked for GNUC and clang could be refined somewhat, but I think these builtins have been a part of those compilers for a long time.  I don't have access to visual studio, so I haven't tested that.

It's not pretty, but I've tried to create as little overhead for this abstraction as possible. I wasn't sure if this should be in the public api or maybe in some internal header.

A generic ref_count/smart pointer abstraction can be built using this as the next step if desired.

alanconway:
I would wrap this (with inline functions or macros to avoid function call overhead) so that the impl details can be changed by platform. I think these ops are gcc specific, they definitely won't work on windows. Check the qpid-cpp codebase it has an atomic counter impl that uses GCC, windows or defaults to mutex. Its C++ but you can do  something similar in C.

lulf:
@alanconway i think you might be commenting on an outdated diff. Look at the 'files changed' tab.

alanconway:
Yep - this looks great, I approve.

PR #98, opened 2016-09-14
DISPATCH-498 - Used qd_config_ssl_profile_free instead of free()

ganeshmurthy:


src/connection_manager.c
```diff
@@ -359,24 +359,21 @@ void qd_connection_manager_free(qd_connection_manager_t *cm)
     qd_config_listener_t *cl = DEQ_HEAD(cm->config_listeners);
     while (cl) {
         DEQ_REMOVE_HEAD(cm->config_listeners);
-        qd_server_listener_free(cl->listener);
         qd_server_config_free(&cl->configuration);
-        free(cl);
+        qd_config_listener_free(cm, cl);
```
ganeshmurthy: This is an additional bug I found. free(cl) and free(cc) were called instead of qd_config_listener_free(cm, cl) and qd_config_connector_free(cm, cc) respectively. I fixed that in this method.

src/connection_manager.c
```diff
@@ -486,11 +483,8 @@ void qd_connection_manager_delete_listener(qd_dispatch_t *qd, void *impl)
 bool qd_connection_manager_delete_ssl_profile(qd_dispatch_t *qd, void *impl)
 {
     qd_config_ssl_profile_t *ssl_profile = (qd_config_ssl_profile_t*) impl;
-    if(ssl_profile) {
-        bool freed = qd_config_ssl_profile_free(qd->connection_manager, ssl_profile);
-
-        return freed;
-    }
+    if(ssl_profile)
```
ted-ross: This looks wrong.  Shouldn't it be if (!ssl_profile) ?

ted-ross: I retract the previous comment.

PR #99, opened 2016-09-14
DISPATCH-500 - Fixed code to use qdr_field_t to pass around the name …

ganeshmurthy:
…and identity of management objects

### DIFF
--- a/src/router_config.c
+++ b/src/router_config.c
@@ -280,7 +280,12 @@ qd_error_t qd_router_configure_address(qd_router_t *router, qd_entity_t *entity)
         qd_parsed_field_t   *in_body = qd_parse(iter);
         qd_field_iterator_free(iter);
 
-        qdr_manage_create(router->router_core, 0, QD_ROUTER_CONFIG_ADDRESS, 0, in_body, 0);
+        qd_field_iterator_t *name_iter = 0;
+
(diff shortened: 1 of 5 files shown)


grs:
Is this for DISPATCH-502?

ganeshmurthy:
I verified that the newly added tests (test_check_address_name, test_check_link_route_name, test_check_auto_link_name) fail if this fix is not there

ganeshmurthy:
@grs, I think that this patch might fix DISPATCH-502 as well. I am taking a look. Thanks.

src/router_config.c
```diff
@@ -280,7 +280,12 @@ qd_error_t qd_router_configure_address(qd_router_t *router, qd_entity_t *entity)
         qd_parsed_field_t   *in_body = qd_parse(iter);
         qd_field_iterator_free(iter);
 
-        qdr_manage_create(router->router_core, 0, QD_ROUTER_CONFIG_ADDRESS, 0, in_body, 0);
+        qd_field_iterator_t *name_iter = 0;
+
+        if (name)
+            name_iter = qd_field_iterator_string(name);
```
ted-ross: I think you are going to leak this iterator.  After the call to qdr_manage_create, you must free the name_iter.

PR #101, opened 2016-09-20
DISPATCH-160 - Modified code to allow propagation of custom message a…

ganeshmurthy:
…nnotations

### DIFF
--- a/include/qpid/dispatch/iterator.h
+++ b/include/qpid/dispatch/iterator.h
@@ -180,7 +180,15 @@ void qd_address_iterator_set_phase(qd_field_iterator_t *iter, char phase);
  */
 void qd_address_iterator_override_prefix(qd_field_iterator_t *iter, char prefix);
 
-void qd_address_iterator_set_length(qd_field_iterator_t *iter, int length);
+/**
+ * Trims the iterator by setting the length of the iterator.
+ *
+ * The passed in length cannot be more than the current length of the iterator.
+ *
+ * @param iter - the iterator whose length should be trimmed
+ * @param length - the length to trim. This length cannot be more the current length of the iterator.
+ */
+void qd_field_iterator_trim(qd_field_iterator_t *iter, int length);
 
 /**
  * Return the current octet in the iterator's view and step to the next.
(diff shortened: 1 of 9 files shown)


include/qpid/dispatch/iterator.h
```diff
@@ -180,6 +180,8 @@ void qd_address_iterator_set_phase(qd_field_iterator_t *iter, char phase);
  */
 void qd_address_iterator_override_prefix(qd_field_iterator_t *iter, char prefix);
 
+void qd_address_iterator_set_length(qd_field_iterator_t *iter, int length);
```
ted-ross: You should add a comment to document this new method.
What happens if this method is called with a length that is greater than the current iterator's length?

src/message.c
```diff
@@ -787,44 +787,103 @@ static void send_handler(void *context, const unsigned char *start, int length)
     pn_link_send(pnl, (const char*) start, length);
 }
 
+/**
+ * Returns true if the data backing the iterator matches any of the router specific annotations like "x-opt-qd.ingress" or
+ * "x-opt-qd.trace" or "x-opt-qd.to" or "x-opt-qd.phase
+ */
+static bool qd_message_is_router_annotation(qd_field_iterator_t *iter)
```
ted-ross: Is this function even needed?  Why not use qd_field_iterator_prefix?

src/message.c
```diff
@@ -787,44 +787,103 @@ static void send_handler(void *context, const unsigned char *start, int length)
...
+{
+    bool is_router_annotation = true;
+    int i = 0;
+    while(! qd_field_iterator_end(iter)) {
```
ted-ross: I believe this function is not needed.   But for the review, there is a bug in this algorithm.  It will return 'true' if the iterator field is blank or a subset of the prefix.  For example, a key of 'x-o' will be incorrectly identified as a router annotation.

include/qpid/dispatch/iterator.h
ted-ross: Two things:  This function should be a "qd_field_iterator..." function, not a "qd_address_iterator..." function.  There is nothing address-specific about this operation.

Also, it might be clearer to call this "qd_field_iterator_trim" with an argument called "new_length".  This suggests that the length can only be reduced.

src/message.c
ganeshmurthy: I failed to read the doc for qd_field_iterator_prefix. I have used it now instead of qd_message_is_router_annotation

include/qpid/dispatch/iterator.h
ganeshmurthy: Done

ganeshmurthy: Done

PR #103, opened 2016-09-23
DISPATCH-521 - Added a version command line option that will display …

ganeshmurthy:
…the version of the dispatch router


CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ set (SO_VERSION_MAJOR 2)
 set (SO_VERSION_MINOR 0)
 set (SO_VERSION "${SO_VERSION_MAJOR}.${SO_VERSION_MINOR}")
 
+set (QPID_DISPATCH_VERSION_MAJOR 0)
```
ted-ross: The CMake file already passes the version from VERSION.txt into the set of defined macros in config.h.  I don't think it's desirable to create yet another authoritative source-of-version that needs to be maintained manually on each release.

PR #106, opened 2016-10-04
DISPATCH-451: Configurable settings for session capacity and max sessions

ChugR:
This patch allows listener and connector settings for session capacity and for max sessions.

It does _not_ allow a setting for max-message-size. Proton does not provide an interface for changing max message size. This issue is tracked at [PROTON-1311](https://issues.apache.org/jira/browse/PROTON-1311)

Work is progressing on self tests for this feature. Proton python and wrapper interfaces tend to hide sessions and make finding session state difficult.

### DIFF
--- a/include/qpid/dispatch/server.h
+++ b/include/qpid/dispatch/server.h
@@ -418,6 +418,22 @@ typedef struct qd_server_config_t {
     uint32_t max_frame_size;
 
     /**
+     * The max_sessions value is the number of sessions allowed on the Connection. 
+     * This value minus one is the Open performative channel-max setting.
+     */
+    uint32_t max_sessions;
+
+    /**
+     * The max_session_window value is the maximum number of outstanding octets that are
+     * allowed to be in flight on a session. This value is used to calculate the number of
+     * outstanding transfers that are allowed by the formula:
+     *   incoming_window = max_session_window / max_frame_size
+     * If max_session_window=1000000 and max_frame_size=32768 then 30 transfers may
+     * be outstanding before session flow control begins.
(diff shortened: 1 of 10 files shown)


include/qpid/dispatch/server.h
```diff
@@ -418,6 +418,22 @@ typedef struct qd_server_config_t {
     uint32_t max_frame_size;
 
     /**
+     * The max_sessions value is the number of sessions allowed on the Connection. 
+     * This value minus one is the Open performative channel-max setting.
+     */
+    uint32_t max_sessions;
+
+    /**
+     * The max_session_window value is the maximum number of outstanding octets that are
```
ted-ross: Might it be easier to explain this if the units were frames instead of octets?

include/qpid/dispatch/server.h
```diff
@@ -418,6 +418,22 @@ typedef struct qd_server_config_t {
     uint32_t max_frame_size;
 
     /**
+     * The max_sessions value is the number of sessions allowed on the Connection. 
+     * This value minus one is the Open performative channel-max setting.
+     */
+    uint32_t max_sessions;
```
ted-ross: What will be do with the value 0?  Should we use channel_max to avoid this question?

ChugR: There should be a comment that says if you specify zero then you get the default number. Using channel_max is awkward in its own way: if the setting is 1 then you get 2.
There is an identical setting in policy that already used the 1..N numbering (and not 0..N-1). They ought to be the same either way.

PR #108, opened 2016-10-31
Fix memory leaks

dskarbek:
Just getting my feet wet really.  This fixes all the memory leaks found by valgrind in just starting up a couple connected routers and shutting them back down.

### DIFF
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -98,9 +98,13 @@ static void qd_server_config_free(qd_server_config_t *cf)
     if (!cf) return;
     free(cf->host);
     free(cf->port);
-    free(cf->name);
     free(cf->role);
-    free(cf->sasl_mechanisms);
+    if (cf->name)            free(cf->name);
+    if (cf->protocol_family) free(cf->protocol_family);
+    if (cf->sasl_username)   free(cf->sasl_username);
+    if (cf->sasl_password)   free(cf->sasl_password);
+    if (cf->sasl_mechanisms) free(cf->sasl_mechanisms);
+    if (cf->ssl_profile)     free(cf->ssl_profile);
 
     memset(cf, 0, sizeof(*cf));
 }
@@ -164,6 +168,10 @@ static void set_config_host(qd_server_config_t *config, qd_entity_t* entity)
          config->host = host;
          free(addr);
     }
+    else {
+        free(host);
+        free(addr);
+    }
 
     assert(config->host);
 }
@@ -174,7 +182,6 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
 
     bool authenticatePeer   = qd_entity_opt_bool(entity, "authenticatePeer",  false);    CHECK();
     bool verifyHostName     = qd_entity_opt_bool(entity, "verifyHostName",    true);     CHECK();
-    char *stripAnnotations  = qd_entity_opt_string(entity, "stripAnnotations", 0);       CHECK();
     bool requireEncryption  = qd_entity_opt_bool(entity, "requireEncryption", false);    CHECK();
     bool requireSsl         = qd_entity_opt_bool(entity, "requireSsl",        false);    CHECK();
     bool depRequirePeerAuth = qd_entity_opt_bool(entity, "requirePeerAuth",   false);    CHECK();
@@ -240,11 +247,14 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
     config->allowInsecureAuthentication = true;
     config->verify_host_name = verifyHostName;
 
+    char *stripAnnotations  = qd_entity_opt_string(entity, "stripAnnotations", 0);
     load_strip_annotations(config, stripAnnotations);
+    free(stripAnnotations);
+    stripAnnotations = 0;
+    CHECK();
 
     config->requireAuthentication = authenticatePeer || depRequirePeerAuth;
     config->requireEncryption     = requireEncryption || !depAllowUnsecured;
-
 
     if (config->ssl_profile) {
         config->ssl_required = requireSsl || !depAllowUnsecured;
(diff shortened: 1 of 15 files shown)


src/router_core/router_core.c
```diff
@@ -106,29 +106,31 @@ void qdr_core_free(qdr_core_t *core)
...
+
+    qdr_address_t *addr = 0;
+    while ( (addr = DEQ_HEAD(core->addrs)) ) {
+        qdr_core_remove_address(core, addr);
     }
-    if (core->data_links_by_mask_bit) {
-        free(core->data_links_by_mask_bit);
+    qdr_address_config_t *addr_config = 0;
+    while ( (addr_config = DEQ_HEAD(core->addr_config))) {
+        free_qdr_address_config_t(addr_config);
```
dskarbek: this is a bug.  I've got an update to the commit that fixes this which I haven't pushed yet.

src/router_config.c
```diff
@@ -26,6 +26,27 @@
...
+
+    qd_field_iterator_t *iter = qd_field_iterator_buffer(DEQ_HEAD(buffers), 0, qd_buffer_list_length(&buffers));
+    qd_parsed_field_t   *in_body = qd_parse(iter);
+    qd_field_iterator_free(iter);
+
+    qd_field_iterator_t *name_iter = 0;
+    if (name)
+        name_iter = qd_field_iterator_string(name);
+
+    qdr_manage_create(core, 0, type, name_iter, in_body, 0, buffers);
```
dskarbek: What do you think about passing the buffers through the action so that the action method can free them when it is done?

ganeshmurthy: You are correct that these buffers are not freed. We are thinking of other ways to free the buffers without having to pass it thru to the qdr_manage_create() method because it might not be relevant to the method itself other than the fact that the method is freeing it.

src/connection_manager.c
```diff
@@ -98,9 +98,13 @@ static void qd_server_config_free(qd_server_config_t *cf)
     if (!cf) return;
     free(cf->host);
     free(cf->port);
-    free(cf->name);
     free(cf->role);
-    free(cf->sasl_mechanisms);
+    if (cf->name)            free(cf->name);
```
ganeshmurthy: This is a good change. It makes sure that it frees all the components of qd_server_config_t

src/connection_manager.c
```diff
@@ -240,12 +247,15 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
     config->allowInsecureAuthentication = true;
     config->verify_host_name = verifyHostName;
 
+    char *stripAnnotations  = qd_entity_opt_string(entity, "stripAnnotations", 0);
```
ganeshmurthy: The existing code already seems to free stripAnnotations

src/dispatch_private.h
```diff
@@ -141,4 +141,16 @@ void qd_dispatch_unregister_entity(qd_dispatch_t *qd, void *impl);
 /** Set the agent */
 void qd_dispatch_set_agent(qd_dispatch_t *qd, void *agent);
 
+/**
```
ganeshmurthy: I am wondering if the function should take ownership of the passed in pointer. I feel more comfortable if the function makes its own copy. Just my 2 cents.

src/connection_manager.c
dskarbek: yes, the change here is to group it closer together with where it is being used and to make sure to free it before the CHECK() call happens.  Otherwise, CHECK might cause the method to be exited before the memory is freed.

src/dispatch_private.h
dskarbek: Yeah, I tried it that way also.  The advantage of this is that I can pass in NULL to cause the memory to be freed.  But, the copy in may be the better way.  Only issue, is if you look at the two places this is being use, one is passing strdup, so no problem there, the strdup could just happen inside the method.  The other one is passing the result of qd_entity_opt_string().  That also returns newly allocated memory, right?  So, I'd have to catch that return separately, pass it in and then free it.  And, we'd basically be doing a wasted extra alloc/free.  That didn't seem so nice.  So, that's why I tended toward having this internal method just manage the setting of the pointer and ensuring that prior pointee's are freed.  And, of course, that's why I  commented the method as taking ownership of the memory.  Let me know if you prefer the other approach.

src/router_config.c
dskarbek: Yeah, I agree this is not the most pretty solution.  But, wasn't sure how else to handle it.  Happy to hear about a nicer approach.

dskarbek:
@ganeshmurthy Ted had mentioned that he pulled my commits and was getting a seg-fault on exit.  I haven't been able to reproduce that, and can't see any change that would be causing that just by code inspection.  Can you confirm this issue and let me know how to reproduce?

ganeshmurthy:
I was able to successfully merge your code to my local master but ran into a compile issue.

In file included from /home/gmurthy/opensource/dispatch/include/qpid/dispatch/ctools.h:27:0,
                 from /home/gmurthy/opensource/dispatch/include/qpid/dispatch/container.h:35,
                 from /home/gmurthy/opensource/dispatch/src/dispatch_private.h:40,
                 from /home/gmurthy/opensource/dispatch/src/router_core/router_core_private.h:22,
                 from /home/gmurthy/opensource/dispatch/src/router_core/route_tables.c:20:
/home/gmurthy/opensource/dispatch/src/router_core/route_tables.c: In function ‘qdr_del_router_CT’:
/home/gmurthy/opensource/dispatch/src/router_core/route_tables.c:353:17: error: ‘qdr_node_t {aka struct qdr_node_t}’ has no member named ‘maskbit’; did you mean ‘mask_bit’?
     assert(rnode->maskbit == router_maskbit);

Fix is easy enough (missing an underscore in maskbit)
assert(rnode->mask_bit == router_maskbit);

After the fix, the code compiled and I was able to run all unit tests successfully. I did not see any segfaults.

dskarbek:
Hmm… I wonder how it compiled on my machine?

-Dan Skarbek

PR #109, opened 2016-11-03
initial attempt at dispatch-553

mgoulish:
Ted -- I don't know yet how to test the json stuff so I am flying blind there, and I do not yet have the entity-refresh-cache thing.  But I wanted to send you what I do have now... just because.


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -841,6 +841,46 @@
                     "type": "string",
                     "description": "Where to send log messages. Can be 'stderr', 'stdout', 'syslog' or a file name.",
                     "update": true
+                },
+                "default": {
```
ted-ross: These attributes should have names that are more descriptive and should suggest that they are counters.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -841,6 +841,46 @@
...
+                    "type": "integer",
+                    "description": "how many entries in the severity-level histogram at the DEFAULT level",
+                    "update": "true"
```
ted-ross: DEFAULT is not a level that is used in raising logs.  It is only used in setting the enabled log threshold.  In other words, it should not be counted as it will always be zero.

src/log.c
```diff
@@ -149,13 +149,16 @@ static log_sink_t* log_sink_lh(const char* name) {
 }
 
 
+typedef enum {DEFAULT, NONE, TRACE, DEBUG, INFO, NOTICE, WARNING, ERROR, CRITICAL, N_LEVELS} level_index_t;
```
ted-ross: DEFAULT and NONE are not used in raising logs.

src/log.c
```diff
@@ -355,6 +358,14 @@ bool qd_log_enabled(qd_log_source_t *source, qd_log_level_t level) {
 
 void qd_log_impl(qd_log_source_t *source, qd_log_level_t level, const char *file, int line, const char *fmt, ...)
 {
+    /*-----------------------------------------------
+      Count this log-event in this log's histogram
+      whether or not this log is currently enabled.
+      We can always decide not to look at it later,
+      based on its used/unused status.
+    -----------------------------------------------*/
+    source->severity_histogram [ level ] ++;
```
ted-ross: You have a type problem here.  qd_log_level_t can't be used as an index into an array since it is a bitmask.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -841,6 +841,46 @@
...
+                },
+                "trace": {
+                    "type": "integer",
+                    "description": "how many entries in the severity-level histogram at the TRACE level",
+                    "update": "true"
```
ted-ross: You shouldn't use update:true here.  This would allow a remote user to update the value of the counter.  This is a read-only quantity.

PR #115, opened 2016-11-18
DISPATCH-553   

mgoulish:


src/log.c
```diff
@@ -149,6 +153,8 @@ static log_sink_t* log_sink_lh(const char* name) {
 }
 
 
+typedef enum {DEFAULT, NONE, TRACE, DEBUG, INFO, NOTICE, WARNING, ERROR, CRITICAL, N_LEVELS} level_index_t;
```
ted-ross: Why are there histograms for DEFAULT and NONE?  These will never be raised in an actual log.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -845,6 +845,45 @@
             }
         },
 
+        "logStats": {
+            "description": "histogram of the different severity-levels of events on the given log.",
+            "extends": "operationalEntity",
+            "attributes": {
+                "id": {
```
ted-ross: Each entity has a built-in "name" and "identity" attribute.  You should use these instead of adding a new one.

PR #117, opened 2016-11-23
DISPATCH-113 -- expose topology_last_changed to management

mgoulish:


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1209,7 +1209,12 @@
                 "cost": {
                     "description": "Reachability cost",
                     "type": "integer"
+                },
+                "lastTopoChange": {
+                     "description": "Timestamp showing the most recent change to this node's neighborhood.",
+                     "type": "string"
```
ted-ross: This should be an integer, not a string.  Seconds since the epoch is more than sufficient granularity.

PR #119, opened 2016-12-02
DISPATCH-583 - Fixed lifecycle management of proton links and session…

ganeshmurthy:
…s. These are now freed after the collector is done with all the events

### DIFF
--- a/include/qpid/dispatch/server.h
+++ b/include/qpid/dispatch/server.h
@@ -477,6 +477,15 @@ typedef int (*qd_pn_event_handler_cb_t)(void *handler_context, void* conn_contex
 
 
 /**
+ * Post event process handler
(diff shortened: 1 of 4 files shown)


include/qpid/dispatch/server.h
```diff
@@ -477,6 +477,15 @@ typedef int (*qd_pn_event_handler_cb_t)(void *handler_context, void* conn_contex
 
 
 /**
+ * Post event process handler
+ * Invoke only after all proton events have been popped from the collector.
+ *
+ * @param conn The connection for which all proton events have been popped.
+ */
+typedef void (*qd_pn_event_complete_cb_t)(qd_connection_t *conn);
```
ted-ross: All of the callback's need to include the context argument (even it you don't use it in your handler).

PR #123, opened 2016-12-08
DISPATCH-209 : anonymous sender test

mgoulish:
This test started working as I originally expected it to, after I recreated this branch from new upstream code.


mgoulish:
This is not working like I thought.  It was using the address from the previous test.  When I corrected that, it no workee.   Looking into it...

PR #124, opened 2016-12-09
Allow passing the password via env variable

dskarbek:
Allows for the password in the config to actually be the name of the environment variable that contains the password.  This is indicated by prefixing the value with "env:".  We also add the option to prefix with "literal:" which indicates that it is a raw password.  That way, you can still give "env:SHELL" as a password, if that's what you really want to use, you would just have to list it as "literal:env:SHELL" in the config.

### DIFF
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -176,6 +176,46 @@ static void set_config_host(qd_server_config_t *config, qd_entity_t* entity)
     assert(config->host);
 }
 
+static char* qd_config_ssl_profile_get_password(qd_config_ssl_profile_t* ssl_profile)
+{
+    char *pw = ssl_profile->ssl_password;
+    if (!pw) return pw;
+
+    /* if the "password" starts with "env:" or "env: " then the remaining
+     * text is the environment variable that contains the password
+     */
+    if (strncmp(pw, "env:", 4) == 0) {
+        char *env = pw + 4;
+        /* skip the space if it is there */
+        if (*env == ' ') ++env;
+
+        const char* passwd = getenv(env);
+        if (passwd) {
+            free(ssl_profile->ssl_password);
+            pw = ssl_profile->ssl_password = strdup(passwd);
+        } else {
+            qd_error(QD_ERROR_NOT_FOUND, "Failed to find a password in the environment variable '%s'", env);
+        }
+        return pw;
+    }
+
+    /* if the "password" starts with "literal:" or "literal: " then
+     * the remaining text is the password and the heading should be
+     * stripped off
+     */
+    if (strncmp(pw, "literal:", 8) == 0) {
+        /* skip the "literal:" header */
+        pw += 8;
+        /* skip the space if it is there */
+        if (*pw == ' ') ++pw;
+        /* return a pointer into the middle of the string where the literal password starts */
+        return pw;
+    }
+
+    /* default case: return the value we found in the config as-is */
+    return pw;
+}
+
 static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *config, qd_entity_t* entity, qd_config_ssl_profile_t **ssl_profile)
 {
     qd_error_clear();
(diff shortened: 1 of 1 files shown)


src/connection_manager.c
```diff
@@ -176,6 +176,46 @@ static void set_config_host(qd_server_config_t *config, qd_entity_t* entity)
...
+        char *env = pw + 4;
+        /* skip the space if it is there */
+        if (*env == ' ') ++env;
+
+        const char* passwd = getenv(env);
+        if (passwd) {
+            free(ssl_profile->ssl_password);
+            pw = ssl_profile->ssl_password = strdup(passwd);
+        } else {
+            qd_error(QD_ERROR_NOT_FOUND, "Failed to find a password in the environment variable '%s'", env);
```
dskarbek: or, would you rather just return the password string as is and not register an error in-case it is an actual password?  If it is referring to an env var that missed getting set there's next to zero chance it will work, so better to catch the error sooner at the source so we can log it appropriately, but this does mean that someone using "env:foobar" as a password will suddenly stop working.

src/connection_manager.c
```diff
@@ -176,6 +176,46 @@ static void set_config_host(qd_server_config_t *config, qd_entity_t* entity)
...
+     * the remaining text is the password and the heading should be
+     * stripped off
+     */
+    if (strncmp(pw, "literal:", 8) == 0) {
+        /* skip the "literal:" header */
+        pw += 8;
+        /* skip the space if it is there */
+        if (*pw == ' ') ++pw;
+        /* return a pointer into the middle of the string where the literal password starts */
+        return pw;
```
dskarbek: Not a problem unless I'm wrong about the memory management of this pointer.

src/connection_manager.c
```diff
@@ -176,6 +176,46 @@ static void set_config_host(qd_server_config_t *config, qd_entity_t* entity)
...
+     */
+    if (strncmp(pw, "env:", 4) == 0) {
+        char *env = pw + 4;
+        /* skip the space if it is there */
+        if (*env == ' ') ++env;
+
+        const char* passwd = getenv(env);
+        if (passwd) {
+            free(ssl_profile->ssl_password);
+            pw = ssl_profile->ssl_password = strdup(passwd);
```
ted-ross: I think your memory management is correct.  This code actually replaces the "env:VAR" text with (a copy of) the contents of the environment variable on the first reference to the ssl_profile.  Subsequent references will then see the contents and won't re-query the environment.
There is one subtle issue with this.  If the password from the environment begins with "env:", the second reference to the ssl_profile will interpret the password as another environment reference.  To prevent this, the copy of the password taken from the environment should be prefixed with "literal:".  Alternatively, you could leave the ssl_profile unmodified and simply return the password from the environment.

src/connection_manager.c
ted-ross: This should definitely be logged.  It might be prudent to not include the password text in the log though.

src/connection_manager.c
dskarbek: Yes, this is replacing the value, and true, could cause problems if the literal value started with "env:".  Actually, the fix you mention would not be so great.  On the first time through the "literal" may or may not be handled depending on a subtle change in the code, and on the 3rd time, might have trouble or not.  I think that the "env:" and "literal:" keywords are supposed to be in-effect on the config file value, and should not apply to the env-var value.  That keeps it cleaner and simpler.  So, that would mean that I shouldn't re-write the ssl_profile value.  What do you think of just assigning the result of getenv() directly?  Can we change the ssl_password member to be a const char*?  Otherwise, I'd have to cast away const.

src/connection_manager.c
dskarbek: Oh, you mean for the case where that isn't an env variable, but actually is part of a password?  Hmm, that's true.  though, it seems like it would help you understand what is happening faster if we log it.  Is the log file less secure than the config file?  Cuz, in that case, they already have the password in the config file.

ted-ross:
Dan,

I committed a modified version of this patch yesterday.  Instead of 
processing the password at connector/listener setup time, it does it at 
ssl_profile setup time.  This ensures that the password is interpreted 
only once, eliminating the possible issues of re-processing.

-Ted

dskarbek:
Ok, that makes sense.  Thanks.

PR #125, opened 2016-12-12
DISPATCH-557 - Moved management of connections from python agent to r…

ganeshmurthy:
…outer core

### DIFF
--- a/include/qpid/dispatch/router_core.h
+++ b/include/qpid/dispatch/router_core.h
@@ -23,6 +23,7 @@
 #include <qpid/dispatch/bitmask.h>
 #include <qpid/dispatch/compose.h>
 #include <qpid/dispatch/parse.h>
+typedef struct qdr_core_t         qdr_core_t;
 #include <qpid/dispatch/router.h>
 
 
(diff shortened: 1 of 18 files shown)


include/qpid/dispatch/router_core.h
```diff
@@ -23,6 +23,7 @@
 #include <qpid/dispatch/bitmask.h>
 #include <qpid/dispatch/compose.h>
 #include <qpid/dispatch/parse.h>
+typedef struct qdr_core_t         qdr_core_t;
```
ted-ross: Please move this back down with the other typedefs.  You can add this line into router.h to solve the compilation issue.

src/router_core/agent_connection.c
```diff
@@ -0,0 +1,344 @@
...
+
+    case QDR_CONNECTION_CONTAINER_ID:
+        if (conn->connection_info->container)
+            qd_compose_insert_string(body, conn->connection_info->container);
+        else
+            qd_compose_insert_null(body);
+        break;
+
+    case QDR_CONNECTION_SASL_MECHANISMS:
+        qd_compose_insert_string(body, conn->connection_info->sasl_mechanisms);
```
ted-ross: sasl_mechanisms is 0 if not specified.  That causes a crash here (easily reproducible by starting the web console).  If it's not present, this should insert a NULL into the body.

src/router_core/connections.c
```diff
@@ -107,6 +130,61 @@ void qdr_connection_set_context(qdr_connection_t *conn, void *context)
         conn->user_context = context;
 }
 
+qdr_connection_info_t *qdr_connection_info(bool             is_encrypted,
+                                           bool             is_authenticated,
+                                           bool             opened,
+                                           char            *sasl_mechanisms,
+                                           const char      *dir,
```
ted-ross: There is an enumerated type for direction.  You should use this rather than passing strings around.

src/router_core/router_core_private.h
```diff
@@ -439,10 +439,28 @@ typedef struct qdr_connection_work_t {
...
+    const  char                *container;
+    char                       *sasl_mechanisms;
+    char                       *host;
+    bool                        is_encrypted;
+    char                       *ssl_proto;
+    char                       *ssl_cipher;
+    char                       *user;
+    bool                        is_authenticated;
+    bool                        opened;
+    const char                 *dir;
```
ted-ross: I think dir and role should use the enumerated types rather than arbitrary strings.

PR #127, opened 2016-12-15
DISPATCH-596 - Preserve the error from the rejected state in relayed …

ganeshmurthy:
…disposition


src/router_node.c
```diff
@@ -947,6 +964,9 @@ static void CORE_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t di
         pn_delivery_settle(pnd);
         qdr_delivery_decref(dlv);
     }
+
+    if (error)
```
ted-ross: When you free the error record here, the pointer (now invalid) remains in the delivery record.  It could be referenced and possibly re-freed in the future.  I think the error record should be freed when the delivery record is freed.

PR #134, opened 2016-12-21
DISPATCH-209 : dynamic request response test

mgoulish:
phew.

### DIFF
--- a/tests/system_tests_three_routers.py
+++ b/tests/system_tests_three_routers.py
@@ -230,7 +230,7 @@ class DynamicRequestResponseTest(MessagingHandler):
         super(DynamicRequestResponseTest, self).__init__(prefetch=0)
         self.address1   = address1
         self.address2   = address2
-        self.dest       = "closest.whatever"
+        self.dest       = "closest.dynamicRequestResponse"
         self.error      = None
         self.sender     = None
         self.server_receiver = None
--- a/tests/system_tests_three_routers.py
+++ b/tests/system_tests_three_routers.py
@@ -242,31 +242,27 @@ class DynamicRequestResponseTest(MessagingHandler):
  
     def timeout(self):
         self.error = "Timeout Expired %d messages received." % self.n_received
-        self.conn1.close()
-        self.conn2.close()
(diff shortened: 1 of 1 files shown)


tests/system_tests_three_routers.py
```diff
@@ -220,6 +225,71 @@ def run(self):
...
+        self.conn1.close()
+        self.conn2.close()
+ 
+    def on_released(self, event):
+        self.n_sent -= 1
+        time.sleep(0.1)
+
+    def on_link_opened(self, event):
+        if event.receiver:
+            self.sender = event.container.create_sender(self.conn1, None)
```
ted-ross: You are going to get two on_link_opened events for receivers.  This means you are creating the same sender twice.  This one sender is used for both client-send and server-send operations.

tests/system_tests_three_routers.py
```diff
@@ -220,6 +225,71 @@ def run(self):
...
+
+    def on_start(self, event):
+        self.timer = event.reactor.schedule(5, Timeout(self))
+        self.conn1 = event.container.connect(self.address1)
+        self.conn2 = event.container.connect(self.address2)
+        self.server_receiver = event.container.create_receiver(self.conn2, self.dest)
+        self.client_receiver = event.container.create_receiver(self.conn2, None, dynamic=True)
```
ted-ross: Both the client and server receivers are established on the same connection.  Is this what was intended?  I would think that one connection would be the client and the other the server.

tests/system_tests_three_routers.py
```diff
@@ -220,6 +225,70 @@ def run(self):
...
+            self.n_sent += 1
+ 
+    def on_accepted(self, event):
+        self.n_accepted += 1
+ 
+    def on_message(self, event):
+
+        # Receiver gets a message and replies to client.
+        if event.receiver == self.server_receiver :
+            self.sender.send ( Message(address=event.message.reply_to, body="Reply hazy, try again later.") )
```
ted-ross: The server is sending its responses on the client's sender!

PR #138, opened 2017-01-19
Pre proactor cleanup

alanconway:
Purely clean-up: deleting dead code and making needlessly public declarations more private to simplify replacing the existing driver  with the proton proactor.


alanconway:
Needs  more testing, seeing some failures.

alanconway:
Fixed, had deleted on function called from python.

PR #142, opened 2017-02-03
DISPATCH-628: Multiple connections per autoLink and linkRoute overwrites connection handle

lulf:
This ensures that an autolink and linkroute can be bound to more than one
connection. The assert checking this is disabled in optimized builds and was
never triggered.

### DIFF
--- a/src/router_core/route_control.c
+++ b/src/router_core/route_control.c
@@ -53,7 +53,7 @@ static void qdr_route_check_id_for_deletion_CT(qdr_core_t *core, qdr_conn_identi
     // If this connection identifier has no open connection and no referencing routes,
     // it can safely be deleted and removed from the hash index.
     //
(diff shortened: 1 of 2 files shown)


src/router_core/route_control.c
```diff
@@ -396,7 +408,18 @@ void qdr_route_connection_closed_CT(qdr_core_t *core, qdr_connection_t *conn)
...
         }
 
-        cid->open_connection = 0;
+        //
+        // Remove our own entry in the connection list
+        //
+        qdr_connection_ref_t * cref = DEQ_HEAD(cid->connection_refs);
+        while (cref) {
+            if (cref->conn == conn) {
+                DEQ_REMOVE(cid->connection_refs, cref);
```
ted-ross: I believe this will leak one qdr_connection_ref_t each time it is run.  Why did you not use qdr_del_connection_ref?

lulf: Agreed, I think I didn't know about that one. Just pushed an update.

PR #143, opened 2017-02-27
DISPATCH-597 - Added ability to selectively log components of a message

ganeshmurthy:


src/router_node.c
```diff
@@ -226,9 +227,22 @@ static void AMQP_rx_handler(void* context, qd_link_t *link, pn_delivery_t *pnd)
     //        send it.
     //
     msg = qd_message_receive(pnd);
+
     if (!msg)
         return;
 
+
+    char repr[qd_message_repr_len()];
```
ted-ross: This declaration should be inside the if statement.  You are needlessly allocating stack space when logging is disabled.

src/router_node.c
```diff
@@ -226,9 +227,22 @@ static void AMQP_rx_handler(void* context, qd_link_t *link, pn_delivery_t *pnd)
...
+
+    if (conn->log_message) {
+        qd_log(qd_message_log_source(), QD_LOG_TRACE, "Received %s on link %s",
```
ted-ross: I think the log would look better if the link id was at the beginning and the message repr was at the end.

include/qpid/dispatch/server.h
```diff
@@ -451,6 +452,28 @@ typedef struct qd_server_config_t {
...
+     *   content-type
+     *   content-encoding
+     *   absolute-expiry-time
+     *   creation-time
+     *   group-id
+     *   group-sequence
+     *   reply-to-group-id
+     *   app-properties.
+     */
+    char *log_message;
```
ted-ross: I would recommend storing this in its bit-mask form here.  Then it only needs to be parsed once at configuration time.  This also allows you to isolate the string definitions and the conversion functions to a single .h/.c module.

src/server.c
```diff
@@ -146,6 +148,26 @@ char *COMPONENT_SEPARATOR = ";";
 
 const char *DEFAULT_USER_ID = "anonymous";
 
+const char *qd_log_message_components[] =
```
ted-ross: You should move this out of server.c where it doesn't really belong.  Consider adding a message_trace.[ch] module for this logic.

src/server.c
```diff
@@ -1634,7 +1710,9 @@ qd_listener_t *qd_server_listen(qd_dispatch_t *qd, const qd_server_config_t *con
     li->server      = qd_server;
     li->config      = config;
     li->context     = context;
+    li->log_message = populate_log_message(config);
```
ted-ross: It's confusing that "log_message" sometimes means a character string and sometimes means a bit-mask.

PR #144, opened 2017-02-27
DISPATCH-640 - Added code to enable the same connection object to be …

ganeshmurthy:
…obtained using a containerId and a connectionId


ted-ross:
This change looks good to me.

PR #149, opened 2017-03-21
DISPATCH-357 - Added terminus_addr field on link to display the termi…

ganeshmurthy:
…nus addr of a link route


src/router_core/connections.c
```diff
@@ -1297,6 +1320,7 @@ static void qdr_link_inbound_first_attach_CT(qdr_core_t *core, qdr_action_t *act
                     //
                     // This is a link-routed destination, forward the attach to the next hop
                     //
+                    set_terminus_address(link, dir);
```
ted-ross: You are calling a non-CT function from within a CT function.  The non-CT function accesses pn_link_t structures that are unsafe to access from a thread that is not the IO thread that owns the connection on which the link flows.  This is very likely to cause memory corruption and crashes.

PR #150, opened 2017-03-21
DISPATCH-357 - Added terminus_addr field on link to display the termi…

ganeshmurthy:
…nus addr of a link route


src/router_core/forwarder.c
```diff
@@ -704,6 +704,16 @@ bool qdr_forward_link_balanced_CT(qdr_core_t     *core,
         out_link->link_type      = QD_LINK_ENDPOINT;
         out_link->link_direction = qdr_link_direction(in_link) == QD_OUTGOING ? QD_INCOMING : QD_OUTGOING;
         out_link->admin_enabled  = true;
+        if (in_link->terminus_addr) {
```
ted-ross: Why is this necessary?  It looks like you are ensuring that both connected links have their own copy of the terminus name.  However, in the agent code, you get the terminus name from either the direct link or the connected link.  Isn't it sufficient to have the terminus name in only one of the connected links?

PR #160, opened 2017-04-22
DISPATCH-749 - unmap_all_addresses unmaps just half of addresses

pmoravec:
iterating over a list being updated needs deep copy of the list

Signed-off-by: Pavel Moravec <pmoravec@redhat.com>

### DIFF
--- a/python/qpid_dispatch_internal/router/node.py
+++ b/python/qpid_dispatch_internal/router/node.py
@@ -533,8 +533,8 @@ class RouterNode(object):
 
     def unmap_all_addresses(self):
         self.mobile_address_sequence = 0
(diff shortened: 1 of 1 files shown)


pmoravec:
This approach (deep copy of the list and calling `.remove`) might be inefficient for long lists. If this can be an issue, some different approach shall be used, e.g.:

    while len(self.mobile_addresses)>0:
        self.unmap_address(self.mobile_addresses[0])

ganeshmurthy:
Can we do this instead

`while self.mobile_addresses: 
    self.unmap_address(self.mobile_addresses.pop(0))
`

pmoravec:
Rebased the PR to more efficient fix.

PR #162, opened 2017-05-10
DISPATCH-209 -- revamp of second three-router test.

mgoulish:
This new version of three-router test #2 survived a 1000-iteration test with no failures.

The main changes are

  1. handle on_sendable event and the sending of messages better.  (Before I was only sending 1 message per on_sendable event, instead of sending until credit exhausteed.)

  2. indirectly detect when the network is not yet ready for me to send (in spite of on_sendable event) and pause for 1 second to allow it to propagate endpoint info everywhere.

### DIFF
--- a/tests/system_tests_three_routers.py
+++ b/tests/system_tests_three_routers.py
@@ -39,56 +39,113 @@ class RouterTest(TestCase):
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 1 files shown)


mgoulish:
closing to replace with new PR containing 3rd test.

PR #165, opened 2017-05-30
DISPATCH-209 : add a linear 3-router linkroute test

mgoulish:
test #4 -- 3-router link-route

### DIFF
--- a/console/hawtio/src/main/webapp/plugin/js/qdrList.js
+++ b/console/hawtio/src/main/webapp/plugin/js/qdrList.js
@@ -100,8 +100,8 @@ var QDR = (function(QDR) {
           $scope.fetchingLog = false;
           var statusCode = context.message.application_properties.statusCode;
           if (statusCode < 200 || statusCode >= 300) {
-            Core.notification('error', context.message.application_properties.statusDescription);
-            //QDR.log.debug(context.message.application_properties.statusDescription)
+            Core.notification('error', context.message.statusDescription);
+            //QDR.log.debug(context.message.statusDescription)
             return;
           }
           $scope.logResults = response.filter( function (entry) {
@@ -639,8 +639,8 @@ var QDR = (function(QDR) {
     function gotMethodResponse (nodeName, entity, response, context) {
       var statusCode = context.message.application_properties.statusCode;
       if (statusCode < 200 || statusCode >= 300) {
-        Core.notification('error', context.message.application_properties.statusDescription);
-        //QDR.log.debug(context.message.application_properties.statusDescription)
+        Core.notification('error', context.message.statusDescription);
+        //QDR.log.debug(context.message.statusDescription)
       } else {
         var note = entity + " " + $filter('Pascalcase')($scope.currentMode.op) + "d"
         Core.notification('success', note);
--- a/console/hawtio/src/main/webapp/plugin/js/qdrOverviewLogsController.js
+++ b/console/hawtio/src/main/webapp/plugin/js/qdrOverviewLogsController.js
@@ -26,7 +26,7 @@ var QDR = (function(QDR) {
       var gotLogInfo = function (nodeId, entity, response, context) {
         var statusCode = context.message.application_properties.statusCode;
         if (statusCode < 200 || statusCode >= 300) {
-          Core.notification('error', context.message.application_properties.statusDescription);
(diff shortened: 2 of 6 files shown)


mgoulish:
I had an 8-in-100 failure rate suddenly appear on test #3 -- so... I made on_sendable handling smarter and that gave me 4-in-100 failure rate -- all on teardown only.    I then changed a little bit how all four tests shut down -- borrowing an interesting line from another test (null out the timer after clearing it) -- and now I have  0 failures in 400 iterations -- also no teardown failures.

mgoulish:
WAIT

I just had a failure on my tests while running them in the context of a full "make check" -- investigating.....

tests/system_tests_three_routers.py
```diff
@@ -96,10 +124,24 @@ def router ( name, connection_1, connection_2=None ):
                          'port': inter_router_port_A,
                          'verifyHostName': 'no'
                        }
+                   ),
+                   ( 'linkRoute',
+                     { 'prefix': cls.linkroute_prefix,
+                       'containerId': 'QDR.A',
```
ted-ross: These containerId lines have no effect and are misleading.

tests/system_tests_three_routers.py
```diff
@@ -335,36 +417,124 @@ def on_start ( self, event ):
...
+    def __init__ ( self, addr_listening_for_route_container, linkroute_prefix, qdstat_address, send_address ):
+        super(LinkRoute, self).__init__(prefetch=0)
+        self.addr_listening_for_route_container = addr_listening_for_route_container
+        self.send_address = send_address
+        self.linkroute_prefix = linkroute_prefix
+        self.qdstat_address = qdstat_address
+        self.error      = None
+        self.send_connection = None
+        self.recv_connection = None
+        self.false_statement = "This is not the message you are looking for."
```
ted-ross: For anyone trying to understand how this test works, this text is misleading.

tests/system_tests_three_routers.py
```diff
@@ -335,36 +417,124 @@ def on_start ( self, event ):
...
+
+
+    def on_start(self, event):
+        # Iff this timer expires, the test fails.
+        self.timer = event.reactor.schedule ( 5, Timeout(self) )
+
+        # Create the receiver only at start time.
+        # Sender-creation has to wait until we know that the address is available.
+        self.recv_connection = event.container.connect ( self.addr_listening_for_route_container )
+        self.receiver = event.container.create_receiver ( self.recv_connection, None, dynamic=True )
```
ted-ross: Why is this receiver here?  It serves no purpose in the test.

tests/system_tests_three_routers.py
```diff
@@ -335,36 +417,124 @@ def on_start ( self, event ):
...
+
+
+    def on_link_opening ( self, event ):
+        if event.receiver:
+            event.receiver.flow(1)
+
+
+    def on_connection_remote_open ( self, event ):
```
ted-ross: Doing this blocking, synchronous operation in a reactive event handler is not a good practice.  The reactor is prevented from maintaining its connections for the entire duration of this function.
There is an example in system_tests_multi_tenancy (LinkRouteTest) that does what's needed to wait for the link-route to propagate within the reactive framework.

tests/system_tests_three_routers.py
```diff
@@ -335,36 +417,124 @@ def on_start ( self, event ):
...
+        # Iff this timer expires, the test fails.
+        self.timer = event.reactor.schedule ( 5, Timeout(self) )
+
+        # Create the receiver only at start time.
+        # Sender-creation has to wait until we know that the address is available.
+        self.recv_connection = event.container.connect ( self.addr_listening_for_route_container )
+        self.receiver = event.container.create_receiver ( self.recv_connection, None, dynamic=True )
+
+
+    def on_link_opening ( self, event ):
```
ted-ross: When handing an unsolicited incoming link-attach, this handler must copy the addresses from the remote to the local.  If you trace the operation of this test, you will see that the returning attach (from the initial attach to "myLinkRoute") does not contain an address.  This is a protocol violation and it is by luck that the link correctly transfers the message.
The system_tests_multi_tenancy/LinkRouteTest test can be used an an example of the correct way to handle a link-routed attach.

PR #167, opened 2017-06-19
DISPATCH-788 - First attempt at adding a core thread API around handl…

ganeshmurthy:
…ing of peers. The next step is to introduce a list to hold more than one peer

### DIFF
--- a/src/router_core/forwarder.c
+++ b/src/router_core/forwarder.c
@@ -101,33 +101,26 @@ static void qdr_forward_find_closest_remotes_CT(qdr_core_t *core, qdr_address_t 
 
 qdr_delivery_t *qdr_forward_new_delivery_CT(qdr_core_t *core, qdr_delivery_t *in_dlv, qdr_link_t *link, qd_message_t *msg)
 {
-    qdr_delivery_t *dlv = new_qdr_delivery_t();
-    uint64_t       *tag = (uint64_t*) dlv->tag;
+    qdr_delivery_t *out_dlv = new_qdr_delivery_t();
+    uint64_t       *tag = (uint64_t*) out_dlv->tag;
 
-    ZERO(dlv);
-    sys_atomic_init(&dlv->ref_count, 0);
-    dlv->link       = link;
-    dlv->msg        = qd_message_copy(msg);
-    dlv->settled    = !in_dlv || in_dlv->settled;
-    dlv->presettled = dlv->settled;
+    ZERO(out_dlv);
+    sys_atomic_init(&out_dlv->ref_count, 0);
+    out_dlv->link       = link;
+    out_dlv->msg        = qd_message_copy(msg);
+    out_dlv->settled    = !in_dlv || in_dlv->settled;
+    out_dlv->presettled = out_dlv->settled;
     *tag            = core->next_tag++;
-    dlv->tag_length = 8;
-    dlv->error      = 0;
+    out_dlv->tag_length = 8;
+    out_dlv->error      = 0;
 
     //
-    // Create peer linkage only if the delivery is not settled
+    // Create peer linkage only if the outgoing delivery is not settled
     //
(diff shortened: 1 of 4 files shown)


src/router_core/transfer.c
```diff
@@ -413,6 +413,43 @@ static void qdr_delete_delivery_internal_CT(qdr_core_t *core, qdr_delivery_t *de
 }
 
 
+void qdr_delivery_link_peers_CT(qdr_delivery_t *in_dlv, qdr_delivery_t *out_dlv)
+{
+    if (in_dlv->peer || out_dlv->peer)
+        // One of the deliveries already has a peer. Don't proceed.
+        return;
```
ted-ross: You should assert here.  This must never happen.

src/router_core/transfer.c
```diff
@@ -413,6 +413,43 @@ static void qdr_delete_delivery_internal_CT(qdr_core_t *core, qdr_delivery_t *de
...
+    qdr_delivery_incref(in_dlv);
+}
+
+
+void qdr_delivery_unlink_peers_CT(qdr_core_t *core, qdr_delivery_t *dlv, qdr_delivery_t *peer)
+{
+    dlv->peer  = 0;
+    peer->peer = 0;
+
+    qdr_delivery_decref_CT(core, dlv);
```
ted-ross: What if these two deliveries weren't linked to one another?  You will cause a data corruption with no indication.  You should assert that the proper mutual linkage is in place before removing it and decrementing reference counts.

src/router_core/transfer.c
```diff
@@ -671,7 +708,8 @@ static void qdr_send_to_CT(qdr_core_t *core, qdr_action_t *action, bool discard)
 static void qdr_update_delivery_CT(qdr_core_t *core, qdr_action_t *action, bool discard)
 {
     qdr_delivery_t *dlv        = action->args.delivery.delivery;
-    qdr_delivery_t *peer       = dlv->peer;
+    qdr_delivery_start_peer_CT(dlv);
```
ted-ross: Now that I'm looking at it, a better API might be: qdr_delivery_t qdr_delivery_first_peer_CT(dlv) instead of start_peer.

src/router_core/forwarder.c
```diff
@@ -101,33 +101,26 @@ static void qdr_forward_find_closest_remotes_CT(qdr_core_t *core, qdr_address_t
...
-    dlv->error      = 0;
+    out_dlv->tag_length = 8;
+    out_dlv->error      = 0;
 
     //
-    // Create peer linkage only if the delivery is not settled
+    // Create peer linkage only if the outgoing delivery is not settled
     //
-    if (!dlv->settled) {
-        if (in_dlv && in_dlv->peer == 0) {
```
ted-ross: You've lost the in_dlv->peer == 0 check

ganeshmurthy: I moved this check (in_dlv->peer == 0) into the qdr_delivery_link_peers_CT, so we are good there.

src/router_core/router_core_private.h
```diff
@@ -693,6 +694,28 @@ void qdr_delivery_release_CT(qdr_core_t *core, qdr_delivery_t *delivery);
...
+ */
+void qdr_delivery_link_peers_CT(qdr_delivery_t *in_dlv, qdr_delivery_t *out_dlv);
+
+/**
+ * Zeroes out peer references from both peers and decrefs ref counts.
+ */
+void qdr_delivery_unlink_peers_CT(qdr_core_t *core, qdr_delivery_t *dlv, qdr_delivery_t *peer);
+
+/**
+ * Called before the call to qdr_delivery_next_peer_CT. This positions the peer pointer to the first peer position.
```
ted-ross: The comments for qdr_delivery_{first,next}_peer are incorrect.

PR #168, opened 2017-06-20
crolke dev annotations - work in progress PR for review

ChugR:
**DO NOT COMMIT THIS PULL REQUEST** It is for review only.

This work adds a new annotation scheme for inter-router annotations. The old method put a variable number (0..4) of key-value pairs into the message annotations map at the end of the map. The new method puts in a single list of four elements at the beginning of the message annotations map.

In order to be backwards compatible there is some amount of version sensing and translating between annotation formats. This could be improved in a number of ways.

This commit still fails two self tests: system_tests_one_router and system_tests_two_routers in their annotation handling sections. The V1 vs V2 annotation feature is not tested much at all. That is in the works after the general strategy is approved.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ include_directories(
     )
 
 if (NOT CMAKE_SYSTEM_NAME STREQUAL SunOS)
- add_compile_options(-Werror)
+ #add_compile_options(-Werror)
  add_compile_options(-Wall)
  include(CheckCCompilerFlag)
  check_c_compiler_flag(-Wpedantic HAS_PEDANTIC_FLAG)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ include_directories(
     )
 
 if (NOT CMAKE_SYSTEM_NAME STREQUAL SunOS)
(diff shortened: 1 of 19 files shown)


ganeshmurthy:
@ChugR , I downloaded this PR and I see the following code repeated twice in qd_message_send() in message.c

    //
    // Send delivery annotation if present
    //
    if (content->section_delivery_annotation.length > 0) {
        buf    = content->section_delivery_annotation.buffer;
        cursor = content->section_delivery_annotation.offset + qd_buffer_base(buf);
        advance(&cursor, &buf,
                content->section_delivery_annotation.length + content->section_delivery_annotation.hdr_length,
                send_handler, (void*) pnl);
    }

ChugR:
Well spotted. That's a mistake from a failed merge conflict. Will fix.

ChugR:
The straight new V2 method offers about a 35% message rate improvement over the existing method that shipped in 0.8.x. However, the new way of doing the original method offers a 25% improvement. I'm withdrawing this PR and working on better processing of "the old way". 

Keeping the old annotations eliminates incompatible versions V1 vs V2, reduces code clutter, eliminates peer version detection, and simplifies testing.

PR #169, opened 2017-06-20
DISPATCH-788 - Part 2 - Added a list of peer references in case there…

ganeshmurthy:
… is more than one peer. This will help in the future when we introduce large message streaming

### DIFF
--- a/src/router_core/transfer.c
+++ b/src/router_core/transfer.c
@@ -412,6 +412,13 @@ static void qdr_delete_delivery_internal_CT(qdr_core_t *core, qdr_delivery_t *de
     free_qdr_delivery_t(delivery);
 }
 
+static bool qdr_delivery_has_peer_CT(qdr_delivery_t *dlv)
+{
+    if (!dlv->peer && DEQ_SIZE(dlv->peers) == 0)
+        return false;
+    return true;
+}
+
 
 void qdr_delivery_link_peers_CT(qdr_delivery_t *in_dlv, qdr_delivery_t *out_dlv)
 {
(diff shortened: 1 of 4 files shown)


src/router_core/transfer.c
```diff
@@ -412,14 +412,41 @@ static void qdr_delete_delivery_internal_CT(qdr_core_t *core, qdr_delivery_t *de
     free_qdr_delivery_t(delivery);
 }
 
+static bool qdr_delivery_has_peer_CT(qdr_delivery_t *dlv)
+{
+    if (!dlv->peer && DEQ_SIZE(dlv->peers) == 0)
```
ted-ross: You could do this more elegantly like this:
    return dlv->peer || DEQ_SIZE(dlv->peers) > 0;

src/router_core/transfer.c
```diff
@@ -412,14 +412,41 @@ static void qdr_delete_delivery_internal_CT(qdr_core_t *core, qdr_delivery_t *de
...
+            // This is the first time we know that in_dlv is going to have more than one peer.
+            // There is already a peer in the in_dlv->peer pointer, move it into a list and zero it out.
+            qdr_add_delivery_ref_CT(&in_dlv->peers, in_dlv->peer);
+
+            // Zero out the peer pointer. Since there is more than one peer, this peer has been moved to the "peers" linked list.
+            // All peers will now reside in the peers linked list. No need to decref/incref here because you are transferring ownership.
+            in_dlv->peer = 0;
+        }
+
+        out_dlv->peer = in_dlv;
```
ted-ross: Minor point:  You repeat this line.  Why not do this once outside of the if-then-else.  The condition does not affect the backward linkage from out to in.

PR #170, opened 2017-06-29
DISPATCH-209 : linkroute test better for proactor

mgoulish:
I stole code from LinkRoute test in multi tenancy system test -- to make my linkroute test better for proactor -- don't hog control so long while checking for linkroute address to appear at the far router.

All 4 tests survived a 100-iteration test on my machine.

### DIFF
--- a/tests/system_tests_three_routers.py
+++ b/tests/system_tests_three_routers.py
@@ -18,11 +18,14 @@
 #
 
 import unittest, os, json
-from subprocess import PIPE, STDOUT
-from proton import Message, PENDING, ACCEPTED, REJECTED, RELEASED, SSLDomain, SSLUnavailable, Timeout
-from system_test import TestCase, Qdrouterd, main_module, DIR, TIMEOUT, Process
+from subprocess      import PIPE, STDOUT
+from proton          import Message, PENDING, ACCEPTED, REJECTED, RELEASED, SSLDomain, SSLUnavailable, Timeout
+from system_test     import TestCase, Qdrouterd, main_module, DIR, TIMEOUT, Process
 from proton.handlers import MessagingHandler
-from proton.reactor import Container, AtMostOnce, AtLeastOnce, DynamicNodeProperties, LinkOption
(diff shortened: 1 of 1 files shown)


tests/system_tests_three_routers.py
```diff
@@ -373,5 +433,189 @@ def run(self):
...
+            event.receiver.flow ( self.count )
+
+        if event.receiver == self.route_check_receiver:
+            self.proxy = RouterProxy ( self.route_check_receiver.remote_source.address )
+            self.poll()
+
+
+    def on_link_opening ( self, event ):
+        if event.receiver:
+            event.receiver.target.address = self.route_container
```
ted-ross: self.route_container is a host address (0.0.0.0:<some-port>), not an AMQP address.  If you look at the trace of the test, you will see that the addresses in the paired attach frames are different.  This should not be the case.

tests/system_tests_three_routers.py
```diff
@@ -373,5 +433,189 @@ def run(self):
...
+
+        if event.receiver == self.route_check_receiver:
+            response = self.proxy.response(event.message)
+            if response.status_code == 200 and (response.remoteCount + response.containerCount) > 0:
+                if self.route_check_timer :
+                    self.route_check_timer.cancel()
+                    self.route_check_timer = None
+                # After address appears at the far end of my router line, (router A)
+                # give more time to let it finish propagating.
+                time.sleep ( 2 )
```
ted-ross: Is this sleep really necessary?  Why do you need to allow more time for propagation when you have just verified that the propagation is complete?

tests/system_tests_three_routers.py
```diff
@@ -373,5 +433,189 @@ def run(self):
...
+
+    def on_timer_task(self, event):
+        self.parent.timeout()
+
+
+class LinkRoute ( MessagingHandler ):
+
+    """
+        Set up and use a link-route, to send a message this way:
+            receiver <--- A <--- B <--- C <--- sender
```
ted-ross: More information is needed here.   Which endpoint is the initiator of the link?  Which endpoint is connected as the route_container?

mgoulish:
new one coming shortly...

PR #172, opened 2017-07-06
DISPATCH-767 - Added code to support multi frame message streaming

ganeshmurthy:


src/message.c
```diff
@@ -1151,89 +1287,140 @@ void qd_message_send(qd_message_t *in_msg,
...
+    // How many receivers does this message have?
+    int                  fanout   = qd_message_fanout(in_msg);
 
-    // Process  the message annotations if any
-    compose_message_annotations(msg, &new_ma, strip_annotations);
+    if (msg->sent_depth < QD_DEPTH_MESSAGE_ANNOTATIONS) {
+        //
+        // Start with the very first buffer;
+        //
+        buf = DEQ_HEAD(content->buffers);
```
alanconway: No lock here? What's the protocol for protecting buffers between send/receive threads? We can't let threads work concurrently on the same buffer (in particular change the buffer size) We could lock the use of buffers, not just the enqueue/deque, or there are some other options.

src/message.c
```diff
@@ -1151,89 +1287,140 @@ void qd_message_send(qd_message_t *in_msg,
...
-        buf = DEQ_NEXT(buf);
+        size_t buf_size = qd_buffer_size(buf);
+
+        // This will send the remaining data in the buffer if any.
+        pn_link_send(pnl, (char*) qd_buffer_at(buf, msg->cursor.offset), buf_size - msg->cursor.offset);
+
+        // If the entire message has been received,  there is no need to lock before sending because no one else is
+        // trying to modify the data structure.
+        if (!receive_complete)
+            sys_mutex_lock(msg->content->lock);
```
alanconway: I wouldn't bother with the conditional lock  - if there's no contention the cost of the lock is small, and the condition introduces one more way for a future programmer to screw up the thread safety logic by mistake.

src/router_core/forwarder.c
```diff
@@ -395,9 +415,22 @@ int qdr_forward_closest_CT(qdr_core_t      *core,
...
+        bool receive_complete = qd_message_receive_complete(msg);
         qdr_subscription_t *sub = DEQ_HEAD(addr->subscriptions);
         if (sub) {
-            qdr_forward_on_message_CT(core, sub, in_delivery ? in_delivery->link : 0, msg);
+
+            //
+            // Only if the message has been completely received, forward it.
+            // Subscriptions, at the moment, dont have the ability to deal with partial messages
+            //
+            if (receive_complete)
```
alanconway: For streaming we need to start forwarding before completion.

PR #176, opened 2017-07-13
DISPATCH 209 -- add 3-router test for  closest message distribution

mgoulish:
After I close the receivers -- but there are still probably messages in-flight -- Proton spews an error (which should probably be a warning) that some handle no longer exists.  I think it ought to be left this way, because this is the most likely way a user would close a receiver when he's done with it.

### DIFF
--- a/doc/man/qdmanage.8.adoc
+++ b/doc/man/qdmanage.8.adoc
@@ -31,10 +31,11 @@ SYNOPSIS
 
 DESCRIPTION
 -----------
-An AMQP management client for use with qdrouterd. Sends AMQP management
-operations requests and prints the response in JSON format. This is a
-generic AMQP management tool and can be used with any standard AMQP
-managed endpoint, not just with qdrouter.
+An AMQP management client for use with the Dispatch router daemon 
+('qdrouterd'). Sends AMQP management operations requests and prints 
+the response in JSON format. This is a generic AMQP management tool 
+and can be used with any standard AMQP managed endpoint, not just 
+with 'qdrouterd'.
 
 OPTIONS
 -------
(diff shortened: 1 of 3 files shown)


tests/system_tests_three_routers.py
```diff
@@ -624,5 +636,151 @@ def run(self):
...
+            # to the mid receivers.  At this point we should have
+            # no messages in the far receivers.
+            self.mid_recv_1.close()
+            self.mid_recv_2.close()
+            if self.far_messages > 0 :
+                self.bail ( "error: far receivers got messages before mid were closed." )
+
+        # Increment the near, mid, or far counts, depending on
+        # which receiver the message came in on.
+        if event.receiver   == self.near_recv_1 or event.receiver == self.near_recv_2:
```
ganeshmurthy: No messages are ever received by self.near_recv_2 because of lack for credit. Same with self.mid_recv_2 and self.far_recv_2. These 3 receivers should also send some flow so that they also receive message

tests/system_tests_three_routers.py
```diff
@@ -215,6 +218,15 @@ def test_04_linkroute ( self ):
         test.run()
         self.assertEqual(None, test.error)
 
+    def test_05_linkroute ( self ):
```
ganeshmurthy: The name of this test should not be test_05_linkroute because it is using message routing

ganeshmurthy:
The "no-such-handle" error comes from the following code in transport.c in a few places.

```
pn_link_t *link = pni_handle_state(ssn, handle);

  if (!link) {
    return pn_do_error(transport, "amqp:invalid-field", "no such handle: %u", handle);
  }
```

Looks like proton is trying to send a message to a receiver on a link that was closed abruptly. The router will try to redeliver this message to another receiver. This error can be ignored in the context of what this test is trying to achieve.

PR #177, opened 2017-07-17
DISPATCH-767 - Added code to support multi frame message streaming

ganeshmurthy:
1. Added new core API function qdr_deliver_continue() that will continue
delivering a large message
2. Modified qdr_link_process_deliveries() to not remove deliveries from
the undelivered list until they are fully delivered
3. Modified qd_message_receive() to recieve partial messages.
4. Modified qd_message_send() to be able to handle streaming send. This
function can now be called multiple times for the same message. It keeps
internal pointers to the point upto which the message has been sent and
is able to continue where it left off. Message content buffers are freed
as soon as the message has been sent to all recipients.
5. Added peer linkage for large settled deliveries and added a settled
list to handle with abrupt connection terminations when large messages
are being transmitted.
6. Added unit tests to test large messages.

### DIFF
--- a/include/qpid/dispatch/message.h
+++ b/include/qpid/dispatch/message.h
@@ -201,9 +201,9 @@ int  qd_message_get_phase_annotation(const qd_message_t *msg);
 void qd_message_set_ingress_annotation(qd_message_t *msg, qd_composed_field_t *ingress_field);
 
 /**
- * Receive message data via a delivery.  This function may be called more than once on the same
- * delivery if the message spans multiple frames.  Once a complete message has been received, this
- * function shall return the message.
+ * Receive message data frame by frame via a delivery.  This function may be called more than once on the same
+ * delivery if the message spans multiple frames. Always returns a message. The message buffers are filled up to the point with the data that was been received so far.
+ * The buffer keeps filling up on successive calls to this function.
  *
  * @param delivery An incoming delivery from a link
  * @return A pointer to the complete message or 0 if the message is not yet complete.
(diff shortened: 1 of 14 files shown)


src/router_core/transfer.c
```diff
@@ -851,6 +977,144 @@ static void qdr_delete_delivery_CT(qdr_core_t *core, qdr_action_t *action, bool
         qdr_delete_delivery_internal_CT(core, action->args.delivery.delivery);
 }
 
+static bool qdr_delivery_is_all_peers_unlinked_CT(qdr_delivery_t *dlv)
```
ganeshmurthy: @ted-ross this function (qdr_delivery_is_all_peers_unlinked_CT) needs a little bit of your extra attention. Please take a look and let me know if you have questions.

src/message_private.h
```diff
@@ -107,11 +106,21 @@ typedef struct {
...
+
+    bool                 discard;                        // Should this message be discarded?
+    bool                 receive_complete;               // true if the message has been completely received, false otherwise
+    unsigned int         fanout;                         // The number of receivers for this message. This number does not include in-process subscribers.
 } qd_message_content_t;
 
 typedef struct {
     DEQ_LINKS(qd_message_t);   // Deque linkage that overlays the qd_message_t
-    qd_message_content_t *content;
+    qd_field_location_t   cursor;          // A pointer to the current location of the outgoing byte stream.
```
ChugR: qd_iterator_pointer_t might be more appropriate for a pointer into buffers. field_location has baggage that is not needed.

src/buffer.c
```diff
@@ -83,7 +85,26 @@ size_t qd_buffer_size(qd_buffer_t *buf)
 void qd_buffer_insert(qd_buffer_t *buf, size_t len)
 {
     buf->size += len;
-    assert(buf->size <= buffer_size);
+    assert(buf->size <= BUFFER_SIZE);
+}
+
+void qd_buffer_add_fanout(qd_buffer_t *buf)
+{
+    buf->fanout++;
```
ganeshmurthy: Currently, the only function calling qd_buffer_add_fanout() is qdr_forward_new_delivery_CT(in forwarder.c) which is only called by the core thread. So, there should be no contention. But I am happy to change the fanout to atomic so that anybody else calling it outside of the core thread will not be affected.

src/message.c
```diff
@@ -846,6 +853,8 @@ qd_message_t *qd_message()
     }
 
     ZERO(msg->content);
+    msg->content->receive_complete = false;
+    msg->content->discard          = false;
```
ganeshmurthy: Agreed. Calling ZERO(msg->content) will set those flags to 0 (false) already. So, why bother setting them to false again. I will remove this.

src/message.c
```diff
@@ -999,11 +1014,84 @@ void qd_message_set_ingress_annotation(qd_message_t *in_msg, qd_composed_field_t
...
+        return 0;
+    qd_message_pvt_t *msg = (qd_message_pvt_t*) in_msg;
+    return msg->content->fanout;
+}
+
+void qd_message_add_fanout(qd_message_t *in_msg)
+{
+    assert(in_msg);
+    qd_message_pvt_t *msg = (qd_message_pvt_t*) in_msg;
+    msg->content->fanout++;
```
ganeshmurthy: yes, agreed, atomic inc

include/qpid/dispatch/message.h
```diff
@@ -296,6 +296,61 @@ qd_parsed_field_t *qd_message_get_trace      (qd_message_t *msg);
...
+bool qd_message_send_complete(qd_message_t *msg);
+
+/**
+ * Returns true if the delivery tag has already been sent.
+ */
+bool qd_message_tag_sent(qd_message_t *msg);
+
+
+/**
+ * Returns true if the delivery tag has already been sent.
```
ganeshmurthy: Fixed the wrong comment produced due to cut and paste

PR #185, opened 2017-08-09
DISPATCH-803 - The following changes were made to support forbidden d…

ganeshmurthy:
…istribution

1. Added a new attribute to the router entity called called defaultDistribution which defaults to balanced
but can be set to forbidden
2. Attaches to forbidden addresses are rejected and the link is detached
3. Anonymous senders sending to forbidden addresses will be sent back a disposition of PN_REJECTED but link will not be closed
4. Added system test system_tests_default_distribution.py to test the above cases

### DIFF
--- a/include/qpid/dispatch/router.h
+++ b/include/qpid/dispatch/router.h
@@ -43,7 +43,7 @@ typedef enum {
     QD_TREATMENT_ANYCAST_CLOSEST  = 2,
     QD_TREATMENT_ANYCAST_BALANCED = 3,
     QD_TREATMENT_LINK_BALANCED    = 4,
-    QD_TREATMENT_LINK_FORBIDDEN   = 5
+    QD_TREATMENT_FORBIDDEN        = 5
 } qd_address_treatment_t;
 
 #include <qpid/dispatch/router_core.h>
--- a/include/qpid/dispatch/router.h
+++ b/include/qpid/dispatch/router.h
@@ -42,7 +42,8 @@ typedef enum {
     QD_TREATMENT_MULTICAST_ONCE   = 1,
     QD_TREATMENT_ANYCAST_CLOSEST  = 2,
     QD_TREATMENT_ANYCAST_BALANCED = 3,
-    QD_TREATMENT_LINK_BALANCED    = 4
+    QD_TREATMENT_LINK_BALANCED    = 4,
+    QD_TREATMENT_LINK_FORBIDDEN   = 5
 } qd_address_treatment_t;
 
 #include <qpid/dispatch/router_core.h>
--- a/include/qpid/dispatch/router.h
+++ b/include/qpid/dispatch/router.h
@@ -43,7 +43,7 @@ typedef enum {
     QD_TREATMENT_ANYCAST_CLOSEST  = 2,
     QD_TREATMENT_ANYCAST_BALANCED = 3,
     QD_TREATMENT_LINK_BALANCED    = 4,
-    QD_TREATMENT_FORBIDDEN        = 5
+    QD_TREATMENT_UNAVAILABLE      = 5
 } qd_address_treatment_t;
 
 #include <qpid/dispatch/router_core.h>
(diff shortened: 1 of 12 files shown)


src/router_core/connections.c
```diff
@@ -1300,9 +1310,15 @@ static void qdr_link_inbound_first_attach_CT(qdr_core_t *core, qdr_action_t *act
...
                 // This link has a target address
                 //
-                bool           link_route;
-                qdr_address_t *addr = qdr_lookup_terminus_address_CT(core, dir, conn, target, true, true, &link_route);
-                if (!addr) {
+                bool  link_route;
+                bool  forbidden;
+                qdr_address_t *addr = qdr_lookup_terminus_address_CT(core, dir, conn, target, true, true, &link_route, &forbidden);
+                if (forbidden) {
+                    qdr_link_outbound_detach_CT(core, link, 0, QDR_CONDITION_FORBIDDEN, true);
```
ted-ross: Does this do the right thing protocol-wise?  Does the router send an attach with NULL terminus followed by a detach?

src/router_core/transfer.c
```diff
@@ -634,6 +634,23 @@ static void qdr_link_forward_CT(qdr_core_t *core, qdr_link_t *link, qdr_delivery
...
         link->total_deliveries++;
     }
+    //
+    // There is no address that we can send this delivery to, which means the addr was not found in our hastable. This
+    // can be because there were no receivers or because the address was not defined in the config file.
+    // If the treatment for such addresses is set to be forbidden, we send back a rejected disposition and detach the link
+    //
+    else if (core->qd->treatment == QD_TREATMENT_LINK_FORBIDDEN) {
+        dlv->disposition = PN_REJECTED;
+        dlv->error = qdr_error("qd:forbidden", "Sending deliveries to this address is forbidden");
```
ted-ross: You should be using one of the AMQP-specified error conditions that is in the amqp.[ch] files.  It seems that the appropriate error condition is "amqp:not-allowed", not "forbidden" (which is an error code only for the management protocol).

include/qpid/dispatch/router.h
```diff
@@ -42,7 +42,8 @@ typedef enum {
     QD_TREATMENT_MULTICAST_ONCE   = 1,
     QD_TREATMENT_ANYCAST_CLOSEST  = 2,
     QD_TREATMENT_ANYCAST_BALANCED = 3,
-    QD_TREATMENT_LINK_BALANCED    = 4
+    QD_TREATMENT_LINK_BALANCED    = 4,
+    QD_TREATMENT_LINK_FORBIDDEN   = 5
```
ganeshmurthy: Changed QD_TREATMENT_LINK_FORBIDDEN to QD_TREATMENT_FORBIDDEN

src/dispatch_private.h
```diff
@@ -52,6 +52,7 @@ struct qd_dispatch_t {
     qd_connection_manager_t *connection_manager;
     qd_policy_t             *policy;
     void                    *dl_handle;
+    qd_address_treatment_t   treatment;
```
ganeshmurthy: Changed treatment to default_treatment

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -504,6 +504,13 @@
                     "deprecated": true,
                     "description": "(DEPRECATED) This value is no longer used in the router.",
                     "create": true
+                },
+                "defaultDistribution": {
+                    "type": ["multicast", "closest", "balanced", "forbidden"],
+                    "description": "Default forwarding treatment for any address without a specified treatment. multicast - one copy of each message delivered to all subscribers; closest - messages delivered to only the closest subscriber; balanced - messages delivered to one subscriber with load balanced across subscribers; linkBalanced - for link-routing, link attaches balanced across destinations; forbidden - this address is forbidden, link attaches to an address of forbidden distribution will be rejected.",
```
grs: linkBalanced should be removed from the comment

src/router_core/transfer.c
grs: Or perhaps even better would be amqp:not-found, since we are really saying that this is not a recognised address?

python/qpid_dispatch/management/qdrouter.json
ganeshmurthy: Removed linkBalanced from comment

PR #190, opened 2017-08-23
DISPATCH-802 - Additional fixes - 1. Modified error name to amqp:prec…

ganeshmurthy:
…ondition-failed. 2. Modified error description 3. Prevented from setting the target on the return attach


gemmellr:
Tested the latest version and the attach target is now null as needed to indicate refusal, and the updated error details look good.

PR #191, opened 2017-09-01
DISPATCH-209 : initial commit for linkroute tests.

mgoulish:
These tests do not send payload messages. They only attach senders,
and then look at where those link-attaches get routed to, in a three-
router linear network. (I will add mesh once we get this code checked in.)

Two tests are disabled because of some functionality improvement opportunities
that they uncovered.

I would like a review of these test because they are a little unusual.
Specifically, the 'scripting' feature that gets different behavior out of the base
test code.
Thanks!

### DIFF
--- a/tests/system_tests_distribution.py
+++ b/tests/system_tests_distribution.py
@@ -370,296 +370,296 @@ class DistributionTests ( TestCase ):
         cls.D_addr = router_D.addresses[0]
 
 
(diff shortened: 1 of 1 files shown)


mgoulish:
I will file a new PR with a single commit -- re-written to use reactor::event_injector  technology.

PR #194, opened 2017-09-13
DISPATCH-825: Fix interlocking between message send and receive

ChugR:
Don't allow a buffer on the message buffer chain that might be removed.
Use a content-based pending buffer instead.
Don't release message lock between adding final buffer and setting
receive_complete flag.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -894,6 +894,9 @@ void qd_message_free(qd_message_t *in_msg)
             buf = DEQ_HEAD(content->buffers);
         }
 
+        if (content->pending)
+            qd_buffer_free(content->pending);
+
         sys_mutex_free(content->lock);
         free_qd_message_content_t(content);
     }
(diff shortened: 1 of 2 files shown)


src/message.c
```diff
@@ -1109,95 +1146,97 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
...
-            if (discard)
-                continue;
+        //
+        // Try to fill the remaining space in the pending buffer.
+        //
+        rc = pn_link_recv(link,
+                          (char*) qd_buffer_cursor(msg->content->pending),
+                          qd_buffer_capacity(msg->content->pending));
+
+        assert (rc != PN_EOS); // Just checked for this moments ago
```
ChugR: This assert is a mistake. Between sensing it a moment ago and now it may have become true.

PR #195, opened 2017-09-15
DISPATCH-818 - Added connection info list to the connector which stor…

ganeshmurthy:
…es the initial connection information and the backup connection information

1. If the primary connection goes down, the primary and the backup connections will be tried in quick succession
2. If the new backup connection does not provide failover-server-list, the list on the server is cleaned up to contain only the one pertinent connection info.
3. The connection info list always has at least one element in it which represents the initial successful connection information

### DIFF
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -633,7 +633,46 @@ qd_error_t qd_entity_refresh_listener(qd_entity_t* entity, void *impl)
 
 qd_error_t qd_entity_refresh_connector(qd_entity_t* entity, void *impl)
 {
-    return QD_ERROR_NONE;
+    qd_connector_t *ct = (qd_connector_t*) impl;
+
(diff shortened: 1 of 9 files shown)


src/connection_manager.c
```diff
@@ -633,7 +633,46 @@ qd_error_t qd_entity_refresh_listener(qd_entity_t* entity, void *impl)
...
+        qd_failover_item_list_t   conn_info_list = ct->conn_info_list;
+
+        qd_failover_item_t *item = DEQ_HEAD(conn_info_list);
+
+        //
+        // As you can see we are skipping the head of the list. The
+        // first item in the list is always the original connection information
+        // and we dont want to display that information as part of the failover list.
+        //
+        char failover_info[250];
```
ted-ross: Having a fixed-length string and later using strcat (not strncat) causes this to be a buffer/stack overflow vulnerability.  Please ensure that the failover_info buffer cannot be overfilled.

PR #196, opened 2017-09-15
DISPATCH-807: Message handling requires flow control to limit memory consumption

ChugR:


ChugR:
Merged through another branch

PR #204, opened 2017-10-03
DISPATCH-845 - Added connection-scoped link-route destinations.

ted-ross:
Link route destinations can be dynamically created for a connection based on a connection property in the OPEN frame.  The lifecycle of the link-route addresses tracks the connection.


ted-ross:
Note that this feature is targeted at the 1.1.0 release.

PR #207, opened 2017-10-06
Print transport handle when accepting a connection

lulf:
See https://issues.apache.org/jira/browse/DISPATCH-853


lulf:
If printing the handle on that log level is too verbose, I can change the patch to mask it on a lower log level

PR #210, opened 2017-10-16
DISPATCH-580 Adding graph=true for logStats statistics

ErnieAllen:
Added graph=true so logStats can be graphed by the console


ganeshmurthy:
Looks good @ErnieAllen feel free to check it in

PR #211, opened 2017-10-18
A pull request with Ganesh's, Chuck's, and my updates for abort handling

ted-ross:


src/message.c
```diff
@@ -1407,6 +1410,17 @@ void qd_message_send(qd_message_t *in_msg,
...
     if (msg->sent_depth < QD_DEPTH_MESSAGE_ANNOTATIONS) {
 
+        if (msg->content->aborted) {
+            // Message is aborted before any part of it has been sent.
+            // Declare the message to be sent,
+            msg->send_complete = true;
+            // the link has an outgoing deliver. abort it.
+            pn_delivery_abort(pn_link_current(pnl));
+
+            // TODO: Dispose of message buffers that may have accumulated
```
ganeshmurthy: Can we remove this TODO since the buffers are eventually freed when the message is freed?

src/message_private.h
```diff
@@ -113,6 +113,7 @@ typedef struct {
     bool                 receive_complete;               // true if the message has been completely received, false otherwise
     sys_atomic_t         fanout;                         // The number of receivers for this message. This number does not include in-process subscribers.
     bool                 q2_input_holdoff;               // hold off calling pn_link_recv
+    bool                 aborted;                        // receive completed with abort flag set
```
ChugR: For performance with smaller cache use it would make sense to group the bools together and not split them with elements that have alignment requirements.

ganeshmurthy:
Ship It!

PR #224, opened 2017-11-22
DISPATCH-884 - Added schema property protocols to allow configurable …

ganeshmurthy:
…TLS protocol versions

### DIFF
--- a/include/qpid/dispatch/server.h
+++ b/include/qpid/dispatch/server.h
@@ -306,6 +306,12 @@ typedef struct qd_server_config_t {
     char *ciphers;
 
     /**
+     * This list is a space separated string of the allowed TLS protocols. The current possibilities are TLSv1 TLSv1.1 TLSv1.2.
+     * For example, if you want to permit only TLS V.1.1 and TLSv1.2, your value for the protocols would be TLSv1.1 TLSv1.2. If this attribute is not set, then all the TLS protocols are allowed.
+     */
+    char *protocols;
+
+    /**
      * Allow the connection to be redirected by the peer (via CLOSE->Redirect).  This is
      * meaningful for outgoing (connector) connections only.
      */
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -515,7 +515,12 @@
                     "type": "string",
                     "description": "Specifies the enabled ciphers so the SSL Ciphers can be hardened. In other words, use this field to disable weak ciphers. The ciphers are specified in the format understood by the OpenSSL library. For example, ciphers can be set to ALL:!aNULL:!EXPORT56:RC4+RSA:+HIGH:+MEDIUM:+LOW:+SSLv2:+EXP; -- The full list of allowed ciphers can be viewed using the openssl ciphers command",
                     "create": true
-                },            
+                },      
+                "protocols": {
+                    "type": "string",
+                    "description": "This list is a space separated string of the allowed TLS protocols. The current possibilities are TLSv1 TLSv1.1 TLSv1.2. For example, if you want to permit only TLS V.1.1 and TLSv1.2, your value for the protocols would be TLSv1.1 TLSv1.2. If this attribute is not set, then all the TLS protocols are allowed.",
+                    "create": true
+                },                            
                 "certDb": {
                     "type": "path",
                     "description": "The absolute path to the database that contains the public certificates of trusted certificate authorities (CA).",
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -45,6 +45,7 @@ struct qd_config_ssl_profile_t {
     char        *ssl_certificate_file;
     char        *ssl_private_key_file;
     char        *ciphers;
+    char        *protocols;
 };
 
 DEQ_DECLARE(qd_config_ssl_profile_t, qd_config_ssl_profile_list_t);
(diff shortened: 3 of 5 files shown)


src/connection_manager.c
```diff
@@ -45,6 +45,7 @@ struct qd_config_ssl_profile_t {
     char        *ssl_certificate_file;
     char        *ssl_private_key_file;
     char        *ciphers;
```
ssorj: *ssl_ciphers?

src/connection_manager.c
```diff
@@ -45,6 +45,7 @@ struct qd_config_ssl_profile_t {
...
+    char        *protocols;
```
ssorj: *ssl_protocols?

include/qpid/dispatch/server.h
```diff
@@ -305,6 +305,12 @@ typedef struct qd_server_config_t {
      */
     char *ciphers;
 
+    /**
+     * This list is a space separated string of the allowed TLS protocols. The current possibilities are TLSv1 TLSv1.1 TLSv1.2.
+     * For example, if you want to permit only TLS V.1.1 and TLSv1.2, your value for the protocols would be TLSv1.1 TLSv1.2. If this attribute is not set, then all the TLS protocols are allowed.
```
ssorj: "TLS V.1.1" -> "TLSv1.1"

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -515,7 +515,12 @@
                     "type": "string",
                     "description": "Specifies the enabled ciphers so the SSL Ciphers can be hardened. In other words, use this field to disable weak ciphers. The ciphers are specified in the format understood by the OpenSSL library. For example, ciphers can be set to ALL:!aNULL:!EXPORT56:RC4+RSA:+HIGH:+MEDIUM:+LOW:+SSLv2:+EXP; -- The full list of allowed ciphers can be viewed using the openssl ciphers command",
                     "create": true
-                },            
+                },      
+                "protocols": {
+                    "type": "string",
+                    "description": "This list is a space separated string of the allowed TLS protocols. The current possibilities are TLSv1 TLSv1.1 TLSv1.2. For example, if you want to permit only TLS V.1.1 and TLSv1.2, your value for the protocols would be TLSv1.1 TLSv1.2. If this attribute is not set, then all the TLS protocols are allowed.",
```
ssorj: "then all the protocols are allowed" is not quite right.  If unset, it defers to the system-wide configuration.

include/qpid/dispatch/server.h
```diff
@@ -305,6 +305,12 @@ typedef struct qd_server_config_t {
      */
     char *ciphers;
 
+    /**
+     * This list is a space separated string of the allowed TLS protocols. The current possibilities are TLSv1 TLSv1.1 TLSv1.2.
```
ssorj: A possible enhancement is to sub spaces for commas and document it as a comma-separated list.

python/qpid_dispatch/management/qdrouter.json
bhardesty: A couple notes about the description:

I think it's most user-friendly to first define the attribute and then describe the syntax for using it. Something like this:

> The TLS protocols that this sslProfile can use. You can specify a list of one or more of TLSv1, TLSv1.1, or TLSv1.2. To specify multiple protocols, separate the protocols with a space. For example, to permit the sslProfile to use TLS v1.1 and TLS v1.2 only, you would set the value to TLSv1.1 TLSv1.2. If you do not specify a value, the sslProfile uses the TLS protocol specified by the system-wide configuration.

Instead of "TLSv1", it would be better to make the value "TLSv1.0". That would make it clear that it's referring to the 1.0 version, and not a superset of 1.x.

Also, we should define "system-wide configuration" - where would this be defined?

bhardesty: Since the value is really a list of values, would it work to define the possible values in "type"? Then we wouldn't have to describe the permitted values in the description, making it simpler. Sort of like this:

`"type": ["TLSv1", "TLSv1.1", "TLSv1.2"]`

ganeshmurthy: setting type to "type": ["TLSv1", "TLSv1.1", "TLSv1.2"] will mean that only one in the list is allowed. That is how router schema syntax works. 

For example look at "stripAnnotations". It type is set to "type": ["in", "out", "both", "no"] which means stripAnnotations can have only one of the 4 values "in" or "out" or "both" or "no"

ganeshmurthy: The proton API requires that I use TLSv1 and not TLSV1.0
Please take a look at the doc of this function
https://github.com/apache/qpid-proton/blob/master/proton-c/include/proton/ssl.h#L242

I will gladly change it to TLSv1.0 if proton does it. This is a question for @ssorj

ganeshmurthy: Your description for protocols - "The TLS protocols that this sslProfile can use. You can specify a list of one or more of TLSv1, TLSv1.1, or TLSv1.2. To specify multiple protocols, separate the protocols with a space. For example, to permit the sslProfile to use TLS v1.1 and TLS v1.2 only, you would set the value to TLSv1.1 TLSv1.2. If you do not specify a value, the sslProfile uses the TLS protocol specified by the system-wide configuration."

looks good to me. I will update.

PR #226, opened 2017-11-28
DISPATCH-872 - Added a counter for dropped-presettleds on links. qdst…

ganeshmurthy:
…at now shows this counter

### DIFF
--- a/tests/system_tests_one_router.py
+++ b/tests/system_tests_one_router.py
@@ -1552,6 +1552,14 @@ class MulticastUnsettledNoReceiverTest(MessagingHandler):
             self.conn.close()
         elif self.n_sent == self.n_released:
             self.error = None
+
+            if not self.error:
+                local_node = Node.connect(self.address, timeout=TIMEOUT)
+                for result in local_node.query(type='org.apache.qpid.dispatch.router.link').results:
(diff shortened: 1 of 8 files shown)


tests/system_tests_one_router.py
```diff
@@ -1552,6 +1552,14 @@ def check_if_done(self):
             self.conn.close()
         elif self.n_sent == self.n_released:
             self.error = None
+
+            if not self.error:
```
ted-ross: This 'if' is unnecessary.  Also, this test doesn't send pre-settled deliveries.  Why does it expect a count of dropped pre-settleds?

ganeshmurthy: Yes, the if statement is unnecessary. It should be removed.

It is true that the test does not send pre-settled deliveries. 

The router has allowUnsettledMulticast: yes
in this specific case we send an unsettled multicast to the router. But there is no receiver, so the router sends back the following to the client - 

[0x7ffb1c014c50]:0 -> @disposition(21) [role=true, first=0, last=0, settled=true, state=@released(38) []]

Hence this delivery can be considered a dropped pre-settled?

https://github.com/apache/qpid-dispatch/blob/master/src/router_core/transfer.c#L783

PR #228, opened 2017-11-29
No JIRA: Add SASL plugin requirement to SASL setup procedure

bhardesty:


doc/new-book/configuration-security.adoc
```diff
@@ -122,14 +122,20 @@ For information about additional `sslProfile` attributes, see xref:router_config
...
 
-// Do we need to say something here about only supporting Cyrus SASL?
-
 .Prerequisites
 
 Before you can set up SASL, you must have the following:
 
 * xref:generating_sasl_database[The SASL database is generated.]
 * xref:configuring_sasl_database[The SASL configuration file is configured.]
+* The Cyrus SASL plugin is installed for each SASL mechanism you plan to use.
```
ganeshmurthy: Do we want to talk here about the EXTERNAL SASL mechanism? EXTERNAL means that the communication channel is external to SASL

doc/common/fragment-router-sasl-para.adoc
```diff
@@ -0,0 +1,2 @@
+
```
ganeshmurthy: This looks good to me.

doc/new-book/configuration-security.adoc
bhardesty: In later sections, we talk about setting the SASL mechanism by using the saslMechanisms attribute (for example: https://github.com/apache/qpid-dispatch/blob/master/doc/new-book/configuration-security.adoc#adding_sasl_authentication_to_incoming_connection). Rather than list them all in the doc, we just point to the official list from the Cyrus SASL doc.

Do you think that's sufficient, or is there something particular about EXTERNAL that we should mention it as a prereq to setting up SASL?

ganeshmurthy: I think that is sufficient. This PR can be merged. Do you want me to merge it? Thanks.

bhardesty: Yes, please do.

PR #232, opened 2017-12-06
DISPATCH 894 - Made the following changes to make the system tests wo…

ganeshmurthy:
…rk on CentOS 6.

1. Modified system tests to use unittest2 instead of unittest test framework.
2. Modified tests/CMakeLists.txt to use the unit2 test runner.
3. Fixed some tests to use backwards compatible assert functions.
4. Remove unused imports in tests


ted-ross:
I see the same failures that are reported by Travis.  None of the python-based tests pass.

PR #233, opened 2017-12-08
DISPATCH-884 - Added schema property protocols to allow configurable …

ganeshmurthy:
…TLS protocol versions


bhardesty:
@ganeshmurthy The improved protocol attribute description looks good.

PR #237, opened 2017-12-20
DISPATCH-905 - Added a general get-as-string facility for proton data…

ted-ross:
… elements.  Use this facility for the port field in the failover list.


src/router_node.c
```diff
@@ -912,7 +913,7 @@ static void AMQP_opened_handler(qd_router_t *router, qd_connection_t *conn, bool
                                                                             strcmp(sym.start, QD_CONNECTION_PROPERTY_FAILOVER_PORT_KEY) == 0) {
                                             pn_data_next(props);
                                             if (pn_data_type(props) == PN_STRING) {
```
ganeshmurthy: since qdpn_data_as_string returns anything as string, the if check is not required - 
if (pn_data_type(props) == PN_STRING) {
}
should be removed

PR #240, opened 2018-01-02
Dispatch memory fixes

alanconway:
These are fixes for valgrind errors in the unit tests that I saw when running tests on master. There are more to fix, in the system tests.

Requesting review as I have been away from dispatch for a while, don't want to cause a regression. ctest with my changes is no worse than without - the policy and user_id_proxy system tests are failing with timeouts for me on master with/without these changes.


ganeshmurthy:
This looks good to me. I will squash and commit

PR #241, opened 2018-01-04
DISPATCH-906 Create procedure for integrating dispatch router with kerberos

bhardesty:
@ted-ross can you please review for technical accuracy?

### DIFF
--- a/doc/new-book/configuration-security.adoc
+++ b/doc/new-book/configuration-security.adoc
@@ -373,21 +373,25 @@ This is the service principal that {RouterName} uses.
 .Procedure
 
 . On the router's host machine, open the `/etc/sasl2/qdrouterd.conf` configuration file.
-
-. Verify that the `mech_list` attribute contains the `GSSAPI` mechanism.
 +
 --
-.The `/etc/sasl2/qdrouterd.conf` Configuration File
+.An `/etc/sasl2/qdrouterd.conf` Configuration File
 ====
 [options="nowrap"]
 ----
 pwcheck_method: auxprop
 auxprop_plugin: sasldb
 sasldb_path: qdrouterd.sasldb
+keytab: /tmp/keytabs/server.keytab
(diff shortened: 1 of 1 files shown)


bhardesty:
@enkeys I added a bit about the keytab attribute. 

I don't think we need to say anything about KRB5_CONFIG since it's specific to the Kerberos implementation (which is out of scope for Dispatch Router). AFAIK, you only need to set that environment variable if you put krb5.conf in a non-default directory (which would be part of the Kerberos configuration/implementation, not Dispatch Router configuration).

doc/new-book/configuration-security.adoc
```diff
@@ -373,22 +373,26 @@ This is the service principal that {RouterName} uses.
...
 --
-.The `/etc/sasl2/qdrouterd.conf` Configuration File
+.An `/etc/sasl2/qdrouterd.conf` Configuration File
 ====
 [options="nowrap"]
 ----
 pwcheck_method: auxprop
 auxprop_plugin: sasldb
 sasldb_path: qdrouterd.sasldb
+keytab: /tmp/keytabs/server.keytab
```
jdanekrh: Probably not `/tmp/`, though. That would be silly thing to do in actual deployment for the fact that RHELs auto-delete old files under `/tmp`. There are AFAIK supposed to be permissions related security issues around using /tmp, but I am now unable to find more on this.

enkeys: :+1: true @jdanekrh

philfest:
Nicely done @bhardesty. Looks ready to merge (from a docs pov) to me.

bhardesty:
@jdanekrh @enkeys I updated the example keytab attribute. I just set it to the default location indicated in the Cyrus SASL docs (https://www.cyrusimap.org/sasl/sasl/options.html#gssapi).

PR #244, opened 2018-01-16
DISPATCH-89: Exchange Binding forwarder (experimental)

kgiusti:
Adds a forwarder that models the pre AMQP 1.0 model of exchanges and
bindings.  Exchanges can be added via management and bound to outgoing
target addresses.  MQTT wildcard patterns are also supported.


src/router_core/exchange_bindings.c
```diff
@@ -0,0 +1,1336 @@
...
+    } else
+        query->more = false;
+    qdr_agent_enqueue_response_CT(core, query);
+}
+
+
+// create a new iterator with its own copy of the data. caller must free both the
+// iterator and the returned data buffer.  Used for retaining data extracted
+// from query messages.
+static qd_iterator_t *clone_iterator(qd_iterator_t *iter, unsigned char **data)
```
ganeshmurthy: Can this function be included as a part of the iterator API instead of being a static function ? Like qd_clone_iterator() ?

tests/system_tests_exchange_bindings.py
```diff
@@ -0,0 +1,676 @@
...
+            raise Exception("%s\n%s" % (e, out))
+        return out
+
+    def _validate_entity(self, name, kind, entities, expected):
+        for entity in entities:
+            if "name" in entity and entity["name"] == name:
+                for k,v in expected.items():
+                    self.assertTrue(k in entity)
+                    self.assertEqual(v, entity[k])
+                return;
```
ganeshmurthy: No semi colon here

src/router_core/exchange_bindings.c
kgiusti: I don't see why not.   I'll update the patch.

tests/system_tests_exchange_bindings.py
kgiusti: Heh.  One a C coder, always a C coder...

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1126,6 +1126,106 @@
...
+                "alternatePhase": {
+                    "type": "integer",
+                    "description": "The address phase for the alternate address.  Defaults to '0'.",
+                    "create": true,
+                    "required": false
+                },
+                "matchMethod": {
+                    "description": "Key matching algorithm used. '0-10' uses the legacy AMQP topic exchange wildcard match method as described in the 0-10 draft. 'mqtt' uses the MQTT topic filter wildcard match method.",
+                    "type": ["0-10", "mqtt"],
+                    "default": "0-10",
```
kgiusti: also includes 0-9,8 etc?   Need more general name.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1126,6 +1126,106 @@
...
+                    "type": "integer",
+                    "create": false
+                },
+                "deliveriesDropped": {
+                    "description": "The total number of deliveries dropped due to the lack of an outgoing subscription.",
+                    "type": "integer",
+                    "create": false
+                },
+                "deliveriesForwarded": {
+                    "description": "The total number of deliveries forwarded via matched bindings.",
```
kgiusti: include alternate in description

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1126,6 +1126,106 @@
...
+                    "description": "The name of the exchange to bind.",
+                    "type": "string",
+                    "create": true,
+                    "required": true
+                },
+                "key": {
+                    "description": "Pattern to compare against incoming message's topic.  The key is a string of zero or more tokens and wildcards. The format depends on the matchMethod configured for the exchange. For 0-10 each token is delimited by the '.' character and wild-card token '*' matches a single token and '#' matches zero or more tokens. For MQTT each token is delimited by the '/' character and wildcard token '+' matches a single token and '#' matches zero or more tokens at the end of the topic.",
+                    "type": "string",
+                    "create": true,
+                    "required": true
```
kgiusti: not required update description to describe null key

src/iterator.c
```diff
@@ -909,3 +913,40 @@ void qd_iterator_get_view_cursor(
     ptr->cursor    = iter->view_pointer.cursor;
     ptr->remaining = iter->view_pointer.remaining;
 }
+
+
+qd_iterator_t *qd_iterator_clone(const qd_iterator_t *iter, unsigned char **data)
```
kgiusti: talk with ted to remove

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1126,6 +1126,106 @@
             }
         },
 
+        "router.config.exchange": {
+            "description":"[EXPERIMENTAL] Defines a topic exchange.",
+            "extends": "configurationEntity",
+            "operations": ["CREATE", "DELETE"],
+            "attributes": {
+                "address": {
```
ssorj: @kgiusti , maybe this should be "addr".  That's the established name for addresses in the config.  (I prefer "address", but it's more important to avoid two different names the same thing.)

kgiusti: We haven't been very good at being consistent with naming: the router.node itself has an "address" attribute, while the config.autoLink's address is labeled with 'addr'.

There are a few attributes that have 'addr' as a prefix or a suffix to the full name, e.g.:
addrCount, owningAddr, and externalAddr.

So we're already inconsistent to some degree.  I'd favor using 'address' as the attribute name for an entity that has a single address attribute, and use the "<prefix>Addr" for those entities with more than one address attribute.  So my opinion is to keep the 'address' as is.

However, I'm no UI expert by any means.  Hmmm.... Let me see if I can channel the Ghost of Steve Jobs and see what his reaction is....

<Ghost of Steve Jobs>
Don't be consistent - be Brave!!!
Oh, and you're holding that wrong....
</Ghost of Steve Jobs>

jk! ;D

ssorj: I think address on router.node should change to "addr".  It's the only outlier.  The rest of the config is using "addr" consistently.

Side note: "address" for one address and "addr" for multiple addresses is an automatic user gotcha.  There is zero help for the user in those names.

And data point: The doc for "addr" on autoLink says it is just one address.

kgiusti: I don't think there really is any consistency - we've had 'router.config.address' and 'router.address' types from the beginning, not 'router.config.addr' etc.  The compound attribute names do use "Addr" as part of the name (while I don't like that either), but when there's a single address attribute it's either "addr" (router.config.autolink) or "address" (router.node).  There are only two cases so consistency hasn't been established.

I just don't like the "addr" contraction in the user-facing API (and this includes <prefix>Addr names).  We use "addr" all over the _code_ as a shortcut for "address", and I think using "addr" for autolink just seemed natural to us programmers.  But for a non-programmer does dropping the "ess" help in any way?

I feel like I'm bikeshedding this.   Are there any other folks that feel strongly either way?  I'll do whatever the consensus is.

ssorj: Ken and I chatted about how to make things more consistent.

In order to avoid two divergent conventions, I've added a request to change from "addr" to "address" to https://issues.apache.org/jira/browse/DISPATCH-918 .

We looked at how the router currently does statistics and just generally audited the names here. We came to the following changes.

Section | Current | Proposed
-- | -- | --
exchange | alternate | alternateAddress
  | deliveriesReceived | receivedCount
  | deliveriesDropped | droppedCount
  | deliveriesForwarded | forwardedCount
  | deliveriesAlternate | divertedCount
binding | exchange | exchangeName
  | nextHop | nextHopAddress
  | phase | nextHopPhase
  | key | bindingKey
  | deliveriesMatched | matchedCount

kgiusti:
Merged https://issues.apache.org/jira/browse/DISPATCH-89

PR #247, opened 2018-01-26
DISPATCH-911 - Added global delivery counts

ganeshmurthy:


ganeshmurthy:
@ErnieAllen you will have to query every router in the network and aggregate the stats. thx

src/router_core/agent_address.c
```diff
@@ -54,6 +56,8 @@ const char *qdr_address_columns[] =
      "deliveriesTransit",
      "deliveriesToContainer",
      "deliveriesFromContainer",
+     "deliveriesToRouteContainer",
+     "deliveriesFromRouteContainer",
```
ted-ross: I find the names of these two new attributes misleading.  They appear to be grouped with the former two attributes "deliveriesToContainer" and "deliveriesFromContainer" because of the similar naming.  The new stats are supposed to be grouped with "deliveriesIngress" and "deliveriesEgress".  I think better names would be "deliveriesIngressRouteContainer" and "deliveriesEgressRouteContainer".

PR #249, opened 2018-01-29
DISPATCH-921 Run npm install after stand-alone console install

ErnieAllen:
Checks to make sure npm is installed and runs npm install after console file have been copied.

### DIFF
--- a/console/CMakeLists.txt
+++ b/console/CMakeLists.txt
@@ -25,11 +25,24 @@ set(CONSOLE_BASE_SOURCE_DIR "${CMAKE_SOURCE_DIR}/console/stand-alone/")
 option(CONSOLE_INSTALL "Install stand-alone console" ON)
 if(CONSOLE_INSTALL)
 
-	# Static console files
-	install(
-	  DIRECTORY ${CONSOLE_BASE_SOURCE_DIR}
+  # Static console files
+  install(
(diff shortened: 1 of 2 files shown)


console/CMakeLists.txt
```diff
@@ -25,33 +25,46 @@ set(CONSOLE_BASE_SOURCE_DIR "${CMAKE_SOURCE_DIR}/console/stand-alone/")
...
+  # Static console files
+  install(
+    DIRECTORY ${CONSOLE_BASE_SOURCE_DIR}
           DESTINATION ${CONSOLE_STAND_ALONE_INSTALL_DIR}
     )
 
+  # run npm install for console if npm is installed
+  find_program(PROG_NPM npm)
+  if (PROG_NPM STREQUAL "PROG_NPM-NOTFOUND")
+    message(WARNING "Program npm not found. You must manually install console dependencies.")
```
kgiusti: Hey - turns out the WARNING keyword is not supported with the minimum Cmake version we require.
But I think this is the Right Way to do it, and since Proton's minimum Cmake is set to 2.8.12 we should update our cmake_minimum_version to the same.

Can you update this patch and change the cmake_minimum_version in the topmost CMakeLists.txt fileto 2.8.12?

kgiusti:
Other than the version issue this patch appears to work for me!

PR #252, opened 2018-01-31
Tross dispatch 845 1

ted-ross:
Add automatic link-routes via a connection property.  See https://issues.apache.org/jira/browse/DISPATCH-845

### DIFF
--- a/doc/book/link_routing.adoc
+++ b/doc/book/link_routing.adoc
@@ -164,3 +164,37 @@ link-routing.
 The above configuration allows Rb and Rp to reject attaches that should
 be routed to B2 with an error message that indicates that there is no
 route available to the destination.
+
+Automatic Link Routes
+~~~~~~~~~~~~~~~~~~~~~
+
+There is an alternative way to use link routes that doesn't involve any
+configuration.  An endpoint connected to a router may declare, using
+connection properties, its availability to accept routed links for a
+specific set of addresses.  This connection may be connected to the router
+through a normal listener.  There is no need to create a listener with the
+'route-container' role.
+
+The connection property that controls this feature is
+'qd.link-route-patterns'.
+
+To use the feature, a client provides a connection property when opening
+the connection to the router.  The property has the key
+'qd.link-route-patterns'.  The value is a list containing an even number of
+elements, arranged in pairs.  The first element in a pair is a string
+containing the link route pattern.  This uses the same syntax that would be
+used in a linkRoutePattern configuration item.  The second element in the
+pair is a string which has one of three values:  "in", "out", or "inout".
+The second element identifies the direction of links for which the pattern
+shall be matched.  Any number of pattern/direction pairs may be supplied in
+the connection property.
+
+The lifecycle of the resulting link routes is matched to the lifecycle of
+the connection.  When the connection closes, the link routes are removed.
+
+Automatic link routes created using this connection property are propagated
+normally across a router network.
+
+If multi-tenancy is in effect for the connection, the link route patterns
+shall be automatically placed in the tenant namespace.
+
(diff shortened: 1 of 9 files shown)


doc/book/link_routing.adoc
```diff
@@ -164,3 +164,37 @@ link-routing.
...
+
+The connection property that controls this feature is
+'qd.link-route-patterns'.
+
+To use the feature, a client provides a connection property when opening
+the connection to the router.  The property has the key
+'qd.link-route-patterns'.  The value is a list containing an even number of
+elements, arranged in pairs.  The first element in a pair is a string
+containing the link route pattern.  This uses the same syntax that would be
+used in a linkRoutePattern configuration item.  The second element in the
```
ganeshmurthy: 'linkRoutePattern' config item has been deprecated and removed from schema. Currently 'linkRoute' is used.

ted-ross: Good point.  I meant the pattern attribute for linkRoute.  This needs to be clarified.

gemmellr:
Couple of questions.

Probably a thing to consider even with the existing config method, but how are routes treated when more than one connection says it matches something?

How does this integrate with the policy stuff, e.g. is it possible to control who can set up routes this way and which addresses they can do it for?

src/router_core/route_control.c
```diff
@@ -266,6 +266,117 @@ static void qdr_auto_link_deactivate_CT(qdr_core_t *core, qdr_auto_link_t *al, q
...
+    if (pn_data_type(p) != PN_MAP)
+        return;
+
+    pn_data_enter(p);
+    pn_data_next(p);
+
+    //
+    // Work across the map looking for keys that are relevant for route modification.
+    //
+    while (pn_data_type(p) == PN_SYMBOL) {
```
ganeshmurthy: After a little bit thought, I do feel that these link routes must show up when you do a qdstat --linkRoutes. Maybe these link routes can have a boolean flag on them called "dynamic" which can be set to true for these link routes which will distinguish them from the pre-configured link routes which will have that flag set to false.

doc/book/link_routing.adoc
```diff
@@ -164,3 +164,37 @@ link-routing.
 The above configuration allows Rb and Rp to reject attaches that should
 be routed to B2 with an error message that indicates that there is no
 route available to the destination.
+
+Automatic Link Routes
+~~~~~~~~~~~~~~~~~~~~~
+
+There is an alternative way to use link routes that doesn't involve any
```
ganeshmurthy: If we decide that these link routes will not show up in the qdstat --linkRoute, should we add a comment here indicating that these link routes will not show up in the qdstat output?

ted-ross:
Re: Questions from @gemmellr 
- This probably merits additional documentation.  When there are multiple link-route destinations that match the same patterns, link attaches are balanced across the destinations.  Multiple destinations with the same address share the link load.
- It would be desirable to have policy hooks to allow/disallow this feature.  More importantly, existing produce/consume rights should be enforced based on allowed addresses.  The complexity is the wildcard patterns.  We don't currently have a way to determine if a pattern is allowed.  We can only do this with individual addresses.  One solution is to enforce access at link-attach time when the specific address is known.  Another is to implement some form of pattern validation.  The former is not good because it might cause links to be dropped by one destination when they could have been accepted by another.

tests/system_tests_dynamic_link_routes.py
```diff
@@ -0,0 +1,349 @@
...
+
+class PollTimeout(object):
+    def __init__(self, parent):
+        self.parent = parent
+
+    def on_timer_task(self, event):
+        self.parent.poll_timeout()
+
+
+class DynamicLinkRouteTest(MessagingHandler):
```
ganeshmurthy: Can we please add a dynamic link route test which makes an outgoing connection? Connection properties containing link route information could be sent by a container that is accepting this connection initiated by the router.

PR #253, opened 2018-02-12
Enhanced link-statistics

ted-ross:
This pull request contains a code change that adds the following contents to the management entities in the router:
- An index number has been added to the router.node entity to facilitate mapping histogram slots to router identities.
- A new attribute in the 'link' entity called 'ingressHistogram'.  This attribute is valid only for outgoing links on normal connections.  It contains an array of numbers that indicate the number of deliveries that arrived on each ingress router in the network.

### DIFF
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -1325,6 +1325,10 @@
                 "lastTopoChange": {
                      "description": "Timestamp showing the most recent change to this node's neighborhood.",
                      "type": "integer"
+                },
+                "index": {
+                    "description": "Index number for this router node used in statistics histograms.  This index is specific to this router.",
+                    "type": "integer"
                 }
 
             }
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -1207,6 +1207,10 @@
                     "type": "integer",
                     "graph": true,
                     "description": "The total number of modified deliveries."
+                },
+                "ingressHistogram": {
+                    "type": "list",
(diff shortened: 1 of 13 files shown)


ErnieAllen:
screenshot from 2018-02-13 15-49-18
In a 5 router network, I'm seeing an ingressHistogram that is a large array. Shouldn't the histogram have one array element per router?

ted-ross:
The histogram holds 128 values.  This is the maximum number of interior
routers.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1207,6 +1207,10 @@
                     "type": "integer",
                     "graph": true,
                     "description": "The total number of modified deliveries."
+                },
+                "ingressHistogram": {
+                    "type": "list",
+                    "description": "For outgoing links on normal connections.  This is a histogram with a value per interior router node.  The value is a count of settled deliveries on this link that ingressed the network at the specific interior router."
```
bhardesty: Suggestion for the description:

> For outgoing links on normal connections, this histogram shows the number of settled deliveries on the link that ingressed the network at each interior router node.

Also, what constitutes a "normal" connection? Does that need to be defined?

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1325,6 +1329,10 @@
                 "lastTopoChange": {
                      "description": "Timestamp showing the most recent change to this node's neighborhood.",
                      "type": "integer"
+                },
+                "index": {
+                    "description": "Index number for this router node used in statistics histograms.  This index is specific to this router.",
```
bhardesty: Minor suggestion for the description:

> The index number used in statistics histograms for this router node. This index is specific to this router.

PR #254, opened 2018-02-12
DISPATCH-911 - Added global delivery counts

ganeshmurthy:


src/router_node.c
```diff
@@ -33,7 +33,6 @@
 const char *QD_ROUTER_NODE_TYPE = "router.node";
 const char *QD_ROUTER_ADDRESS_TYPE = "router.address";
 const char *QD_ROUTER_LINK_TYPE = "router.link";
-const char *CORE_AGENT_ADDRESS = "$management";
```
ted-ross: How is this related to the PR?

ganeshmurthy: You are correct. This is not related to the PR. I saw that this was not being used anywhere, so I removed it.

PR #256, opened 2018-02-14
DISPATCH-926: Fix offered capability encoding

s1061123:
This code fix encoding of offered capability (symbol -> symbol[]), as OASIS AMQP1.0 spec.
http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-open

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -429,6 +429,8 @@ static void decorate_connection(qd_server_t *qd_server, pn_connection_t *conn, c
     //
     // Offer ANONYMOUS_RELAY capability
     //
+    pn_data_put_array(pn_connection_offered_capabilities(conn), false, PN_SYMBOL);
+    pn_data_enter(pn_connection_offered_capabilities(conn));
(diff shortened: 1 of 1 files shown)


rgodfrey:
The encoding of a single offered capability as a symbol (rather than as an array of symbols) is valid.  As per http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#section-composite-type-representation

> The multiple attribute of a field description controls whether multiple element values are permitted in the representation. A single element of the type specified in the field description is always permitted. Multiple values are represented by the use of an array where the type of the elements in the array is the type defined in the field definition. Note that a null value and a zero-length array (with a correct type for its elements) both describe an absence of a value and MUST be treated as semantically identical.

So I don't think this change is necessary

s1061123:
@rgodfrey thank you for your comment! I understood it.

rgodfrey:
@s1061123 as an author of the AMQP spec it is probably my fault that it is confusing :)

PR #261, opened 2018-02-23
DISPATCH-918 -  Initial modifications to Python schema code to introd…

ganeshmurthy:
…uce and use the deprecationName attribute. Also modified dir to direction in linkRoute and autoLink and they use the deprecationName attribute


tests/system_tests_autolinks.py
```diff
@@ -53,17 +53,18 @@ def setUpClass(cls):
...
             # Create a pair of auto-links on non-default phases for container-to-container transfers
             #
-            ('autoLink', {'addr': 'xfer.2', 'containerId': 'container.2', 'dir': 'in',  'phase': '4'}),
-            ('autoLink', {'addr': 'xfer.2', 'containerId': 'container.3', 'dir': 'out', 'phase': '4'}),
+            ('autoLink', {'addr': 'xfer.2', 'containerId': 'container.2', 'direction': 'in',  'phase': '4'}),
+            ('autoLink', {'addr': 'xfer.2', 'containerId': 'container.3', 'direction': 'out', 'phase': '4'}),
 
             #
             # Create a pair of auto-links with a different external address
+            # Leave the direction as dir to test backward compatibility.
```
ChugR: Is there a test that has both dir and direction specified to see that it fails?

ganeshmurthy: No there is not such a test. Thanks for pointing that out. I will add such a test to this PR shortly.

bhardesty:
The "dir" -> "direction" doc updates look good.

If someone attempts to use a deprecated attribute, will the operation still succeed (in addition to the deprecation log warning)? Or will the operation fail?

ganeshmurthy:
Specifying the "dir" attribute only will generate a warning that will tell you that you are using a deprecated attribute. The warning will look something like this - 
2018-02-27 11:34:41.138705 -0500 AGENT (warning) Attribute 'dir' of entity 'linkRoute' has been deprecated. Use 'direction' instead

The router will still start  

If both 'dir' and 'direction' attributes are specified, the router will display the following error and fail to start - 
2018-02-27 11:36:54.911916 -0500 MAIN (critical) Router start-up failed: Python: Exception: Cannot load configuration file test.conf: org.apache.qpid.dispatch.router.config.linkRoute: Both 'dir' and 'direction' cannot be specified for entity 'linkRoute'

PR #262, opened 2018-02-26
Crolke 918 vhost hostname

ChugR:


ChugR:
This PR adds vhost attribute 'hostname' as the name of the entity. It deprecates vhost attribute 'id'.
Policy code warns if 'id' is used and fails if neither 'id' nor 'hostname' is specified.

console/config/schema.json
```diff
@@ -1273,6 +1273,10 @@
           "type": "map", 
           "description": "A map where each key is a user group name and the value is a map of the corresponding settings for that group."
         }, 
+        "hostname": {
```
ganeshmurthy: Could this use the deprecationName attribute from my earlier PR - https://github.com/apache/qpid-dispatch/pull/261/files#diff-981a10a2c9cab3708f003b402e7f8be4R1060

console/config/schema.json
```diff
@@ -1273,6 +1273,10 @@
...
+          "type": "string", 
+          "description": "The vhost name."
+        },
```
bhardesty: In PR 255, I updated the policy attribute descriptions (including "id") to be more descriptive. For the hostname description, you could use something like this:

The hostname of the vhost. This vhost policy will be applied to any client connection that is directed to the hostname that you specify.

console/config/schema.json
ChugR: Can't be used until you check it in...

ganeshmurthy: yes, agreed

console/config/schema.json
ChugR: will do.

ChugR:
Close this PR and create a new one using @ganeshmurthy deprecationName

PR #263, opened 2018-02-28
DISPATCH-933 - Created new RouterStats entity. Moved some attributes …

ganeshmurthy:
…from Router entity to RouterStats. Router entity is not handled by the Python management agent and the RouterStats entity is handled by the C management agent


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -431,37 +431,100 @@
...
+                    "description": "The number of threads that will be created to process message traffic and other application work (timers, non-amqp file descriptors, etc.) .",
+                    "create": true
                 },
-                "nodeCount": {
-                    "type": "integer",
-                    "description":"Number of known peer router nodes.",
-                    "graph": true
+                "debugDump": {
+                    "type": "path",
+                    "description": "A file to dump debugging information that can't be logged normally.",
```
bhardesty: What does it it mean that this information can't be logged normally - just that it isn't captured by one of the logging modules?

ganeshmurthy: debug dumps are the copies of the contents of the computer's memory. When the router crashes, it will dump all its memory content into the debug dump file. which is stored in the hard disk. This file is useful for technical support people/programmers.

bhardesty: OK, so essentially a core dump. I didn't get that at first. Anyways, I'd suggest the following description (assuming of course that debugDump requires the absolute path, and that the path is to a directory, not the name of the actual debug dump file):

The absolute path to the location for debug dump files. If the router crashes, a debug dump file is created in this directory to assist in troubleshooting the problem.

ganeshmurthy: I was wrong about the debug dump file. It is not a process dump. It seems to be used only in alloc_pool.c

This is a file to which debug information is written when the logger has not yet started or has already been shutdown. When the loggers are not available, the router writes to this file.

bhardesty: In that case, how about this:

The absolute path to the location for the debug dump file. The router writes debug-level information to this file if the logger is not available.

PR #266, opened 2018-03-08
DISPATCH-940 - Allow configuration file to be used from relative path…

fgiorgetti:
DISPATCH-940 - Allow configuration file to be used from relative path when running as a daemon.


fgiorgetti:
I will create a unit test for it. Thanks for the advise on the branch in use.

fgiorgetti:
Unit test has been created.

PR #267, opened 2018-03-09
DISPATCH-154 - Added unit test

fgiorgetti:
Added a unit test to validate the issue, previously reported through DISPATCH-154.
It ensures the router is able to react to the problem and remains up and running.

### DIFF
--- a/router/src/main.c
+++ b/router/src/main.c
@@ -176,6 +176,46 @@ static void daemon_process(const char *config_path, const char *python_pkgdir,
             //
             umask(0);
 
+
+            //
+            // If config path is not represented by its full path, then
+            // save current path before changing to /
+            //
+            char *config_path_full = NULL;
+            if (strncmp("/", config_path, 1)) {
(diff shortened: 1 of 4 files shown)


router/src/main.c
```diff
@@ -176,6 +176,46 @@ static void daemon_process(const char *config_path, const char *python_pkgdir,
...
+            //
+            char *config_path_full = NULL;
+            if (strncmp("/", config_path, 1)) {
+                char *cur_path = NULL;
+                int path_size = 256;
+                int getcwd_error = 0;
+
+                cur_path = (char *) malloc(path_size * sizeof(char));
+                memset(cur_path, 0, path_size * sizeof(char));
+
```
orpiske: I'd probably replace these w/ calloc, since the general behavior should be the same as per calloc spec.

fgiorgetti:
I will recreate this pull request. This one mistakenly included changes from other PR.

PR #268, opened 2018-03-09
DISPATCH-154 - Added new test case to ensure router recovers from unresolvable hosts

fgiorgetti:
Created a new system test class to run validations on a router that has a connector with an unresolvable host, and it expects that the router wont crash and instead will log an error message accordingly.


fgiorgetti:
Submitting a new PR for this new test following recommendations.

PR #270, opened 2018-03-20
DISPATCH-933 - Created new RouterStats entity. Moved some attributes …

ganeshmurthy:
…from Router entity to RouterStats. Router entity is not handled by the Python management agent and the RouterStats entity is handled by the C management agent


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -431,37 +431,107 @@
...
+            "description": "Overall router statistics.",
+            "extends": "operationalEntity",
+            "attributes": {
+	            "id": {
+	                "description":"Router's unique identity. The router will fail to start without id.",
+	                "type": "string",
+	                "required": false,
+	                "create": true
+	            },
+	            "mode": {
```
ted-ross: Mode is now in both router and routerStats and with create access in both instances.  I don't think this belongs in routerStats.

ganeshmurthy: Good point. I will remove it

PR #274, opened 2018-03-29
DISPATCH-952 - Limit number of sessions to one on a route-container c…

ganeshmurthy:
…onnection. All links created will be created in this one session


alanconway:
Why not simply make single-session the default and (for now) only behaviour? I can't see any benefit to multiple sessions, in particular not session-per-link - AMQP is designed on the assumption of many links per session. So why leave this unhelpful behaviour lying around? If there's some future use case that requires it we can add it then, when we have some context where it is useful.

ganeshmurthy:
@ted-ross do you agree with @alanconway comments ? If yes, I will make single session the default.

ted-ross:
@ganeshmurthy I agree with @alanconway.  Let's simplify this and use single session for all out-going links for all connection roles.

PR #278, opened 2018-04-04
No JIRA yet -- example perf test in unit test environment

mgoulish:
I'd like some feedback on this. It's using some code that I am using in a non-unit-test context.
The goal here is to have a set of quick performance smoke-tests in out dispatch unit-test environment.

This code:
  * provides a high-level interface for test code to set up a router network including c-language quiver clients
  * no need for hand-written config files
  * but it needs two environment variables to show it where to get the quiver c-client and the router.
  * parses the quiver client output and writes results to a file in /tmp


mgoulish:
I don't think the Python testing framework is the right venue for large-scale performance testing.
Better technology coming soon.

PR #279, opened 2018-04-04
Dispatch 952 1

ganeshmurthy:


src/container.c
```diff
@@ -769,11 +777,12 @@ qd_link_t *qd_link(qd_node_t *node, qd_connection_t *conn, qd_direction_t dir, c
     link->node       = node;
     link->drain_mode = pn_link_get_drain(link->pn_link);
     link->remote_snd_settle_mode = pn_link_remote_snd_settle_mode(link->pn_link);
-    link->close_sess_with_link = true;
+    link->close_sess_with_link = false;
```
alanconway: Could delete this field entirely, it is never true

src/container.c
```diff
@@ -756,8 +756,16 @@ qd_link_t *qd_link(qd_node_t *node, qd_connection_t *conn, qd_direction_t dir, c
     sys_mutex_lock(node->container->lock);
     DEQ_INSERT_TAIL(node->container->links, link);
     sys_mutex_unlock(node->container->lock);
-    link->pn_sess = pn_session(qd_connection_pn(conn));
-    pn_session_set_incoming_capacity(link->pn_sess, cf->incoming_capacity);
+
+    bool open_session = false;
+
+    if (!conn->pn_sess) {
+        open_session = true;
```
alanconway: Why not just do  pn_session_open() directly here? I don't see anything that requires it to be delayed till the end.

ganeshmurthy: Good point. I removed it. That flag is *gone*

PR #288, opened 2018-04-18
Refactor cmake to use modern feature

matlo607:
I wanted to create a recipe for [Conan](https://www.conan.io/) and I experienced some issue with the build system.


ted-ross:
@matlo607 Can you please close this pull request?  It is not going to be merged.
Thanks,
-Ted

src/CMakeLists.txt
```diff
@@ -116,24 +103,42 @@ endif()
...
+endif()
+
 install(TARGETS qpid-dispatch
   LIBRARY DESTINATION ${QPID_DISPATCH_HOME})
 
-set (QPID_DISPATCH_LIB "libqpid-dispatch.so")
+if (UNIX)
+  set (QPID_DISPATCH_LIB "libqpid-dispatch.so")
+else()
+  set (QPID_DISPATCH_LIB "qpid-dispatch.dll")
```
jdanekrh: @matlo607 Did you actually managed to get useful binary of qdrouterd on Windows? I thought it won't compile on Windows.

matlo607: No, I only built it on Linux.
This work was a draft and has never been finalized.
I was expecting a more enthusiast review than the first comment I got.
That's a pity that the project wants to stick to completely outdated cmake versions only because it is installed by default on a distribution.

jdanekrh: @matlo607 What is the crucial part of the PR that was necessary for your conan package? I'd be interested in trying to get that working, even given the restriction to CMake 2.8.12. Conan seems like a really helpful tool.

jiridanek:
Partially implemented in https://issues.apache.org/jira/browse/DISPATCH-1268, for certain definition of modern.

PR #289, opened 2018-04-20
DISPATCH-972 - Use "sudo" consistently in doc

bhardesty:
In the doc, the syntax for commands that are used with root privileges is inconsistent. This PR standardizes on the "$ sudo" prefix (as opposed to "#" or "root@localhost #").


bhardesty:

PR #290, opened 2018-04-20
NO JIRA - Update links to qdrouterd.conf man page

bhardesty:
In the Dispatch Router doc, even though each link to the qdrouterd.conf man page targets the correct section of the man page, the link text does not include the name of the section to which the link points. This PR adds the name of the section to the link text.


bhardesty:

PR #293, opened 2018-04-25
DISPATCH-977: Add transaction coordinator step to link routing procedure

bhardesty:
The Dispatch Router doc mentions that you can use link routing for transactions, but it didn't describe how to configure this. 

This doc update describes how to configure a link route that supports transactions.

### DIFF
--- a/docs/books/user-guide/routing.adoc
+++ b/docs/books/user-guide/routing.adoc
@@ -575,6 +575,20 @@ connector {
 For information about additional attributes, see link:{qdrouterdConfManPageUrl}#_connector[connector] in the `qdrouterd.conf` man page.
 --
 
+. If you want clients to send local transactions to the broker, create a link route for the transaction coordinator:
++
+--
+[options="nowrap",subs="+quotes"]
+----
+linkRoute {
+    prefix: $coordinator  <1>
+    connection: __CONNECTOR_NAME__
+    direction: in
+}
+----
+<1> The `$coordinator` prefix designates this link route as a transaction coordinator. When the client opens a transacted session, the transactional state is propagated along this link route to the broker.
+--
+ 
 . If you want clients to send messages on this link route, create an incoming link route:
 +
 --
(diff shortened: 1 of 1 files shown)


bhardesty:
@grs could you please review this doc update for technical accuracy? This just adds an additional step to the procedure for creating a link route (to optionally configure the link route to support transacted messages).

doc/new-book/routing.adoc
```diff
@@ -575,6 +575,20 @@ connector {
...
+--
+[options="nowrap",subs="+quotes"]
+----
+linkRoute {
+    prefix: $coordinator  <1>
+    connection: __CONNECTOR_NAME__
+    direction: in
+}
+----
+<1> The `$coordinator` prefix designates this link route as a transaction coordinator. When the client opens a transacted session, the transactional state is propagated along this link route to the broker.
```
grs: The coordinator link is used for declaring and completing (committing or rolling back) the transactions. In the AMQP spec, the term 'transactional state' is used to tie the publication and consumption of messages to a given transaction, and is part of the transfer or disposition - loosely the message or its acknowledgement - which do not go over the coordinator link.

Perhaps it would be less (potentially) confusing to change from 'the transactional state is propagated along this link route to the broker' to 'requests to start or end a transaction are propagated along this link route to the broker.' ??

doc/new-book/routing.adoc
```diff
@@ -575,6 +575,20 @@ connector {
 For information about additional attributes, see link:{qdrouterdConfManPageUrl}#_connector[connector] in the `qdrouterd.conf` man page.
 --
 
+. If you want clients to send local transactions to the broker, create a link route for the transaction coordinator:
```
grs: One thing I think needs to be clear is that routing the coordinator really only works if there is only one broker with which any transactional interactions are undertaken.

fgiorgetti:
Looks good to me. I have created a simple scenario to validate the coordinator, as described in the doc, and it worked just fine. The final doc looks good as well.

PR #294, opened 2018-04-26
DISPATCH-960 - Resolving service port when using http listener

fgiorgetti:
As suggested by @ted-ross, I am re-submitting this PR again now using getservbyname_r which is multi-thread safe.

If you find any other possible issue, please let me know.

### DIFF
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -21,6 +21,9 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#include <netdb.h>
+#include <sys/types.h>
+#include <sys/socket.h>
 
 const char * const QD_MA_PREFIX  = "x-opt-qd.";
 const char * const QD_MA_INGRESS = "x-opt-qd.ingress";
@@ -90,11 +93,29 @@ const char * const QD_AMQP_COND_CONNECTION_FORCED = "amqp:connection:forced";
 const char * const QD_AMQP_PORT_STR = "5672";
 const char * const QD_AMQPS_PORT_STR = "5671";
 
+const char * const QD_AMQP_DFLT_PROTO = "tcp";
+
 int qd_port_int(const char* port_str) {
-    if (!strcmp(port_str, QD_AMQP_PORT_STR)) return QD_AMQP_PORT_INT;
-    if (!strcmp(port_str, QD_AMQPS_PORT_STR)) return QD_AMQPS_PORT_INT;
     errno = 0;
     unsigned long n = strtoul(port_str, NULL, 10);
     if (errno || n > 0xFFFF) return -1;
+
+    // Port is not an integer (port = 'amqp' or 'amqps')
+    if ( !n && strlen(port_str) > 0 ) {
+        // Resolve service port
+        struct servent serv_info;
+        struct servent *serv_info_res;
+        int buf_len = 4096;
+        char *buf = calloc(buf_len, sizeof(char));
+
+        // Service port is resolved
+        if ( !getservbyname_r(port_str, QD_AMQP_DFLT_PROTO, &serv_info, buf, buf_len, &serv_info_res) ) {
+            n = ntohs(serv_info.s_port);
+        } else {
+            n = -1;
+        }
+        free(buf);
+    }
+
     return n;
 }


src/amqp.c
```diff
@@ -73,11 +76,30 @@ const char * const QD_AMQP_COND_FRAME_SIZE_TOO_SMALL = "amqp:frame-size-too-smal
 const char * const QD_AMQP_PORT_STR = "5672";
 const char * const QD_AMQPS_PORT_STR = "5671";
 
+const char * const QD_AMQP_DFLT_PROTO = "tcp";
+
 int qd_port_int(const char* port_str) {
     if (!strcmp(port_str, QD_AMQP_PORT_STR)) return QD_AMQP_PORT_INT;
```
ted-ross: I understand that this is not part of the pull request, but aren't these lines (with strcmp) redundant with the strtoul below?

fgiorgetti: You're right. Maybe I should submit a new PR without the strcmp, as service name is now being resolved when strtoul returns 0. And also, using static values for amqp/amqps can also be a problem, in case someone changes the default values at /etc/services (no supposed to happen, but it is possible).

src/amqp.c
```diff
@@ -90,11 +93,29 @@ const char * const QD_AMQP_COND_CONNECTION_FORCED = "amqp:connection:forced";
...
+    // Port is not an integer (port = 'amqp' or 'amqps')
+    if ( !n && strlen(port_str) > 0 ) {
+        // Resolve service port
+        struct servent serv_info;
+        struct servent *serv_info_res;
+        int buf_len = 4096;
+        char *buf = calloc(buf_len, sizeof(char));
+
+        // Service port is resolved
+        if ( !getservbyname_r(port_str, QD_AMQP_DFLT_PROTO, &serv_info, buf, buf_len, &serv_info_res) ) {
```
jiridanek, line 112: This should also check if `serv_info_res == NULL`.

src/amqp.c
```diff
@@ -90,11 +93,29 @@ const char * const QD_AMQP_COND_CONNECTION_FORCED = "amqp:connection:forced";
...
+            n = ntohs(serv_info.s_port);
+        } else {
+            n = -1;
```
jiridanek, line 115: `n` is unsigned, so assigning `-1` to it is wrong

src/amqp.c
jiridanek, line 112: Also, `getservbyname_r` is unavailable on macOS, there `getservbyname` is already reentrant.

gemmellr:
@jiridanek Im confused by the updates here. You left comments suggesting there are things needing fixed with this, but then merged it without adjusting those bits yourself, and the JIRA is still just 'Open'.  Was the merge an accident?

jiridanek:
@gemmellr It already does what it is supposed to do, so I thought it is ready to be merged. The improvement is there, macOS travis job is allowed to fail, and the behavior on invalid protocol name is not any worse than it was before.

I did not close the Jira because I was planning to revisit the issue regarding my comments, either myself when there is the c_unittests pr merged, which should be soon now, or see if @fgiorgetti maybe fixes it in the meantime.

gemmellr:
Ok. If there are comments like 'x is wrong', 'should also do x', and 'gets a random port instead of error when doing <foo>' (I was unclear if that was also true previously or not, but it seems like part of the same issue ither way) I'd personally leave it unmerged until addressed one way or anotehr by someone or other...or perhaps resolve the original JIRA and raise another to consider as seperate issues later if not important.

Fair re: OSX failure and the job not existing (rebases can always be pushed to forks for testing first though hehe)

src/amqp.c
```diff
@@ -90,11 +93,29 @@ const char * const QD_AMQP_COND_CONNECTION_FORCED = "amqp:connection:forced";
...
     unsigned long n = strtoul(port_str, NULL, 10);
     if (errno || n > 0xFFFF) return -1;
+
+    // Port is not an integer (port = 'amqp' or 'amqps')
+    if ( !n && strlen(port_str) > 0 ) {
+        // Resolve service port
+        struct servent serv_info;
+        struct servent *serv_info_res;
+        int buf_len = 4096;
+        char *buf = calloc(buf_len, sizeof(char));
```
jiridanek, line 109: This can be allocated on stack, so `alloca` or `enum {buf_len = 4096 }; char buf[buf_len];`

jiridanek:
This PR is a resubmit of #284.

PR #296, opened 2018-04-27
Introduce a new structure under docs for multiple books

ssorj:
The structure:

```
  docs/
    notes/
    man/ <- Much of the non-book content and scripting that was previously directly under doc/
    books/
      common/ <- Common book stuff
      user-guide/ <- Was new-book/
      old-user-guide/ <- Was book/
      reference-guide/ <- To be added later on
```

 - Add scripting to generate the now primary "user-guide"
 - Rename doc/ to the more Qpid conventional docs/ and add a "make docs" target
 - Stop installing some files that IMO are likely not used much or at all since they are available online
 - Convert docs/README to markdown


bhardesty:
@ssorj this looks good. I agree with the proposed directory structure.

Just one question: is books/reference-guide/ for the "new" router config reference content that we've discussed?

I've been working on some doc contribution instructions/standards that are still based on the current doc directory structure. I can rebase off of this PR, though.

ssorj:
That is indeed what I was thinking for book/reference-guide/.  But it's not actually there in the source.  You should feel free to choose another name if it's a better fit.

BTW, I don't mind rebasing this at all, so go ahead with any commits you have lined up.

bhardesty:
@ssorj is there a convention or mitigating reason why we need the notes/ dir? As I'm putting together some guidelines for content contributors, I was planning to put them under docs/internal-resources/. If we don't need notes/, maybe just change it to internal-resources/.

Other than that, from my perspective, I think it's fine to go ahead and merge this. The commits I've lined up are based on this new directory structure, but I've been waiting to finalize them until I saw your proposal.

ssorj:
@bhardesty , I think internal-resources/ is too restrictive.  We usually notes to mean freestyle topic-oriented stuff, usually in something informal such as markdown or plain text.  The content need not be internal in any sense.  It's just not part of some bigger doc component.

That's not to say having internal-resources/ is necessarily wrong, but I don't think I understand what "internal" means here.

bhardesty:
@ssorj I think this can be rebased and merged now.

ssorj:
@bhardesty, agreed.  I'll get final confirmation from the dispatch folks and then go ahead.

ssorj:
Squashed and merged at https://git-wip-us.apache.org/repos/asf?p=qpid-dispatch.git;h=f77f92c .

PR #303, opened 2018-05-10
DISPATCH-994 - Added null checks on links/connections before enqueing…

ganeshmurthy:
… actions or calling action inseting functions

### DIFF
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -117,7 +117,8 @@ void qdr_connection_closed(qdr_connection_t *conn)
 {
     qdr_action_t *action = qdr_action(qdr_connection_closed_CT, "connection_closed");
     action->args.connection.conn = conn;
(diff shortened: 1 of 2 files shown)


src/router_core/connections.c
```diff
@@ -473,7 +474,8 @@ void qdr_link_second_attach(qdr_link_t *link, qdr_terminus_t *source, qdr_termin
     action->args.connection.link   = link;
     action->args.connection.source = source;
     action->args.connection.target = target;
-    qdr_action_enqueue(link->core, action);
+    if (link)
```
grs: I wonder if the test might be better done in AMQP_link_attach_handler in router_node.c? That is the only place that calls the method, and is the one that retrieves the null context.

ganeshmurthy: I added the null check in AMQP_link_attach_handler and in qdr_link_second_attach() just to be doubly sure.

PR #305, opened 2018-05-16
Dispatch 965 - Port to Python 3.x

kgiusti:
This patch is about as done as I can get it.  Unit tests pass on 2.6, 2.7, and 3.6.

Other than the fixups there are 2 major changes here:

1) using the __futures__ modules we're enabling python 3.x features in our python 2.x environment.  IOW your python 2.x interpreter will behave like a 3.x interpreter in some cases.  Reason: unless we're using both 2.x and 3.x development environments we'll continue to break compatibility .  Forcing us to write python 3 compatible code in a python 2.x environment should help mitigate the bit rot and keep us on track for when python 2.x is dropped in ~18 months.

Python 2.x behavior will change in the following ways:

- the print statement is disabled - only the print() function can be used
- all string literals are of the type unicode.  So type('foo') is now unicode, not str.  If you need a binary literal string use the 'b' prefix ala b'bintext'
- the division operator (/) now returns a **float** rather than an integer, even if the types involved in the division are integer. IOW 4/5 now returns 0.8, not 0  (this is a nasty one - we need to be careful here)
- module inclusion is absolute unless '.' prefix is used - we've done this for awhile so likely this won't make a difference.

2) added a python linter check to the unit tests

This will flag most other incompatible code changes to help avoid breaking stuff. This new test runs under the tox environment.  I urge you to install the python-tox package so you can take advantage of this test.


console/config/mock/section.py
```diff
@@ -17,6 +17,11 @@
...
 #
 
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+from __future__ import absolute_import
+
 import json
 import re
 from schema import Schema
```
ErnieAllen: I think this needs to be:
from .schema import Schema
since the schema.py is in the same directory as the section.py file.

kgiusti: Good catch - I'll push a fix...

ChugR:
I ran this branch OK on two systems:

* Fedora 27, Python 2.7
* Fedora 28, Python 3.6

The F27 system is my development laptop. It has python 2.6, 2.7, 3.3, 3.4, 3.4m, 3.5 3.5m, 3.6, and 3.6m. I believe that my Dispatch build runs python2.7 but to be honest I can't be sure given all the subprocesses that get launched in dispatch tests.

The F28 system was carefully groomed to have only python 3.6 and 3.6m. On that system I'm confident that the Dispatch tests ran python 3 all the way. In particular the dispatch run-time picked up embedded python3. I did not run tox on this system. Tox wants to bring along all kinds of other versions of python and I wanted as pure of a python3-only system as I could get.

Great work. Ship it.

PR #309, opened 2018-05-25
DISPATCH-1013 - Enable vhost policies to be defined on router config

fgiorgetti:


ChugR:
This is great. Ship it!

An addition to consider for the future: policy supports a _policyDir_ wherein all .json files are read for policy. This could be expanded to include .conf files as well.

PR #316, opened 2018-06-05
DISPATCH-1021 - Introduced a more flag in the action so every action …

ganeshmurthy:
…can accurately know if there is more data to come in the message


src/router_core/transfer.c
```diff
@@ -1237,7 +1246,7 @@ void qdr_drain_inbound_undelivered_CT(qdr_core_t *core, qdr_link_t *link, qdr_ad
         qdr_delivery_t *dlv = DEQ_HEAD(deliveries);
         while (dlv) {
             DEQ_REMOVE_HEAD(deliveries);
-            qdr_link_forward_CT(core, link, dlv, addr);
+            qdr_link_forward_CT(core, link, dlv, addr, 0);
```
ted-ross: You should use false for a boolean literal, not 0.

PR #318, opened 2018-06-05
DISPATCH-1023 - Added a multicast bit on the qdr_delivery_t object to…

ganeshmurthy:
… use in places where owning_addr is not available. Also added code to free multi frame non-multicast presettled messages.

### DIFF
--- a/src/router_core/transfer.c
+++ b/src/router_core/transfer.c
@@ -877,7 +877,8 @@ static void qdr_link_forward_CT(qdr_core_t *core, qdr_link_t *link, qdr_delivery
         qdr_delivery_decref_CT(core, dlv, "qdr_link_forward_CT - removed from action (1)");
         qdr_link_issue_credit_CT(core, link, 1, false);
     } else if (fanout > 0) {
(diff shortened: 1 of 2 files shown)


src/router_core/transfer.c
```diff
@@ -1162,7 +1164,7 @@ static void qdr_deliver_continue_CT(qdr_core_t *core, qdr_action_t *action, bool
         }
 
         // This is a multicast delivery
-        if (qdr_is_addr_treatment_multicast(in_dlv->link->owning_addr)) {
+        if (qdr_is_addr_treatment_multicast(in_dlv->link->owning_addr) || in_dlv->multicast || in_dlv->settled) {
```
ted-ross: Why did you keep the call to qdr_is_addr_treatment_multicast here?  Isn't it always true that the in_dlv->multicast flag will be initialized appropriately in qdr_link_forward_CT?

Also, why did you add the in_dlv->settled condition?  That is new/changed functionality.

ted-ross: If the latter is handling for multi-frame unicasts, then the comment above is no longer accurate.

PR #323, opened 2018-06-12
DISPATCH-976 - Commenting last assert which is causing router to crash

fgiorgetti:


fgiorgetti:
I believe I found the reason for the failure. I will provide a new PR and ask Chuck to review it.

PR #324, opened 2018-06-13
DISPATCH-976 - Fixed issue with policy validation of allowed addresses

fgiorgetti:


fgiorgetti:
@ganeshmurthy and @ChugR after debugging it a little bit more, I noticed that the first if statement that checks for an absent user key in the iterating allowed address, was also checking the "need_check_nosubst" boolean, enforcing that only one address with "absent" user key is defined.
So with this boolean flag, in case more than one address is defined in the tree (without the user key), it  causes the last "else" statement to be processed and then executing the "assert(false);" which was crashing the router.
In this PR I have removed the need_check_nosubst flag which seems to resolve the issue reported in DISPATCH-1026 as well.

ChugR:
Looking at it again I think all three of these need_check_* booleans are suspect. Don't delete just the one, get rid of them all.

fgiorgetti:
Done. Seems to be working fine.

PR #326, opened 2018-06-21
Logging

alanconway:
Some minor changes to dispatch logging based on experience debugging the HTTP Envoy bridge.
See the commit comments for explanation.

One additional question: I was confused by the new messageLoggingComponents config, when I set MESSAGE enable trace+ I saw nothing at all, which was surprising. Perhaps the default should be "all" so that enabling MESSAGE trace does what folks expect and they can turn the logging down per connector/listener if they wish. Since trace logging will be off in production (I hope!) this won't have any effect except on users that are deliberately debugging AND have turned on trace logging for at least the MESSAGE module, so are expecting a lot of log data.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -543,8 +543,8 @@ static void on_accept(pn_event_t *e)
     }
     ctx->listener = listener;
     qd_log(listener->server->log_source, QD_LOG_TRACE,
-           "[%"PRIu64"] Accepting incoming connection from %s to %s",
-           ctx->connection_id, qd_connection_name(ctx), ctx->listener->config.host_port);
+           "[%"PRIu64"] Accepting incoming connection to %s",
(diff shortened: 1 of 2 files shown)


src/server.c
```diff
@@ -651,8 +651,9 @@ static void on_connection_bound(qd_server_t *server, pn_event_t *e) {
         pn_sasl_set_allow_insecure_mechs(sasl, config->allowInsecureAuthentication);
         sys_mutex_unlock(ctx->server->lock);
 
-        qd_log(ctx->server->log_source, QD_LOG_INFO, "Accepted connection to %s from %s",
-               name, ctx->rhost_port);
+        qd_log(ctx->server->log_source, QD_LOG_INFO,
```
ted-ross: There appear to be more arguments than format directives in this call.
Why did you remove the information about the remote host?

PR #327, opened 2018-06-22
DISPATCH-1041 - Add tests to validate each global delivery count accu…

ganeshmurthy:
…rately


fgiorgetti:
@ganeshmurthy it looks great to me. One thing I noticed is that the only new metric that is not present in this test is: "acceptedDeliveries". But I see it being used in system_tests_one_router.

ganeshmurthy:
Good catch. Not sure how I missed that. I will add anew test for acceptedDeliveries and will let you know

fgiorgetti:
It covers everything now!

PR #328, opened 2018-06-22
DISPATCH-1045 - Release the delivery only after the entire message ha…

ganeshmurthy:
…s been received

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -316,7 +316,8 @@ static void AMQP_rx_handler(void* context, qd_link_t *link)
         // The entire message has been received but this message needs to be discarded
         //
         if (qd_message_is_discard(msg)) {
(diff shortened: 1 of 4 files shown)


src/router_node.c
```diff
@@ -312,12 +307,30 @@ static void AMQP_rx_handler(void* context, qd_link_t *link)
...
+        //
+        // The entire message has been received and we are ready to consume the delivery by calling pn_link_advance().
+        //
+        pn_link_advance(pn_link);
+
+        //
+        // The entire message has been received but this message needs to be discarded
+        //
+        if (qd_message_is_discard(msg)) {
+            pn_delivery_update(pnd, qdr_delivery_disposition(delivery));
```
ted-ross: Is the delivery disposition _always_ set when discard is set?  What if the disposition was not set?  I think the pn_delivery_update should not be called in this case.

PR #334, opened 2018-06-27
Dispatch 1046 - fix policy handling in python3 environment

kgiusti:
Had to change the way we pass args to the host address parse tree api under python3

c_char_p is a bytes type under python3.  Passing and returning a unicode string causes it to fail.  Had to manually convert between the two.  Also need to save a copy of the pattern in the tree as a payload.


ChugR:
This patch fixes the c_char_p in py3 and improves interfaces with some const additions.

The patch also repairs some incorrect handling of policy's parse tree. It looks like there were resource leaks when the trees were deleted. I don't think that was just a py3 issue.

This PR looks OK to me.

PR #338, opened 2018-07-09
DISPATCH-1066 - Document capability to configure TLS protocol version

bhardesty:
In the Dispatch Router user guide, I added some information about the "protocols" attribute for users who want to restrict the permitted versions of TLS to use with the router.

@ganeshmurthy, can you please review for technical accuracy?


bhardesty:
@ganeshmurthy, this is now ready to merge.

PR #345, opened 2018-07-23
DISPATCH-1085 - Modified AMQP_link_detach_handler to flush out the re…

ganeshmurthy:
…maining bytes in the message buffers before responding to detaches

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -1128,13 +1128,15 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
         return discard_receive(delivery, link, (qd_message_t *)msg);
     }
 
+    // if q2 holdoff has been disabled (disable_q2_holdoff=true), we keep receiving.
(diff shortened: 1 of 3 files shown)


src/message.c
```diff
@@ -1128,13 +1128,15 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
...
+    //      have been processed and freed by outbound processing then
+    //      message holdoff is cleared and receiving may continue.
     //
-    // If input is in holdoff then just exit. When enough buffers
-    // have been processed and freed by outbound processing then
-    // message holdoff is cleared and receiving may continue.
-    //
-    if (msg->content->q2_input_holdoff) {
-        return (qd_message_t*)msg;
+    if (!msg->content->disable_q2_holdoff) {
```
ted-ross: Is this change necessary?  Was there a bug in qd_message_holdoff_Q2_disable?

PR #357, opened 2018-08-10
DISPATCH-1099 - Added plumbing that sends ticks into the core

ganeshmurthy:


ganeshmurthy:
It adds one action record into the action list every one second. That should not bring down the performance but I will try using quiver and see what I get

include/qpid/dispatch/router_core.h
```diff
@@ -797,4 +799,9 @@ qdr_connection_info_t *qdr_connection_info(bool             is_encrypted,
                                            int              ssl_ssf,
                                            bool             ssl);
 
+typedef struct qdr_timer_work_t qdr_timer_work_t;
+typedef void (*qdr_timer_cb_t)(qdr_core_t *core, void* context);
+qdr_timer_work_t *qdr_timer_schedule(qdr_core_t *core, qdr_timer_cb_t callback, void *timer_context, int timer_delay);
+void qdr_timer_delete(qdr_core_t *core, qdr_timer_work_t *timer_work);
```
ted-ross: Some comments on these function prototypes would be helpful?  What exactly does qdr_timer_delete do?  Should there be a qdr_timer_cancel()?

src/router_core/core_timer.c
```diff
@@ -0,0 +1,46 @@
...
+
+qdr_timer_work_t *qdr_timer_schedule(qdr_core_t *core, qdr_timer_cb_t callback, void *timer_context, int timer_delay)
+{
+    qdr_timer_work_t *timer_work = new_qdr_timer_work_t();
+    ZERO(timer_work);
+    timer_work->handler = callback;
+    timer_work->timer_delay = timer_delay;
+    timer_work->on_timer_context = timer_context;
+
+    DEQ_INSERT_TAIL(core->timer_list, timer_work);
```
ted-ross: The timer algorithm should be reviewed, I think.  This at-tail insertion with absolute delay is going to be very inefficient if there are a large number of timers (which is a probably scenario for cases where there are many auto-links).

src/router_core/transfer.c
```diff
@@ -717,6 +725,35 @@ void qdr_delivery_decref_CT(qdr_core_t *core, qdr_delivery_t *dlv, const char *l
         qdr_delete_delivery_internal_CT(core, dlv);
 }
 
+static void qdr_process_tick_CT(qdr_core_t *core, qdr_action_t *action, bool discard)
+{
+    if (discard)
+        return;
+
+    if (DEQ_SIZE(core->timer_list) == 0)
+        return;
```
ted-ross: This check is not needed.  If the timer list is empty, the while loop below will exit immediately.

src/router_core/transfer.c
```diff
@@ -717,6 +725,35 @@ void qdr_delivery_decref_CT(qdr_core_t *core, qdr_delivery_t *dlv, const char *l
...
+            next_timer_work = DEQ_NEXT(timer_work);
+            DEQ_REMOVE(core->timer_list, timer_work);
+            free_qdr_timer_work_t(timer_work);
+            timer_work = next_timer_work;
+        }
+        else {
+            timer_work->timer_delay -=1;
+            timer_work = DEQ_NEXT(timer_work);
+        }
+
```
ted-ross: As stated in another comment: This algorithm is going to be costly when there are a large number of timers in the list.  The tick handler must visit every timer on every tick.

If the list were ordered by duration and the time stored in the timer_work were relative (i.e. number of ticks after the previous work), then the tick handler only needs to visit the head of the list.

src/router_node.c
```diff
@@ -1179,6 +1180,9 @@ static void qd_router_timer_handler(void *context)
     // Periodic processing.
     //
     qd_pyrouter_tick(router);
+
+    qdr_process_tick(router->router_core);
+
```
ted-ross: Nitpick:  Why so much whitespace between these lines?

include/qpid/dispatch/router_core.h
ted-ross: Other timer APIs we've created look more like this:

qdr_core_timer_t *qdr_core_timer();
void qdr_core_timer_schedule(qdr_core_timer_t *timer, ...);
void qdr_core_timer_cancel(qdr_core_timer_t *timer);
void qdr_core_timer_free(qdr_core_timer_t *timer);

This prevents you from having to create/allocate a new timer each time you reschedule.  The timers we will be using will be recurring, so this more classic API fits well.

PR #359, opened 2018-08-10
DISPATCH-1100: add auto delete linkRoutes

kgiusti:


src/router_core/route_control.c
```diff
@@ -189,37 +189,40 @@ static void qdr_link_route_activate_CT(qdr_core_t *core, qdr_link_route_t *lr, q
...
     //
-    if (lr->addr) {
-        qdr_del_connection_ref(&lr->addr->conns, conn);
-        if (DEQ_IS_EMPTY(lr->addr->conns)) {
+    if (lr->addr && qdr_del_connection_ref(&lr->addr->conns, conn)) {
+        assert(lr->active_ct > 0);
+        lr->active_ct -= 1;
+        if (lr->active_ct == 0) {
             char *address = qdr_link_route_pattern_to_address(lr->pattern, lr->dir);
             qd_log(core->log, QD_LOG_TRACE, "Deactivating link route pattern [%s]", address);
```
ganeshmurthy: Should this log message say "Deleting" instead of "Deactivating" ?

kgiusti: not at that point, no.  If the lr is not configured delete-on-close then the lr is simply deactivated (not deleted) at that point.

if the lr is delete-on-close it will deactivate first, then the call to qdr_route_del_link_route_CT will log a trace message that the link route has been removed.

PR #363, opened 2018-08-14
Dispatch 1103

ganeshmurthy:


src/router_core/connections.c
```diff
@@ -34,6 +34,9 @@ static void qdr_link_delete_CT(qdr_core_t *core, qdr_action_t *action, bool disc
 ALLOC_DEFINE(qdr_connection_t);
 ALLOC_DEFINE(qdr_connection_work_t);
 
+const int AUTO_LINK_FIRST_RETRY_INTERVAL = 2;
```
ted-ross: All of this auto-link-retry logic should be in the route_control module, not the connections module.  One hint that this is the case is the fact that you moved qdr_route_log_CT from static to non-static.  qdr_route_log_CT is a static helper function for route_control logic.

tests/system_tests_autolinks.py
```diff
@@ -35,6 +36,124 @@
 CONNECTION_PROPERTIES = {u'connection': u'properties', u'int_property': 6451}
```
ted-ross: Please also provide a test which receives attaches for pre-configured auto-links.  Have the test reject the auto-links and see that they retry.  Also, have the test accept the auto-link and then detach it to see the retry.

PR #366, opened 2018-08-24
DISPATCH-1103 - Added code to retry failed auto links using the core …

ganeshmurthy:
…timeer API

### DIFF
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -1329,6 +1329,9 @@ static void qdr_connection_closed_CT(qdr_core_t *core, qdr_action_t *action, boo
     while (link_ref) {
         qdr_link_t *link = link_ref->link;
 
+        if (link->auto_link && link->auto_link->retry_timer)
+            qdr_core_timer_cancel_CT(core, link->auto_link->retry_timer);
+
         //
         // Clean up the link and all its associated state.
         //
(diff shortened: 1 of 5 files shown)


src/router_core/route_control.h
```diff
@@ -43,6 +43,13 @@ qdr_auto_link_t *qdr_route_add_auto_link_CT(qdr_core_t          *core,
                                             qd_parsed_field_t      *connection_field,
                                             qd_parsed_field_t   *external_addr);
 
+/**
+ * Attempts re-establishing auto links across the related connections/containers
+ * @param core Pointer to the core object returned by qd_core()
+ * @param context - the auto link that needs to be activated
+ */
+void qdr_route_attempt_auto_link_CT(qdr_core_t      *core, void   *context);
```
ted-ross: This is not part of the route_control API.  It is simply an internal helper within the module.  It should be removed from the header file (API) and made static in the source file.

src/router_core/connections.c
```diff
@@ -1329,6 +1329,9 @@ static void qdr_connection_closed_CT(qdr_core_t *core, qdr_action_t *action, boo
     while (link_ref) {
         qdr_link_t *link = link_ref->link;
 
+        if (link->auto_link && link->auto_link->retry_timer)
+            qdr_core_timer_cancel_CT(core, link->auto_link->retry_timer);
```
ted-ross: The fact that there is a retry-timer in the auto-link is an implementation detail that should not be visible at this level.  It would be better to have a call in the route_control module for "auto_link_connection_closed" or similar.

src/router_core/connections.c
```diff
@@ -1753,6 +1756,12 @@ static void qdr_link_inbound_detach_CT(qdr_core_t *core, qdr_action_t *action, b
         link->auto_link->state = QDR_AUTO_LINK_STATE_FAILED;
         free(link->auto_link->last_error);
         link->auto_link->last_error = qdr_error_description(error);
+
+        //
+        // The auto link has failed. Periodically retry setting up the auto link until
+        // it succeeds.
+        //
+        qdr_route_retry_auto_link_CT(core, link);
```
ted-ross: Same comment here:  At this location in the architecture, the fact that there is a retry is abstracted.  This call should be "auto_link_detached" or similar.  Let the route_control module be responsible for what actions it takes when an auto-link is detached by the peer.

PR #372, opened 2018-09-06
DISPATCH-1116: prevent module being updated on log instances

grs:


grs:
@ganeshmurthy @ted-ross does this look ok?

PR #373, opened 2018-09-06
DISPATCH-1117: suggested doc fixes

grs:


grs:
@bhardesty does this look ok?

PR #382, opened 2018-09-24
DISPATCH-1129 - Moved call to forwarding multicast messages to end of…

ganeshmurthy:
… forwarder function. Also modified qd_message_check_LH to not validate the message if bufferes have already been freed


src/router_core/router_core_private.h
```diff
@@ -440,6 +440,17 @@ struct qdr_link_ref_t {
 ALLOC_DECLARE(qdr_link_ref_t);
 DEQ_DECLARE(qdr_link_ref_t, qdr_link_ref_list_t);
 
+
+typedef struct qdr_forward_deliver_info_t {
```
ted-ross: This is only used in forwarder.c.  Why not move all of this into that file?

PR #383, opened 2018-09-25
DISPATCH-1129 - Moved call to forwarding multicast messages to end of…

ganeshmurthy:
… forwarder function. Also modified qd_message_check_LH to not validate the message if bufferes have already been freed


ChugR:
+1 Ship It
Looks good and works. Failing tests could never get beyond a few hundred messages. With this patch I'm past 100 million messages.

PR #384, opened 2018-09-26
DISPATCH-1096 - priority sheaf cleanup bug

mgoulish:
This is for the bug that Ganesh and Chuck saw while killing and restarting routers.
The test succeeded 10 times each (with 0 false results) in these 4 cases:
   { Debug, RelWithDebInfo } x { Should_Fail, Should_Succeed }

### DIFF
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -25,6 +25,7 @@
 #include "router_core_private.h"
 #include "core_link_endpoint.h"
 
+
 static void qdr_connection_opened_CT(qdr_core_t *core, qdr_action_t *action, bool discard);
 static void qdr_connection_closed_CT(qdr_core_t *core, qdr_action_t *action, bool discard);
 static void qdr_link_inbound_first_attach_CT(qdr_core_t *core, qdr_action_t *action, bool discard);
@@ -1348,6 +1349,7 @@ static void qdr_connection_closed_CT(qdr_core_t *core, qdr_action_t *action, boo
 
     qdr_connection_t *conn = action->args.connection.conn;
 
+
     //
     // Deactivate routes associated with this connection
     //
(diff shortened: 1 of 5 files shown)


ChugR:
This patch clears up the too-many-links for sheaf error and improves my router network stability. FOr this I approve the patch.

But in tests before and after this patch my throughput went from 2600 msg/S to 200-300 msg/S. The performance drop needs to be addressed as a separate issue.

src/router_core/connections.c
```diff
@@ -1466,8 +1470,8 @@ static void qdr_attach_link_data_CT(qdr_core_t *core, qdr_connection_t *conn, qd
         // are assigned priorities in the order in which they are attached.
         int next_slot = core->data_links_by_mask_bit[conn->mask_bit].count ++;
         if (next_slot >= QDR_N_PRIORITIES) {
-            qd_log(core->log, QD_LOG_ERROR, "Attempt to attach too many inter-router links for priority sheaf.");
-            return;
+            qd_log(core->log, QD_LOG_CRITICAL, "Attempt to attach too many inter-router links for priority sheaf.");
+            exit(1);
```
ted-ross: Exiting the process here is a bit heavy handed.  I think it's sufficient to log the critical error.

ted-ross:
> This patch clears up the too-many-links for sheaf error and improves my router network stability. FOr this I approve the patch.
> 
> But in tests before and after this patch my throughput went from 2600 msg/S to 200-300 msg/S. The performance drop needs to be addressed as a separate issue.

I don't see any changes in this PR that should affect throughput performance.

ted-ross:
> > This patch clears up the too-many-links for sheaf error and improves my router network stability. FOr this I approve the patch.
> > But in tests before and after this patch my throughput went from 2600 msg/S to 200-300 msg/S. The performance drop needs to be addressed as a separate issue.
> 
> I don't see any changes in this PR that should affect throughput performance.

I withdraw the above statement.  There appears to be a memory leak that is causing performance (and other) problems.

mgoulish:
I yanked that exit(1).
I thought I needed that to detect failure in my python test, but that is not true.
Tests all still did the right thing, in 10 trials each (40 total) in the set:    {should_fail, should_succeed} *  { DEBUG, RelWithDebInfo}

PR #385, opened 2018-09-27
DISPATCH-1131: add better validation of address patterns and prefixes

kgiusti:
This moves the address pattern/prefix validation to a common function
that can be share by both link routes add configured addresses.


kgiusti:
Oops left in an extraneous check..

PR #389, opened 2018-10-10
DISPATCH-1130 Expose link priority

ErnieAllen:
Adds priority to router.link management entity in the schema.
Connects link.priority to schema attribute.
Adds priority column to qdstat -l


ganeshmurthy:
@ErnieAllen  This change looks good. As part of this PR, can you also please add a test to system_tests_qdstat to verify that the correct link priority is showing up? Thanks.

PR #393, opened 2018-10-19
DISPATCH-1143: New feature: connection-scoped link routes

kgiusti:
Haven't changed the name yet - still attachSubscription.   Will do that with a separate patch once the link route replication stuff is done.


python/qpid_dispatch_internal/management/agent.py
```diff
@@ -417,6 +417,13 @@ def create(self):
     def __str__(self):
         return super(LinkRouteEntity, self).__str__().replace("Entity(", "LinkRouteEntity(")
 
+class RouterConnectionAttachsubscriptionEntity(EntityAdapter):
```
ted-ross: Why is this even mentioned in the Python management code?  It can't be configured and management commands are handled in the router core.

python/qpid_dispatch_internal/management/config.py
```diff
@@ -67,6 +67,7 @@ def transform_sections(sections):
             if s[0] == "autoLink":  s[0] = "router.config.autoLink"
             if s[0] == "exchange":  s[0] = "router.config.exchange"
             if s[0] == "binding":   s[0] = "router.config.binding"
+            if s[0] == "attachSubscription":   s[0] = "router.connection.attachSubscription"
```
ted-ross: Why go to all the trouble of adding support for configuration of a non-configurable entity?

python/qpid_dispatch_internal/management/agent.py
kgiusti: Error reporting.  Without it the router fails with a generic parse error. With this we get:

qdrouterd: Python: CError: Configuration: attachSubscription cannot be created via the configuration file

I'm going to have to re-work this on changing the name in any case.  I can pull it if you want.

python/qpid_dispatch_internal/management/config.py
kgiusti: See previous.

include/qpid/dispatch/router_core.h
```diff
@@ -721,10 +723,11 @@ typedef struct qdr_query_t qdr_query_t;
  * @param name The name supplied for the entity
  * @param in_body The body of the request message
  * @param out_body A composed field for the body of the response message
+ * @param in_conn The identity of the connection over which the mgmt message arrived (0 if config file)
  */
 void qdr_manage_create(qdr_core_t *core, void *context, qd_router_entity_type_t type,
                        qd_iterator_t *name, qd_parsed_field_t *in_body, qd_composed_field_t *out_body,
-                       qd_buffer_list_t body_buffers);
+                       qd_buffer_list_t body_buffers, uint64_t in_conn);
```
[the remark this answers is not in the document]
kgiusti: I think that may introduce a timing window.  Is it possible that the connection can be removed before the core gets the processing the create request?

python/qpid_dispatch_internal/management/agent.py
kgiusti: Ya know, the default error message isn't too bad:

2018-10-19 15:37:07.199514 -0400 ERROR (error) Python: Exception: Cannot load configuration file X.conf: No such entity type 'org.apache.qpid.dispatch.attachSubscription'
2018-10-19 15:37:07.199789 -0400 ERROR (error) Traceback (most recent call last):
  File "/home/kgiusti/work/dispatch/qpid-dispatch/python/qpid_dispatch_internal/management/config.py", line 169, in configure_dispatch
    config = Config(filename)
  File "/home/kgiusti/work/dispatch/qpid-dispatch/python/qpid_dispatch_internal/management/config.py", line 56, in __init__
    % (filename, e))
Exception: Cannot load configuration file X.conf: No such entity type 'org.apache.qpid.dispatch.attachSubscription'

2018-10-19 15:37:07.199805 -0400 MAIN (critical) Router start-up failed: Python: Exception: Cannot load configuration file X.conf: No such entity type 'org.apache.qpid.dispatch.attachSubscription'
qdrouterd: Python: Exception: Cannot load configuration file X.conf: No such entity type 'org.apache.qpid.dispatch.attachSubscription'

A little better than a generic parse error.

ted-ross: I think we can dispense with all this work just to improve the error message.  We should make sure that the man page for the configuration file doesn't mention this type.

include/qpid/dispatch/router_core.h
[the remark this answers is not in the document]
ted-ross: Yes, I think that is possible.  I'd like to give this some more thought.

PR #394, opened 2018-10-23
DISPATCH-1150: Router core messaging client

kgiusti:


src/router_core/core_client_api.h
```diff
@@ -0,0 +1,165 @@
...
+ * @param body - the message body for the sent message. Ownership is transferred
+ *        to this call - the caller must not reference the composed field on return.
+
+ * @param on_reply_cb - (optional) invoked when reply message arrives
+ * @param on_ack_cb - (optional) invoked when sent message disposition is set
+ * @param done_cb - (optional) called once request is done (for cleanup)
+ * @return zero on success.
+ */
+int qdrc_client_request_CT(qdrc_client_t *client,
+                           uintptr_t  request_context,
```
ted-ross: Why is the context not a void* ?

PR #415, opened 2018-11-08
DISPATCH-1176 Prevent uninitialized delivery from crashing router

ErnieAllen:
This prevents the crash but I'm concerned about:
- what happens to this delivery? does it get dropped?
- what about router inter-operability between different versions of the router?

### DIFF
--- a/src/router_core/router_core_private.h
+++ b/src/router_core/router_core_private.h
@@ -321,8 +321,7 @@ ALLOC_DECLARE(qdr_router_ref_t);
 DEQ_DECLARE(qdr_router_ref_t, qdr_router_ref_list_t);
 
 typedef enum {
-    QDR_DELIVERY_UNINITIALIZED = 0,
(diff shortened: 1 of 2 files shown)


src/router_core/router_core_private.h
```diff
@@ -321,7 +321,8 @@ ALLOC_DECLARE(qdr_router_ref_t);
 DEQ_DECLARE(qdr_router_ref_t, qdr_router_ref_list_t);
 
 typedef enum {
-    QDR_DELIVERY_NOWHERE = 0,
+    QDR_DELIVERY_UNINITIALIZED = 0,
+    QDR_DELIVERY_NOWHERE,
```
ted-ross: I claim that QDR_DELIVERY_NOWHERE is the proper default value for this type.  This type is used to indicate in which list a particular delivery is contained.  If it has never been inserted into a list, it should indicate NOWHERE.

PR #416, opened 2018-11-08
DISPATCH-1176 - Do not continue processing settled streaming deliveri…

ganeshmurthy:
…es when they have nowhere to go. If the settled delivery is not in any of the lists, dont process it

### DIFF
--- a/src/router_core/transfer.c
+++ b/src/router_core/transfer.c
@@ -1252,6 +1252,14 @@ static void qdr_deliver_continue_CT(qdr_core_t *core, qdr_action_t *action, bool
 
         // This is a multicast delivery or if this is a presettled multi-frame unicast delivery.
         if (in_dlv->multicast || in_dlv->settled) {
+
+            //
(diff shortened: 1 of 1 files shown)


src/router_core/transfer.c
```diff
@@ -1252,6 +1252,14 @@ static void qdr_deliver_continue_CT(qdr_core_t *core, qdr_action_t *action, bool
 
         // This is a multicast delivery or if this is a presettled multi-frame unicast delivery.
         if (in_dlv->multicast || in_dlv->settled) {
+
+            //
+            // If a delivery is settled but did not go into one of the lists, that means that it is going nowhere.
+            // We dont want to deal with such deliveries.
+            //
+            if (in_dlv->settled && in_dlv->where == QDR_DELIVERY_NOWHERE)
```
ted-ross: Can a test be written that exercises this case?
Did you verify that the delivery being processed has not been freed?

PR #420, opened 2018-11-13
DISPATCH-1177 - Added new option -e to qdstat. This will show only th…

ganeshmurthy:
…e edge connections and show which of those connections is active

### DIFF
--- a/src/router_core/agent_connection.c
+++ b/src/router_core/agent_connection.c
@@ -200,11 +200,15 @@ static void qdr_connection_insert_column_CT(qdr_core_t *core, qdr_connection_t *
 
     case QDR_CONNECTION_ACTIVE:
         if (conn->role == QDR_ROLE_EDGE_CONNECTION) {
-            if (core->active_edge_connection == conn)
+            if (core->router_mode == QD_ROUTER_MODE_INTERIOR) {
                 qd_compose_insert_bool(body, true);
-            else
-                qd_compose_insert_bool(body, false);
-
+            }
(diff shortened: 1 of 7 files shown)


fgiorgetti:
@ganeshmurthy I have executed qdstat with -e on a formed network here and I see the "edge" routers from an interior router, and the "interior" router when executed against edge. It makes sense.
What about the message when we run "qdstat -n" on an edge router, which shows:
"Router is Standalone - No Router List".
Is it changing?
Thanks.

ganeshmurthy:
Should it say  "Router is Standalone or Edge - No Router List".

ChugR:
If I have a network of interior routers, each with some edge routers, I can qdstat -e each OK. But if I add -r some-other-router and the other router is on another host them qdstat fails with **KeyError: 'active'.**

    > qdstat -e -b amqp://$INTB_normal -r INTB
    Connections
     id  host                container  role  dir  security     authentication  tenant  active
        ===========================================================================================
      2   127.0.0.1:58780     EB2        edge  in   no-security  anonymous-user          no
      72  192.168.1.16:33324  EB1        edge  in   no-security  anonymous-user          no
    > qdstat -e -b amqp://$INTB_normal -r INTC
    KeyError: 'active'

ChugR:
The problem was router INTC was running master branch code and did not have this PR's changes. With both routers running DISPATCH-1177-1@0e4cd the command works ok. 

This PR could be improved if there was a way to give a meaningful error message instead of KeyError.

PR #428, opened 2018-12-19
DISPATCH-1115: make TravisCI job use a Proton release in addition to master

gemmellr:
There have been various cases that using only Proton master for CI caused Dispatch test failures, or lead to Dispatch master only working with proton master when this wasn't particularly desired at the time. This can cause wasted time when investigating test failures, and complicate release cycles / compatibility unnecessarily.

This change swaps to getting Proton from a checkout rather than submodule, and makes the proton version an env variable forming a matrix of sub-jobs to test different versions, such as master and last release as currently used here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ dist: trusty
 sudo: true
 language: c
 cache: ccache
+env:
+  - PROTON_VERSION=master BUILD_TYPE=Coverage
+  - PROTON_VERSION=0.26.0 BUILD_TYPE=RelWithDebInfo
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


gemmellr:
Ignore this comment (but not the PR), testing integration.

gemmellr:
Ignore this comment (but not the PR), testing integration again.

gemmellr:
Updated to only do the coverage build and reporting tasks on one of the sub-jobs

gemmellr:
Pushed this as it seems to work as desired, if folks have other improvements we can add them later.

PR #439, opened 2019-01-15
DISPATCH-1214: remove shutdown leak reported by Valgrind

kgiusti:
This leak is filling up the valgrind results with noise.

### DIFF
--- a/src/router_core/router_core.c
+++ b/src/router_core/router_core.c
@@ -116,7 +116,6 @@ void qdr_core_free(qdr_core_t *core)
     sys_mutex_free(core->id_lock);
     qd_timer_free(core->work_timer);
 
-
     //we can't call qdr_core_unsubscribe on the subscriptions because the action processing thread has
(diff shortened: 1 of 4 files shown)


src/router_core/router_core.c
```diff
@@ -173,6 +173,9 @@ void qdr_core_free(qdr_core_t *core)
         free(link->insert_prefix);
         free(link->strip_prefix);
         link->name = 0;
+        for (int i = 0; i < QDR_LINK_LIST_CLASSES; ++i) {
+            free_qdr_link_ref_t(link->ref[i]);
```
ganeshmurthy: instead of calling free_qdr_link_ref_t is it worth calling 
bool qdr_del_link_ref(qdr_link_ref_list_t *ref_list, qdr_link_t *link, int cls)

kgiusti: Good point - it would be better to clean up the ref lists propery.
I'll update this PR shortly...

PR #448, opened 2019-02-04
DISPATCH-1262: fix GCC 8.2 format-truncation error in router/src/main.c

jdanekrh:
Gcc warns because `getcwd` on Linux may allocate and return new buffer,
so don't ignore the returned value. And cur_path may in theory be null
going in if previous `calloc` failed.

It is unlikely we would end up with cur_path = NULL as parameter to `%s`
in a formatting string later below, but it is possible, given very
careful timing and sufficiently malicious user, I think.

Change is made to silence GCC warning, so that ``-fsanitize=undefined` compiles.


jdanekrh:
@ganeshmurthy thanks for looking into this

PR #449, opened 2019-02-16
DISPATCH-1268 - Set cmake_minimum_required to VERSION 2.8.12

jdanekrh:
A lot of changes from PR #288 is compatible with CMake 2.8.12 and can therefore still be used by the project.

This PR contains only the first change, to `cmake_minimum_required` version, to see if there is a chance to get it in.


ted-ross:
I think this update is OK.  The oldest downstream environment that I'm aware of is RHEL6, which appears to use cmake 2.8.12.

jdanekrh:
All right, I'll try to find more useful stuff to improve, given 2.8.12 features can be used.

jdanekrh:
I think this is as far as I am going to go at this point. There are still some tweaks in the original PR which I did not do here, so I might open new PR for that, but this PR seems to me has enough changes already.

I can squash the commits if you'd like.

jdanekrh:
I'll merge this later today if no one objects. Thanks.

PR #450, opened 2019-02-25
DISPATCH-1270 - Use CLOCK_MONOTONIC for purpose of calling pn_transport_tick

jdanekrh:


jdanekrh:
I've marked this as "draft" pending resolution of https://issues.apache.org/jira/browse/DISPATCH-1271 The LWS_CALLBACK_USER handler in http-libwebsockets.c is never triggered

jdanekrh:
Thanks for review. I'll commit this later today, then.

PR #451, opened 2019-02-25
NO-JIRA - Update doc for httpRootDir default value removal (DISPATCH-1155)

jdanekrh:


jdanekrh:
Found one more doc file to update. I'll squash and merge this later today if there are no further comments.

PR #452, opened 2019-02-28
DISPATCH-1273: enforce target policy for anonymous senders

grs:


grs:
@ted-ross @ganeshmurthy many thanks!

PR #453, opened 2019-03-06
DISPATCH-1278: initial support for prometheus metrics export

grs:


src/http-libwebsockets.c
```diff
@@ -361,6 +386,135 @@ static void connection_wake(qd_connection_t *qd_conn)
     }
 }
 
+static void handle_stats_results(void *context, qdr_global_stats_t *results)
+{
+    stats_t* stats = (stats_t*) context;
+    qd_http_server_t *hs = stats->server;
+    if (hs) {
+        work_t w = { W_HANDLE_STATS, stats->wsi };
+        work_push(hs, w);
```
ted-ross: It should be noted that this handler is being called on the core-thread.  Since work_push can block on a condition, the core thread is at risk for being blocked by the HTTP processing.  It might be better for the core-thread to schedule this callback as a general-work item so it runs on an I/O thread.

grs: Agreed. What's the call for doing that can you remind me?

ted-ross: Since the metrics consists only of counters, it's possible that they could be accessed synchronously by the I/O thread handling the metric request without using a lock.  There is the possibility that any given response might be inconsistent (i.e. being written at the same time as read), but that might be acceptable in this case.

grs: I believe the nature of the interaction means this will never happen. Each request uses a separate instance of the counters. Only the core thread updates those counters and they are only read by the http thread once the core thread has made the callback.

ted-ross: Sorry, I meant to not even use an action for the core thread.  Just go read the current values of the counters from the HTTP thread, ignoring any glitches that occur as a result of concurrency.

If you use the general-work method, the call is qdr_post_general_work.  I think there's a void* you can use to point to the metric object and you can add fields to qdr_general_work_t if needed.  You of course need to write the I/O-thread-side work function that calls the metric callback.

grs:
I've updated the patch to move the callback on to the general work queue as suggested. 

I've also made it work with older versions of libwebsockets (due to DISPATCH-1280).

One aspect of supporting the older libwebsockets api is that I can't use the lws_add_http_common_headers method (not available before 3 it seems). That function allowed the content length to be omitted (meaning we can stream the data rather than buffering it all before sending it) and marked the connection to be closed at the end of the data. Without this clients may wait for more data. I have tested with prometheus and it seems to cope with this fine. When testing with a web browser though there appears to be a one or two second delay in displaying the results.

One option is to add explicit chunking (libwebsockets has no built in support for that but it should be easy enough). Alternatively once DISPATCH-1280 is fixed we could require the later websockets library.

grs:
I believe the test failure - FAIL: test_13_mobile_address_edge_to_edge_one_interior (system_tests_edge_router.RouterTest) - is not related to this change.

PR #455, opened 2019-03-06
DISPATCH-1275 - Enabled deletion of connections based on connection i…

ganeshmurthy:
…d. Added related policy code. By default, a user does not have privileges to delete a connection. A policy needs to allow this. Inter-router connections can never be deleted


python/qpid_dispatch_internal/policy/policy_local.py
```diff
@@ -72,6 +72,7 @@ class PolicyKeys(object):
     KW_ALLOW_USERID_PROXY        = "allowUserIdProxy"
     KW_ALLOW_WAYPOINT_LINKS      = "allowWaypointLinks"
     KW_ALLOW_DYNAMIC_LINK_ROUTES = "allowDynamicLinkRoutes"
+    KW_ALLOW_AMIN_STATUS_UPDATE  = "allowAdminStatusUpdate"
```
ted-ross: This looks like a typo.

src/router_core/agent.c
```diff
@@ -414,7 +414,7 @@ static void qdr_manage_delete_CT(qdr_core_t *core, qdr_action_t *action, bool di
     case QD_ROUTER_CONFIG_ADDRESS:    qdra_config_address_delete_CT(core, query, name, identity); break;
     case QD_ROUTER_CONFIG_LINK_ROUTE: qdra_config_link_route_delete_CT(core, query, name, identity); break;
     case QD_ROUTER_CONFIG_AUTO_LINK:  qdra_config_auto_link_delete_CT(core, query, name, identity); break;
-    case QD_ROUTER_CONNECTION:        break;
+    case QD_ROUTER_CONNECTION:        qdra_connection_delete_CT(core, query, identity); break; //Connection objects are created by the router and they have an id and no name
```
ted-ross: The forced closing of a connection is supposed to be signalled by setting the adminStatus to "deleted", not by using the DELETE method in AMQP management.

PR #459, opened 2019-03-10
DISPATCH-1282 - Support for building on macOS

jdanekrh:
* macOS does not use -rt linker option
* pthread_t on macOS is a typedef’d struct, not long
* dlopen on macOS substitutes rpath for @rpath placeholder, it does not prepend it as on Linux

~~CMake is now configured to append rpath to all libraries linked into targets. This has also effect on Linux.~~ edit: this was merged separately as DISPATCH-1321.

With this patch, I compiled dispatch and started it on macOS without errors

$ cmake .. -DProton_DIR=/Users/jdanek/projects/qpid-proton/build/install/lib/cmake/Proton -DCMAKE_INSTALL_PREFIX=install
$ make install
$ PYTHONPATH=install/lib/python2.7/site-packages/ install/sbin/qdrouterd

~~The tests are not yet passing. Some of it is because I apparently did not have OpenSSL installed, some because `saslpasswd2` is not present, and some I have no clue about (like the segfault).~~

edit: see comments/travis for current list of still-broken tests on macOS


jdanekrh:
I've created StackOverflow question about what to do regarding SASL on macOS, because I have no clue myself. https://stackoverflow.com/questions/55240554/what-is-the-status-of-sasl-on-macos-in-2019

ldebello:
Hi guys, I am trying to execute the router in osx, did you have luck?

jdanekrh:
@ldebello I did! With the changes in the PR, the router started just fine and passed most of it's testsute (everything except sasl and ssl tests). I believe if I then just installed openssl and cyrus sasl using brew, it would all work, but I only realized this afterwards and did not try.

If you are willing to try yourself, maybe applying this patch and seeing what happens, I can try to help with any specific issues. It would be fun to get this working. The problem is, macOS is not my primary machine and so far nobody else really wanted to get dispatch running there.

ldebello:
@jdanekrh thanks for the answer, I will try to check this and I will let you know.

Regards,
Luis

jdanekrh:
@ldebello Any luck with QDR on macOS?

ldebello:
@jdanekrh sorry, I have it in my list but I haven't had time to check it

jdanekrh:
I got it to work on my machine, but I cannot do the same on Travis CI. Probably some silly oversight on my part. I'll try look into it more.

On my machine, the dispatch tests end up like this

```
89% tests passed, 6 tests failed out of 57

Total Test time (real) = 1484.23 sec

The following tests FAILED:
          9 - unit_tests (SEGFAULT)
         28 - system_tests_sasl_plain (Failed)
         38 - system_tests_auth_service_plugin (Failed)
         39 - system_tests_authz_service_plugin (Failed)
         48 - system_tests_bad_configuration (Failed)
         50 - system_tests_edge_router (Timeout)
Errors while running CTest
```

jdanekrh:
(Do new commits here send any notifications to people? I haven't noticed any notifications for commits, only for comments.)

In any case, the previous problem was `DYLD_FALLBACK_LIBRARY_PATH` which macOS ignores, unless a security feature is turned off, which I did for myself, but is on on Travis. Now the tests run on Travis as well.

jdanekrh:
Regarding the future of the PR, I'd like to clean it up and to get it merged even before all macOS tests are green. That means, either disable failing tests on `ctest` command line, or set the macOS build as "allowed to fail", https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail.

RoddieKieley:
@jdanekrh Great to see this getting done! Just back today after a two week vacation and will add this to the 'todo'. It will likely be later this week / into the weekend when I get a chance to look if all goes well.

jdanekrh:
I think I've figured out the segfault in tests. It is problem with tests. The test provides its own fake `qd_server_timeout` in `timeout_tests.c` which must be used during the test, but under macOS, the "real" version in `server.c` is picked up. On Linux, the test version is used because of linking with `-rdynamic`. On macOS, this option is not present (looks to be noop?), so it has to be achieved differently. Don't know how.

```
Process 4806 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x28)
    frame #0: 0x000000010013cec4 libqpid-dispatch.dylib`qd_server_timeout(server=0x0000000000000000, duration=0) at server.c:1511:37 [opt]
   1508 }
   1509
   1510 void qd_server_timeout(qd_server_t *server, qd_duration_t duration) {
-> 1511     pn_proactor_set_timeout(server->proactor, duration);
   1512 }
   1513
   1514 qd_dispatch_t* qd_server_dispatch(qd_server_t *server) { return server->qd; }
Target 0: (unit_tests) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x28)
  * frame #0: 0x000000010013cec4 libqpid-dispatch.dylib`qd_server_timeout(server=0x0000000000000000, duration=0) at server.c:1511:37 [opt]
    frame #1: 0x000000010013ed69 libqpid-dispatch.dylib`qd_timer_schedule(timer=0x000000010033c298, duration=<unavailable>) at timer.c:157:5 [opt]
    frame #2: 0x00000001000059ff unit_tests`timer_tests [inlined] test_immediate at timer_test.c:66:5 [opt]
    frame #3: 0x00000001000059db unit_tests`timer_tests at timer_test.c:313 [opt]
    frame #4: 0x0000000100003eb1 unit_tests`main(argc=<unavailable>, argv=0x00007ffeefbff898) at run_unit_tests.c:57:15 [opt]
    frame #5: 0x00007fff70cd3ed9 libdyld.dylib`start + 1
    frame #6: 0x00007fff70cd3ed9 libdyld.dylib`start + 1
```

jdanekrh:
Asked a SO question about the `-rdynamic`. I think I have it figured out, except I don't know how to avoid compiling with `-fno-inline`, which is a problem. https://stackoverflow.com/questions/56711149/what-is-the-rdynamic-linker-option-equivalent-on-macos.

Anyways, I'd like to have a separate Jira and PR for fixing the tests, as I've mentioned.

RoddieKieley:
@jdanekrh I got your branch up and running on macOS 10.14/Xcode 10.2 and OSX 10.11.6/Xcode 7.3.1 with relatively little problem outside of forgetting what the PYTHONPATH needed to be. 

That being said neither of my operating system installs had python-unittest2, I'm using MacPorts and I have py-unittest2 and variants whereas my fedora box has python-unittest2 variants. Will get unittest2 up and running and see about the tests.

RoddieKieley:
@jdanekrh I've been looking closely at the unittest2 issues I've been having. It turns out that there are differences in behaviour and immediate availability of unittest2 when installed via python -m easy_install, pip install, or port install py-unittest2.

Of the three only the first works here as per the travis setup that installs the egg to the main system /Library/Python/2.7 path. While this is probably only really a pathing issue, it would be nice if this were less fragile.

Once the tests run I see that there are indeed issues from a default cmake configuration as well as a segfault during the unit tests. More investigation required.

jdanekrh:
@RoddieKieley Thank you for the review. I've opened followup Jira for fixing tests https://issues.apache.org/jira/browse/DISPATCH-1384

Unless anyone complains, I'd merge this PR and then continue fixing the tests in subsequent PRs.

jdanekrh:
@RoddieKieley I spent some time with the build issues: python interpreter and libs detection, qpid proton python on macos, and unittest2.

It turned out that I had the least issues when I installed Python 3.7 from MacPorts and put it on PATH. CMake then never failed to resolve both interpreter and libs correctly.

Regarding unittest2, I put creating a virtual environment with it into .travis.yml. I think that is a reasonable way for actual users/developers to work with it.

I did not find a way to get rid of the `install_name_tool -add_rpath` hack to be able to `import proton`.

RoddieKieley:
@jdanekrh Sounds like it's about as good as it can be for the moment. I think the unittest2 issue resolution via a virtual environment was a reasonable way to go.

Going to grab master now and see how it goes here.

PR #470, opened 2019-03-19
[dummy] Dispatch 1288 policy for router-initiated connector connections

ChugR:
This PR is a dummy and not intended to be merged. It shows the mission code as it is intended but it has not yet been tested thoroughly, seen internal self tests, or been documented.


ChugR:
Dummy PR closed as better PR is available

PR #472, opened 2019-03-19
Dispatch 1288 1 - policy for outbound connectors

ChugR:
This PR includes the mission code and supporting self test.
This PR does not include docs/book writeup for the feature

### DIFF
--- a/src/policy.c
+++ b/src/policy.c
@@ -665,6 +665,9 @@ bool _qd_policy_approve_link_name(const char *username, const char *allowed, con
         return false;
     }
 
+    if (!username)
+        username = "";
+
     size_t username_len = strlen(username);
 
     // make a writable, disposable copy of the csv string
@@ -781,6 +784,9 @@ bool _qd_policy_approve_link_name_tree(const char *username, const char *allowed
         // no names in 'allowed'.
         return false;
     }
+
+    if (!username)
+        username = "";
 
     size_t username_len = strlen(username);
     size_t usersubst_len = strlen(user_subst_key);
@@ -1155,7 +1161,6 @@ void qd_policy_amqp_open_connector(qd_connection_t *qd_conn) {
     if (policy->enableVhostPolicy &&
         (!qd_conn->role || !strcmp(qd_conn->role, "normal") || !strcmp(qd_conn->role, "route-container"))) {
         // Open connection or not based on policy.
-        pn_transport_t *pn_trans = pn_connection_transport(conn);
         const char *hostip = qd_connection_remote_ip(qd_conn);
         const char *conn_name = qd_connection_name(qd_conn);
         uint32_t conn_id = qd_conn->connection_id;
@@ -1173,11 +1178,10 @@ void qd_policy_amqp_open_connector(qd_connection_t *qd_conn) {
             if (qd_policy_open_fetch_settings(policy, qd_conn->user_id, hostip, policy_vhost, conn_name,
                                         POLICY_VHOST_GROUP, conn_id,
                                         qd_conn->policy_settings)) {
-                // Apply transport policy settings
-                if (qd_conn->policy_settings->maxFrameSize > 0)
-                    pn_transport_set_max_frame(pn_trans, qd_conn->policy_settings->maxFrameSize);
-                if (qd_conn->policy_settings->maxSessions > 0)
-                    pn_transport_set_channel_max(pn_trans, qd_conn->policy_settings->maxSessions - 1);
+                // It's too late to apply transport policy settings as the local
+                // AMQP Open has already been sent.
+                // TODO: Apply transport max_frame and channel_max to outgoing
+                //       connector Open if policy is enabled.
             } else {
                 // Failed to fetch settings
                 qd_log(policy->log_source,
--- a/src/policy.c
+++ b/src/policy.c
@@ -1182,6 +1182,9 @@ void qd_policy_amqp_open_connector(qd_connection_t *qd_conn) {
                 // AMQP Open has already been sent.
                 // TODO: Apply transport max_frame and channel_max to outgoing
                 //       connector Open if policy is enabled.
+
+                // Count senders and receivers for this connection
+                qd_conn->policy_counted = true;
             } else {
                 // Failed to fetch settings
                 qd_log(policy->log_source,
--- a/src/policy.c
+++ b/src/policy.c
@@ -66,6 +66,12 @@ static const char * const user_subst_i_embed    = "e";
 static const char * const user_subst_i_suffix   = "s";
 static const char * const user_subst_i_wildcard = "*";
 
+//
+// Fixed vhost policy usergroup used when storing connector policy.
+// The connector attribute 'policyVhost' defines a vhost and within
+// that vhost the connector policy values are in '$connector'.
+static const char * const POLICY_VHOST_GROUP = "$connector";
+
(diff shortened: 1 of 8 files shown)


src/policy.c
```diff
@@ -404,18 +409,47 @@ bool qd_policy_open_lookup_user(
...
+    const char *username,
+    const char *hostip,
+    const char *vhost,
+    const char *conn_name,
+    const char *name_buf,
+    uint64_t    conn_id,
+    qd_policy_settings_t *settings)
+{
+    bool res = false;
+    qd_python_lock_state_t lock_state = qd_python_lock();
```
ganeshmurthy: Is it possible to pre-populate the settings into a linked list and just search thru the list instead of taking out this global python lock every time during response open processing ?

src/policy.c
```diff
@@ -404,18 +409,47 @@ bool qd_policy_open_lookup_user(
...
+ * @param[out] settings pointer to settings object to be filled with policy values
+ **/
+bool qd_policy_open_fetch_settings(
+    qd_policy_t *policy,
+    const char *username,
+    const char *hostip,
+    const char *vhost,
+    const char *conn_name,
+    const char *name_buf,
+    uint64_t    conn_id,
```
ganeshmurthy: username, conn_id, vhost, conn_name is not used in this function ?

src/policy.c
```diff
@@ -1110,6 +1152,62 @@ void qd_policy_amqp_open(qd_connection_t *qd_conn) {
 }
 
 
+void qd_policy_amqp_open_connector(qd_connection_t *qd_conn) {
```
ganeshmurthy: Should this function be called something like qd_policy_populate_conn_policy_settings or something like that. The connector has already been "opened"

src/policy.c
```diff
@@ -1110,6 +1152,62 @@ void qd_policy_amqp_open(qd_connection_t *qd_conn) {
...
+        }
+
+        if (strlen(policy_vhost) > 0) {
+            // This connector connection is controlled by policy.
+            if (qd_policy_open_fetch_settings(policy, qd_conn->user_id, hostip, policy_vhost, conn_name,
+                                        POLICY_VHOST_GROUP, conn_id,
+                                        qd_conn->policy_settings)) {
+                // It's too late to apply transport policy settings as the local
+                // AMQP Open has already been sent.
+                // TODO: Apply transport max_frame and channel_max to outgoing
```
ganeshmurthy: This TODO is going to be done as a part 2 of this functionality ?

src/policy.c
ChugR: Well spotted. These are left over from when there was a log message that used them.

src/policy.c
```diff
@@ -425,34 +459,39 @@ bool qd_policy_open_lookup_user(
...
-                    settings->maxSenders           = qd_entity_opt_long((qd_entity_t*)upolicy, "maxSenders", 0);
-                    settings->maxReceivers         = qd_entity_opt_long((qd_entity_t*)upolicy, "maxReceivers", 0);
-                    if (!settings->allowAnonymousSender) { //don't override if enabled by authz plugin
-                        settings->allowAnonymousSender = qd_entity_opt_bool((qd_entity_t*)upolicy, "allowAnonymousSender", false);
-                    }
-                    if (!settings->allowDynamicSource) { //don't override if enabled by authz plugin
-                        settings->allowDynamicSource   = qd_entity_opt_bool((qd_entity_t*)upolicy, "allowDynamicSource", false);
+                    int truthy = PyObject_IsTrue(result2);
+                    if (truthy) {
+                        settings->maxFrameSize         = qd_entity_opt_long((qd_entity_t*)upolicy, "maxFrameSize", 0);
```
ganeshmurthy: Since we are not sending the max-frame-size in the first outgoing open frame, should we populate settings->maxFrameSize which will start enforcing it when the frames start coming in.

src/policy.c
ChugR: The connector connection hasn't really opened yet. In this function policy is approving or denying the connection. Sure, the second Open has arrived but policy may slam the connection closed for some reason. The name is similar to the peer function that handles the remote open for incoming connections.

src/policy.c
```diff
@@ -404,18 +409,47 @@ bool qd_policy_open_lookup_user(
...
+    const char *conn_name,
+    const char *name_buf,
+    uint64_t    conn_id,
+    qd_policy_settings_t *settings)
+{
+    bool res = false;
+    qd_python_lock_state_t lock_state = qd_python_lock();
+    PyObject *module = PyImport_ImportModule("qpid_dispatch_internal.policy.policy_manager");
+    if (module) {
         res = false;
```
ganeshmurthy: res is already false ? (initialized to false)

src/policy.c
ChugR: There are duplicate settings for the same AMQP values in connector and in vhostUserGroupSettings. At the point of this comment it is too late to for the vhostUserGroupSettings value to be applied. This sounds like a need to document which settings from the user group get applied.

src/policy.c
ChugR: true. it's a wasted line of code as it gets optimized away.

src/policy.c
```diff
@@ -1083,15 +1119,21 @@ void qd_policy_amqp_open(qd_connection_t *qd_conn) {
...
+            if (qd_policy_open_fetch_settings(policy, qd_conn->user_id, hostip, vhost, conn_name,
+                                        settings_name, conn_id,
+                                        qd_conn->policy_settings)) {
+                if (qd_conn->policy_settings->maxFrameSize > 0)
+                    pn_transport_set_max_frame(pn_trans, qd_conn->policy_settings->maxFrameSize);
+                if (qd_conn->policy_settings->maxSessions > 0)
+                    pn_transport_set_channel_max(pn_trans, qd_conn->policy_settings->maxSessions - 1);
+            } else {
+                // failed to fetch settings
+                connection_allowed = false;
```
ganeshmurthy: so if there was some issue fetching the policy settings from the python side, the connection will be closed ? Is there a way we can try to pre load the policy settings and warn if that fails? Please ignore this comment if this is already happening.

ChugR: The common pattern in dispatch C code is if any qd_entity_get call fails then the entity in question is failed. Same goes for connector connection policy: if the policy does not load then that connection is denied. Pre-loading the policy is a separate issue. Suppose there was a preload that failed then all the connections using that failed preload must fail as well.

src/policy.c
ChugR: [DISPATCH-1298](https://issues.apache.org/jira/browse/DISPATCH-1298)

src/policy.c
ChugR: This function is shared by the incoming connection and outgoing connection handlers. The incoming connection handler needs it.

src/policy.c
ChugR: fixed at 7baea

src/policy.c
ChugR: Resolved at 7baea. In outbound connections all the transport settings are from the connector and none come from policy.

src/policy.c
ChugR: Fixed at 7baea.

src/policy.c
ChugR: I'm leaving the name as-is. It looks good at the call site in container.c.

There are two hard problems in software engineering: cache invalidation, naming things, and off-by-one errors.

PR #475, opened 2019-03-21
Remove /stand-alone from console install dir. 

ErnieAllen:
This patch removes the last part of the console install directory: /stand-alone.
In addition, the httpRoot config file entry for a listener is now optional. The router will attempt to serve the console from the new install dir.

To disable serving the console on an http enabled listener, set the httpRoot attribute to 'None'.

This PR also updated the readme files for the console and the httpRoot attribute description in the schema.


grs:
Looks good to me!

PR #476, opened 2019-03-23
DISPATCH-1295 - Reduce lock contention and make fewer system calls.

cliffjansen:
Flamegraphs show on a single quiver run that there is lock contention in setting timers.  Locks are held across system calls and also cause thread stalls as seen in offcpu runs.

This change restricts the main timer lock to updating internal data structures.  A second lock is used to prevent setting timer values out of order.

The main benefit is that immediate timer calls have much less contention with non-immediate timers.

Using a standard quiver C client run of 9,000,000 message, which lasts about a minute, we get 19% of the offcpu related to qd_timer_schedule and qd_timer_visit.  This drops to 0 with the change.

Regular on-cpu profiling samples in qd_timer drop from 9.1% of samples to 0.2%.  Samples in qd_immediate drop modestly from 0.6% to 0.5%.

### DIFF
--- a/src/immediate.c
+++ b/src/immediate.c
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+#include "dispatch_private.h"
 #include "immediate_private.h"
 #include "server_private.h"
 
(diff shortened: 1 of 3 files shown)


cliffjansen:
Since this pull request builds on top of DISPATCH-1274, and that has been removed for the next release, I am closing this for now.  Next steps TBD.

PR #477, opened 2019-03-24
DISPATCH-1301 - Updated test test_18_management_not_implemented to wa…

ganeshmurthy:
…it for a couple of seconds to get receiver.remote_source.address

### DIFF
--- a/tests/system_tests_one_router.py
+++ b/tests/system_tests_one_router.py
@@ -717,6 +717,18 @@ class SemanticsMulticast(MessagingHandler):
         Container(self).run()
 
 
+class CheckAddressTimer(object):
+    def __init__(self, parent):
+        self.parent = parent
+
+    def on_timer_task(self, event):
+        if self.parent.receiver.remote_source.address:
+            self.parent.cancel_custom()
+            self.parent.request.reply_to = self.parent.receiver.remote_source.address
+            self.parent.sender.send(self.parent.request)
+        else:
+            self.parent.cancel_custom()
+
 class ManagementNotImplemented(MessagingHandler):
     def __init__(self, address):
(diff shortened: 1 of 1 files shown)


ChugR:
This may get the test to pass but there has been a bad behavior change somewhere in the router.
"on_sendable" now means "start a timer and when that expires then you can send". Is that the case? If so then that will break customer code as surely as it broke the self tests.

ganeshmurthy:
I was able to reproduce this test failure in version 1.5.0. This is not a regression.

ganeshmurthy:
Here is the output of the test failure when running the same test in v1.5.0

[gmurthy@localhost build]$  /usr/bin/python "/home/gmurthy/opensource/qpid-dispatch/build/tests/run.py" "-m" "unittest" "-v" "system_tests_one_router.OneRouterTest.test_18_management_not_implemented"
test_18_management_not_implemented (system_tests_one_router.OneRouterTest) ... FAIL

======================================================================
FAIL: test_18_management_not_implemented (system_tests_one_router.OneRouterTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gmurthy/opensource/qpid-dispatch/tests/system_tests_one_router.py", line 239, in test_18_management_not_implemented
    self.assertEqual(None, test.error)
AssertionError: None != u'No response received for management request'

----------------------------------------------------------------------
Ran 1 test in 60.176s

FAILED (failures=1)
[gmurthy@localhost build]$

ted-ross:
Don’t use a timer.  You can get the reply address in the on_link_opened event.  There are many examples of this in the tests.

ganeshmurthy:
Closing this PR. Will be using on_link_opened as @ted-ross suggested. Will open a new PR with the changes.

PR #488, opened 2019-04-10
DISPATCH-1310: move qdr_delivery_t API to its own source files

kgiusti:
This patch merely moves delivery-related code and does some renaming of functions to adhere to the "qdr_delivery_*" naming convention.  

No changes to actual functionality have occurred - it is simply a reorganization of code layout.


kgiusti:
Ooops wrong branch!

PR #492, opened 2019-04-19
DISPATCH-1310: refactor the receive handler code

kgiusti:
Refactor the rx handler to exit early if the message is being
discarded or the delivery has already been handed to the core thread.

Avoid calling the receive code entirely if the PN_DELIVERY being read
does not contain the links current delivery.

### DIFF
--- a/src/container.c
+++ b/src/container.c
@@ -177,9 +177,8 @@ static void handle_link_open(qd_container_t *container, pn_link_t *pn_link)
 }
 
 
-static void do_receive(pn_delivery_t *pnd)
+static void do_receive(pn_link_t *pn_link, pn_delivery_t *pnd)
 {
-    pn_link_t     *pn_link  = pn_delivery_link(pnd);
     qd_link_t     *link     = (qd_link_t*) pn_link_get_context(pn_link);
 
     if (link) {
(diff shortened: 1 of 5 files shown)


src/container.c
```diff
@@ -642,8 +641,8 @@ void qd_container_handle_event(qd_container_t *container, pn_event_t *event,
         delivery = pn_event_delivery(event);
         pn_link  = pn_event_link(event);
 
-        if (pn_link_is_receiver(pn_link))
-            do_receive(delivery);
+        if (pn_delivery_readable(delivery))
```
ganeshmurthy: If you look at commit 417975e17406dcc48a648d9fc6da38651a4ff956, the pn_delivery_readable was changed to pn_link_is_receiver. Please talk to @ted-ross  about this just to make sure we don't have a regression.

kgiusti: Turns out using pn_delivery_readable massively decreases the number calls to do_receive() that do nothing (i.e. check if delivery is current and returns since it isn't) in the case of single frame deliveries.

The way do_receive() works is that loops over all deliveries that have rx complete starting with the links current delivery and advancing until it either exhausts the links deliveries or it hits a delivery that does not have rx data.

This means that it processes all complete deliveries on receipt of the first PN_DELIVERY event from proton.

That means all other complete rx has a "stale" PN_DELIVERY event on the event queue, which leads to the useless calls to do_receive.

I'd like to keep this change unless as you mentioned there is some edge condition it breaks that's not obvious from the code (and should be).

PR #493, opened 2019-04-26
DISPATCH-1322 - Prevent credit starvation in router-to-router links.

ted-ross:
Added protection for deliveries held in link processing outside the lock.


ChugR:
+1
Passes self test and thousands of passes of my stand-alone test case.

PR #494, opened 2019-04-29
DISPATCH-1328 Added deliveries-delayed stats to /metrics response

ErnieAllen:
These stats are needed by graphana to set an alert


grs:
Are the formatting changes/fixes deliberate? If so it might be better to split them into a separate commit at least (if not indeed a separate PR) for clarity.

The actual logical changes lgtm though.

PR #496, opened 2019-05-01
DISPATCH-1325 - Treat newly attached outgoing links as stalled since …

ted-ross:
…they have not yet received initial credit.  This will prevent stalling of deliveries that arrive before the link is fully attached.


ChugR:
+1
From a bash script I start the edge router, wait one second for it to boot, and then connect to it. After a few seconds the test runs normally. This satisfies the original issue in DISPATCH-1325.

PR #503, opened 2019-05-08
DISPATCH-974 - Added code to qdstat and qdmanage clients to repeatedl…

ganeshmurthy:
…y get rows to overcome the limitations imposed by PROTON-1846

### DIFF
--- a/python/qpid_dispatch/management/client.py
+++ b/python/qpid_dispatch/management/client.py
@@ -225,12 +225,61 @@ class Node(object):
         @keyword count: A count of the maximum number of results to return.
         @return: A L{QueryResponse}
         """
-        request = self.node_request(
-            {u'attributeNames': attribute_names or []},
-            operation=u'QUERY', entityType=type, offset=offset, count=count)
 
-        response = self.call(request)
-        return Node.QueryResponse(self, response.body[u'attributeNames'], response.body[u'results'])
+        # There is a bug in proton (PROTON-1846) wherein we cannot ask for
+        # too many rows. So, as a safety we are going to ask only for
+        # MAX_ALLOWED_COUNT_PER_REQUEST. Since this is used by both qdstat
+        # and qdmanage, we have determined that the optimal value for
+        # MAX_ALLOWED_COUNT_PER_REQUEST is 700
+        MAX_ALLOWED_COUNT_PER_REQUEST = 700
+
+        response_results = []
+        response_attr_names = []
+        if offset is None:
+            offset = 0
+
+        if count is None:
+            # count has not been specified. For each request the
+            # maximum number of rows we can get without proton
+            # failing is MAX_ALLOWED_COUNT_PER_REQUEST
+            request_count = MAX_ALLOWED_COUNT_PER_REQUEST
+        else:
+            if count > MAX_ALLOWED_COUNT_PER_REQUEST:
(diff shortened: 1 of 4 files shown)


ChugR:
+1 the implementation looks OK and the tests pass.
The technique used here, pulling successive slices on repeated calls, leaves room for the underlying list in proton to change between calls. This could lead to duplicate or missing entries. Is there a place to put doc text explaining this so that users' expectations are properly framed when the lists are longer than the 700 entries?

python/qpid_dispatch/management/client.py
```diff
@@ -225,12 +225,61 @@ def query(self, type=None, attribute_names=None, offset=None, count=None):
...
+        if offset is None:
+            offset = 0
+
+        if count is None:
+            # count has not been specified. For each request the
+            # maximum number of rows we can get without proton
+            # failing is MAX_ALLOWED_COUNT_PER_REQUEST
+            request_count = MAX_ALLOWED_COUNT_PER_REQUEST
+        else:
+            if count > MAX_ALLOWED_COUNT_PER_REQUEST:
```
kgiusti: nit: more "pythonic":   count = min(MAX_ALLOWED_COUNT_PER_REQUEST, count)

python/qpid_dispatch/management/client.py
```diff
@@ -225,12 +225,61 @@ def query(self, type=None, attribute_names=None, offset=None, count=None):
...
+                {u'attributeNames': attribute_names or []},
+                operation=u'QUERY', entityType=type, offset=offset,
+                count=request_count)
+
+            response = self.call(request)
+
+            if not response_attr_names:
+                response_attr_names += response.body[u'attributeNames']
+
+            response_results += response.body[u'results']
```
kgiusti: You might consider using a map indexed by entity id here as a way to filter out duplicates.

tests/system_tests_qdmanage.py
```diff
@@ -515,5 +514,66 @@ def test_create_delete_address_pattern(self):
...
                     self.assertNotEqual(p[0], pattern)
 
+    def test_yy_query_many_links(self):
+        # This test will fail without the fix for DISPATCH-974
+        c = BlockingConnection(self.address())
+        count = 0
+        links = []
+        COUNT = 5000
+
+        print ("Creating 10,000 links, please wait")
```
kgiusti: These prints won't necessarily appear on the console since stdout is by default buffered.

You can add a call to sys.stdout.flush() after each print to force the output immediately.

Personally I'm not a big fan of printing debug stuff during the tests - unless you can figure out a way to only do it when ctest's "-V" flag has be specified.

ganeshmurthy: Removed all prints, they are GONE!

python/qpid_dispatch/management/client.py
ganeshmurthy: request_count = min(MAX_ALLOWED_COUNT_PER_REQUEST, count) 

done!

python/qpid_dispatch/management/client.py
ganeshmurthy: I discussed this with @ted-ross . The new entries to the linked list are added at the tail. So no need to remove duplicates. It might show stale data but so does the previous approach

kgiusti: 1

PR #505, opened 2019-05-14
NO-JIRA: add tests that update and settle deliveries before rx completes

kgiusti:


kgiusti:
Doesn't work - abandoned

PR #508, opened 2019-05-24
DISPATCH-1342 Replaced mutex with spin lock on qd_message content

franz1981:


src/message.c
```diff
@@ -933,7 +933,7 @@ void qd_message_free(qd_message_t *in_msg)
         // counts starting with the msg cursor.  If the buffer count drops to
         // zero we can free it.
         //
-        LOCK(content->lock);
```
ted-ross: Why not add the '&' to the macro rather than having to add '&' to all of the invocations?

franz1981: Done!  Thanks :+1:

franz1981:
@astitcher Totally agree, that's why I haven't replaced the mutexes for any possible use case; there are some requirements that need to be satisfied in order to make this change effective (or feasable).
We need to:

- be sure no reentrancy is happening on the mutual exclusion zone (and it seems the case, looking the code)
- be sure that the mutual exclusion zone is not held for too much time (to be defined) and none is blocked on it

If both these requirements are met, only proper benchmarking and tests can tell us if the spin lock is the right answer here. I would stick to something that won't require OS arbitrations on the hot path anyway: OS is like burocrazy and I won't bother it when user-space matters can already solve things with a more predictable latencies

astitcher:
@franz1981 All true, but I think the point about using *realtime scheduling* is important. And making dispatch use realtime scheduling will be 'non-trivial'

franz1981:
@astitcher 
I believe that the doc of spin lock is excessively conservative TBH (with good reasons, given how many archs it has to cover), but it gives nice hints to make things more resilient on not properly configured box/limited machines.
Although not written on the stone, the router need at least 2 HW threads to run properly (1 core + 1 I/O worker) and it's true that a single core machine with HyperThread will suffer by using spin locks.  Same things can be said if you run the router configuring (much) more threads then real cores...
In those cases I'm not quite sure the router will perform optimally, but makes sense to account for them too: in all the other cases, assuming the previous correctness conditions to be met, I believe using spin locks would be just a win resource-wise, given the peculiar design of the router (no thread pools, just fixed size thread). 

I suggest, if not possible to accept this PR, to consider pooling mutexes to reduce the allocation/deallocation rate for such heavyweight kernel resources.

franz1981:
On my tests with 9 workers I've gotten an improvement from 73 K msg/sec -> 80 K msg/sec, but only if applied over https://github.com/apache/qpid-dispatch/pull/537

franz1981:
@astitcher @kgiusti That's the reason why I strongly suggest to replace mutexes on the router: https://twitter.com/forked_franz/status/1219241412497158144

PR #510, opened 2019-05-24
DISPATCH-1343 Faster qd_buffer and qd_message allocations

franz1981:
It contains an additional commit from https://github.com/apache/qpid-dispatch/pull/508, referenced here as required: this branch will be rebased over master if the latter will be merged, so do not merge this one yet.
This PR has been sent only for reviewing purposes.

### DIFF
--- a/include/qpid/dispatch/atomic.h
+++ b/include/qpid/dispatch/atomic.h
@@ -58,6 +58,10 @@ static inline int32_t sys_atomic_set(sys_atomic_t *ref, uint32_t value)
     return atomic_exchange(ref, value);
 }
 
+static inline void sys_atomic_release_fence()
+{
+    atomic_thread_fence(memory_order_release);
+}
 
 static inline void sys_atomic_destroy(sys_atomic_t *ref) {}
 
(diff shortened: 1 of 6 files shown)


include/qpid/dispatch/atomic.h
```diff
@@ -203,6 +223,14 @@ static inline void sys_atomic_destroy(sys_atomic_t *ref)
     sys_mutex_free(ref->lock);
 }
 
+static inline void sys_atomic_release_fence()
+{
+    //it shoudn't leak any OS resources, because (unshared) spin locks
+    //are just user space atomic ints
+    sys_spinlock_t lock;
+    sys_spin_init(&lock);
```
franz1981: This can be seen here: https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/sysdeps/x86_64/nptl/pthread_spin_unlock.S#L22

franz1981:
I've spent some time looking at the perf of this changes and it seems that the new way to allocate qd_buffer without issuing a full memory barrier is slower then the original one, because I was expecting the latter to use C11 variant of sys_atomic_init in atomic.h, while instead it was using the GNU version.
The point is that some of atomic operation doesn't seem correct to me from a memory model point of view and that's why the original version (that was just doing a plain store!) was faster then this one, that was issuing a compiler barrier for correctness.

franz1981:
I will close this one because such safe initalization isn't needed according to sys_atomic_init implementation for GCC.

PR #513, opened 2019-05-29
DISPATCH-1346: Create doc for prioritized message delivery

bhardesty:
This describes how Dispatch Router handles message deliveries when messages have priority levels assigned.

@mgoulish can you please review and make sure the content is accurate technically (and that I'm not missing anything that might be important)?

### DIFF
--- a/docs/books/user-guide/routing.adoc
+++ b/docs/books/user-guide/routing.adoc
@@ -321,6 +321,22 @@ If a producer sends an unsettled delivery, the disposition may be ACCEPTED or RE
 
 |===
 
+[id='prioritized-message-delivery']
+=== Prioritized Message Delivery
+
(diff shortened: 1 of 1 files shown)


mgoulish:
* priority range shall be between 0 and 9. Neither less than 0 nor greater than 9 shall it be, but 0-9 shall be the range of the priority.

* there are two ways that a user can make a message get a priority.   One is by assigning the message a priority directly. The other is by sending the message to an address that has a priority assigned to it.    In the Dispatch Router config file, when you are defining an address, you can stick a priority descriptor in there.  And if a message with a priority gets sent to an address with a priority, then the address priority wins.

* priority only affects how messages are processed on inter-router links between interior routers.

* Higher number are higher priority. It's not like golf.

* The reason the default  priority is 4 is so that you can set some message to be *lower* than default if you want to.

* If you set a message's priority to 0 we just don't bother sending it at all. Who cares? Instead we break it down into its constituent 1s and 0s which are then used as lubricant in the code.

* That last point ^^^ is completely false. Sorry.  Low blood sugar.

PR #514, opened 2019-05-30
DISPATCH-1349 qd_buffer_list_clone is wasting dst buffers available capacity

franz1981:


ChugR:
This is not a general purpose function called for complete messages.
This function is called when message routing annotation fields (to_override, trace, ingress)  are copied as a message is sent to multiple places. Each field may be overwritten with different information as it is sent to each place; or it may not.
My bet is that the vast majority of the copies involve only one buffer. In the event that a copy goes on to a second buffer that buffer will (almost) always hold the rest of the field. The change would coalesce two buffers into one only very rarely.
I'm not downvoting this change, just wondering how we could measure the benefit.

franz1981:
@ChugR Totally agree. I was (wrongly) expecting this method to often have >= 1 buffer to be copied, but I see that is not the case.
That means that the most of the cost of this method comes by just allocating the dst buffer and touching the src buffer (that seems to always be "cold" cache-wise).

franz1981:
Given that it would make the code more complex I can close this: considering the context of utilization it won't bring any improvement :+1:

ganeshmurthy:
My first question which is not related to the contents of this PR is why is this function returning an unsigned int ? The returned value is not used anywhere inside the actual router source code.

ganeshmurthy:
My next observation is that the msg->ma_to_override, msg->ma_trace and msg->ma_ingress  always has at least one buffer associated with them even in the one router single sender, single receiver case.
If we have only one router and the sender is not sending in any annotations and there is only one receiver on the same standalone router, why are these buffer populated in the first place and since they are populated, we are performing the extra overhead of copying them. If a message visits only one router in its lifetime, this seems like a waste of time

ganeshmurthy:
This PR looks good and can be committed once @ted-ross has had a chance to look at it.

ganeshmurthy:
> Given that it would make the code more complex I can close this: considering the context of utilization it won't bring any improvement +1

Say a message has 4 router hops on its way from sender to receiver, will there ever be more than one buffer worth of each annotation (to_override or trace or ingress)  given each of them get a buffer size of 512 bytes. @franz1981 is trying to compact these buffers but it comes into play *only* if there is more than one buffer. In 99% percent of the cases one buffer is more than enough space to hold one of these annotations,

I just realized that I am restating what @ChugR said earlier. 

Although this PR is a good one, I agree that the context of utilization it won't bring any improvement.

franz1981:
@ganeshmurthy totally agree: I can close this one and will open another issue to discuss together a better solution to the origianl problem ie qd_buffer_list_clone has to be improved

PR #515, opened 2019-05-30
DISPATCH-1348 - Allocate qdr_error_t objects only when necessary. Pre…

ganeshmurthy:
…viously, these objects were always getting created even in the absence of an error

### DIFF
--- a/src/router_core/error.c
+++ b/src/router_core/error.c
@@ -37,35 +37,25 @@ qdr_error_t *qdr_error_from_pn(pn_condition_t *pn)
     qdr_error_t *error = 0;
 
     const char *name = pn_condition_get_name(pn);
-
-    if (name && *name) {
-        if (!error) {
-            error = new_qdr_error_t();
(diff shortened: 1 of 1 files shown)


src/router_core/error.c
```diff
@@ -34,19 +34,42 @@ qdr_error_t *qdr_error_from_pn(pn_condition_t *pn)
...
-    qdr_error_t *error = new_qdr_error_t();
-    ZERO(error);
+    qdr_error_t *error = 0;
 
     const char *name = pn_condition_get_name(pn);
-    if (name && *name)
+
+    if (name && *name) {
+        if (!error) {
+            error = new_qdr_error_t();
```
ted-ross: I think this would be easier to understand if you only called new_qdr_error_t() in one place, rather than having to repeatedly see if you already created it.  Why not extract the data from the pn_error first, determine if a qdr_error is required and if so, create and populate it?

PR #518, opened 2019-06-03
DISPATCH-1354: Annotation processing performance improvements

ChugR:
Message annotation processing on received messages stages key names
byte by byte into a flat buffer and then uses strcmp to check them.

Easy improvements are:

 * Use name in raw buffer if it does not cross a buffer boundary
 * If name crosses a boundary then use memmoves to get the name in chunks
 * Check the name prefix only once and then check variable parts of name strings
 * Don't create unnecessary qd_iterators and qd_parsed_fields
 * Don't check names whose lengths differ from the given keys

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -722,61 +722,123 @@ const char *qd_parse_annotations_v1(
         return parse_error;
     }
 
+    // define a shorthand name for the qd message annotation key prefix length
+#define QMPL QD_MA_PREFIX_LEN
+
+#define MIN(a,b) (((a)<(b))?(a):(b))
+
+    // trace, phase, and class keys are all the same length
+    assert(QD_MA_TRACE_LEN == QD_MA_PHASE_LEN);
+    assert(QD_MA_TRACE_LEN == QD_MA_CLASS_LEN);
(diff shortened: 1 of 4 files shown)


ChugR:
Informal quiver tests show about a 10% improvement.
* Linear three router network
* qpid-proton-c arrows
* Message routed transfers
* 1,000,000 message with 100 byte payload
* No user annotations
* All routers and quiver arrows run on single laptop

src/parse.c
```diff
@@ -722,61 +722,123 @@ const char *qd_parse_annotations_v1(
         return parse_error;
     }
 
+    // define a shorthand name for the qd message annotation key prefix length
+#define QMPL QD_MA_PREFIX_LEN
+
+#define MIN(a,b) (((a)<(b))?(a):(b))
```
kgiusti: ooh!  Please put this (and maybe MAX) into ctools.h - don't know how many times I've ended up doing this exact same thing.

ChugR:
Includes MIN, MAX macros in ctools.h

Closed at commit 95c8f4

PR #526, opened 2019-06-25
mercury initial checkin

mgoulish:
The Mercury testing system, written in Go, as a Possible Proposal for inclusion as a tool in the Dispatch Router.


tools/mercury/README.md
```diff
@@ -0,0 +1,520 @@
...
+1. the go language on your machine
+2. installed proton and dispatch router software
+
+
+
+<br/>
+
+###Mercury Audience
+
+I see Mercury as a tool for developers. The user has one or more installed versions of the Dispatch Router + Proton code and wantsto easily set up a complex network including Dispatch routers, edge routers, and clients with nontrivial addressing patterns. Currently I think that creation of nontrivial networks is difficult enough that it discourages extensive testing during development.
```
ganeshmurthy: Should it be "set up a complex network including interior routers"

ganeshmurthy: Can we remove "I see". Can it simply be "Mercury is a tool...."

Also remove "I think" etc.

tools/mercury/README.md
```diff
@@ -0,0 +1,520 @@
...
+2. Tell the client to form multiple links with multiple "--address" arguments.
+3. Tell it where to send its log files and so on.
+
+
+<br/>
+
+
+###Starting Mercury
+
+The directory from which to run Mercury is also called mercury.  So it's mercury/mercury.  And the run-script that I use is 'r'.
```
ganeshmurthy: run script is r ? Where is r ?

tools/mercury/README.md
mgoulish: Yes, it certainly should.  Fixed.

tools/mercury/README.md
mgoulish: I changed this to say "example_run_script" instead of "r", and then checked that in to the code tree.  And that run script calls "example_script" which is also checked in.

tools/mercury/README.md
mgoulish: Well ... I was planning to make all those occurrences of the word "I" into hyperlinks that would show my picture, contact information, LinkedIn page, a small biography, and an audio file of me singing _None Shall Sleep_ from the final act of the opera _Turandot_.   

But, OK.   I took them out.

tools/mercury/README.md
```diff
@@ -0,0 +1,508 @@
...
+
+1. throttle send-speed with the "--throttle" argument
+2. Tell the client to form multiple links with multiple "--address" arguments.
+3. Tell it where to send its log files and so on.
+
+
+<br/>
+
+
+###Starting Mercury
```
ganeshmurthy: I read the "Starting Mercury" section and was eager to start mercury. I opened the mercury/example_run_script, changed the   MERCURY_ROOT, then launched it. I see - 

☿ error: rn.check_path: Path |dispatch_root| does not exist at |/home/mick/latest/install/dispatch|
Should this "Starting Mercury" section address how this path needs to set *before* running example_run_script ?

tools/mercury/README.md
```diff
@@ -0,0 +1,508 @@
...
+You can define a version in one of two different ways
+
+1. You can provide the root directories for the proton and dispatch installation, and let Mercury calculate from them all the paths it needs, or
+
+2. You can directly provide all the paths. This second option is meant for situations where your installation is different somehow from what Mercury expects.
+
+To define a version with roots, use the 'version\_roots' command something like this:  (Here we define two different versions.)
+
+
+    version_roots name latest dispatch /home/mick/latest/install/dispatch proton /home/mick/latest/install/proton
```
ganeshmurthy: Remove all "mick" from folder names and anywhere else from the code.

tools/mercury/README.md
```diff
@@ -0,0 +1,508 @@
...
+    version_roots name later  dispatch /home/mick/later/install/dispatch  proton /home/mick/later/install/proton
```
ganeshmurthy: Remove dispatch and proton and just assume that the first is dispatch path and second is proton path

tools/mercury/README.md
```diff
@@ -0,0 +1,508 @@
...
+3. Tell it where to send its log files and so on.
+
+
+<br/>
+
+
+###Starting Mercury
+
+The directory from which to run Mercury is also called mercury.  So it's mercury/mercury.  And an example of a simple run-script is 'example_run_script'.
+In that script you will see that it sets an environment variable MERCURY\_ROOT. On your system, please change this to something appropriate to your installation.
```
ganeshmurthy: Look in the default installation places for Dispatch and Proton. The users can override this by providing their own DISPATCH_INSTALL_DIR and PROTON_INSTALL_DIR

mgoulish: OK, I have made it look in the standard places if no other versions are defined. 
( This does *not* complete the changes I need to make, but it's a big one. )

No, the idea about overriding with environments variables isn't quite right.  An important part of the idea with Mercury is that it makes it easy to define multiple versions of dispatch & proton, and then use them in a heterogeneous network.   An environment variable like DISPATCH_INSTALL_DIR would imply that there is only one.     The different versions of dispatch+proton  are defined by the "versions" command.

tools/mercury/README.html
```diff
@@ -0,0 +1,493 @@
+<h1>Mercury</h1>
```
jiridanek, line 1: This file is generated from README.md? In that case I'd say don't check it in.

tools/mercury/src/utils/utils.go
```diff
@@ -0,0 +1,306 @@
...
+  return freqs
+}
+
+
+
+
+
+func Set_Top_Freq ( freqs [] string ) ( err error ) {
+  command := "sudo"
+  args    := "cpupower frequency-set --freq " + freqs [ 0 ] + "GHz"
```
jiridanek, line 296: This does not pass `go fmt`, /me thinks. The same is true for pretty much all the .go files in the PR.

jiridanek (commented):
I am currently adding https://github.com/apache/qpid-dispatch/pull/1303 and this PR made me realize that there are now at least 3 test setup tools for Dispatch in existence. First, the system_tests router configuration approach in Python, which starts the router network as subprocesses and watches for error status codes on exit. Second, mercury here, assuming it is still alive and in use, somewhere. Finally, what I have in the PR, which is basic string concatenation specific to each test to put together some minimal configs with configurable port numbers.

@mgoulish I am wondering what's a good way forward.

To be honest, I am not thrilled by the use of Go. Since Python 3.6 introduced async, that for me kills the main advantage of Go, which is the ability to structure tests with goroutines and channels. Async is not perfect (https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) but it is actually kind of nice, syntactically, and it does a lot of things goroutines do, except differently. If anything, I'd like to see rust here ;P But mostly I'd stick with Python, just because it's already around.

C++ is poor in that it cannot do string templating very well, on the other hand it interoperates with C almost perfectly, so it could be used for writing those unittests with mocks, and all the other nice things that reduce the scope of a test and make it fast (avoiding unneeded setup of parts of the program that are not under test currently). That's why I think C++ is a good language for writing tests.

I feel that Mercury does not bring many advantages over Python test infrastructure, but if that is something that is already developed, and that you still actively use, and which has the useful features needed for testing, it would IMO make sense to have it be part of Dispatch codebase.

PR #528, opened 2019-06-26
DISPATCH-1375 qdr_general_work_t could be shrinked to increase cache localty

franz1981:


franz1981:
@kgiusti @ganeshmurthy @astitcher It seems that only C11 support anonymous struct...I don't know if we want to bring in C11 support just to allow this..wdyt?

kgiusti:
> @kgiusti @ganeshmurthy @astitcher It seems that only C11 support anonymous struct...I don't know if we want to bring in C11 support just to allow this..wdyt?

At this point we still support older distributions like Centos6 whose default C compiler does not yet support C11 (IIRC). 

It would be better to discuss C version compatibility on the Qpid mailing list users@qpid.apache.org

astitcher:
Proton actually supports C99ish rather than actual C99!
Initially we chose C99, but then had to back off because the then current Microsoft compilers didn't support the C99 features we used.
So we switched to compiling with C++ on Visual Studio. So we actually use the common subset of C99 and C++03.
Visual Studio C compilers from VS2015 actually support enough C99 for the earlier Proton. However I would steer clear of C11 for another decade if you want to compile on VS!

astitcher:
Having said that it would be trivial (if irritating) to just add in a struct name. As a non-standard extension some C99 compilers support this anyway (I'm pretty sure that gcc and clang do using -std=gnu99)

franz1981:
@astitcher I agree, I will just need to create a clean API to allow the caller to use the work item correctly (like action already does)

PR #531, opened 2019-07-01
DISPATCH-1380: clear pn_link context when releasing qd_link

kgiusti:


ganeshmurthy:
Looks good. The commit needs to have the JIRA number on it. That's it.

kgiusti:
I've updated the git commit message with DISPATCH-1380.

Turns out the extra two calls to pn_link_set_context() causes a leak of qd_link_t's, so reverting those.

PR #532, opened 2019-07-03
DISPATCH-1381: Create procedure for handling undeliverable messages

bhardesty:


bhardesty:
@ted-ross this is the doc for the fallback destination feature (DISPATCH-1337). Please review for technical accuracy.

docs/books/user-guide/routing.adoc
```diff
@@ -551,6 +551,88 @@ autoLink {
...
++
+For more information, see xref:connecting-to-external-amqp-containers-router[].
+
+.Procedure
+
+This procedure enables fallback for an address
+and configures autolinks to connect to the broker queue
+that provides the fallback destination for the address.
+
+. Enable fallback destinations for the address.
```
jenmalloy: Change the period to a colon? (I realize this is a stylistic element that might conflict with the rest of the docs)

docs/books/user-guide/routing.adoc
```diff
@@ -551,6 +551,87 @@ autoLink {
...
 
+=== Handling Undeliverable Messages for an Address
+
+You handle undeliverable messages for an address by configuring _fallback destinations_.
+A fallback destination is a queue on a broker that stores messages that are not directly routable to any consumers.
+
+During normal message delivery,
+{RouterName} delivers messages to the consumers that are attached to the router network.
+However, if no consumers are reachable,
+the messages are diverted to any fallback destinations that were configured for the address.
```
ted-ross: ... any fallback destinations that are attached for the address...
It's possible that the auto-links are inactive, in which case messages aren't delivered anywhere.  Also, it is possible for a fallback destination to attach by itself (using the qd.fallback capability) without using an auto-link.

docs/books/user-guide/routing.adoc
```diff
@@ -551,6 +551,87 @@ autoLink {
...
+When a consumer reconnects and becomes reachable again,
+it receives the messages stored at the fallback destination.
+
+[NOTE]
+====
+{RouterName} does not preserve delivery order for messages stored at a fallback destination. When a consumer reconnects, it will receive the stored messages interleaved with any new messages sent by producers.
```
ted-ross: To be more accurate: message delivery is preserved in the queue and the queue will deliver those messages in the same order in which they were received.  However, new messages produced while the queue is draining will be interleaved with the draining messages.

docs/books/user-guide/routing.adoc
```diff
@@ -551,6 +551,87 @@ autoLink {
 For information about additional attributes, see link:{qdrouterdConfManPageUrl}#_autolink[autoLink] in the `qdrouterd.conf` man page.
 --
 
+=== Handling Undeliverable Messages for an Address
+
+You handle undeliverable messages for an address by configuring _fallback destinations_.
+A fallback destination is a queue on a broker that stores messages that are not directly routable to any consumers.
```
ted-ross: This is probably the most common use for fallback-destination.  However, there is no reason that the fallback destination must be a queue on a broker.

docs/books/user-guide/routing.adoc
```diff
@@ -551,6 +551,87 @@ autoLink {
...
+--
+
+. If you want the router to send queued messages to attached consumers as soon as they connect to the router network,
+add an _incoming_ autolink.
++
+--
+As soon as a consumer attaches to the router,
+it will receive the messages stored in the broker queue,
+along with any new messages sent by the producer.
+However, the original delivery order of the queued messages is not preserved; the queued messages will be interleaved with the new messages.
```
ted-ross: The original delivery order of the queued messages _is_ preserved.  New messages produced during queue-draining are not run through the queue, they are sent directly to the consumers, interleaved with the draining messages.

docs/books/user-guide/routing.adoc
```diff
@@ -551,6 +551,87 @@ autoLink {
...
+
+If you do not add the incoming autolink,
+the messages will be stored on the broker,
+but will not be automatically sent to consumers when they attach to the router.
```
ted-ross: I would delete "automatically" as it's not adding any valuable meaning.

PR #533, opened 2019-07-03
DISPATCH-1382: Create procedure for closing connection from console

bhardesty:


docs/books/user-guide/using-console.adoc
```diff
@@ -120,3 +120,17 @@ In the web console, you use the tabs to monitor the router network.
 | `Schema` | The management schema that controls each of the routers in the router network.
 
 |===
+
+== Closing a connection
+
+If a consumer is processing messages too slowly, or has stopped processing messages without settling its deliveries, you can close the connection. When you close the connection, the "stuck" deliveries are released.
```
jbyrne-redhat: This might be just me, but is it worth a few more words to say what 'released' means? Released to another available consumer?

PR #537, opened 2019-07-16
DISPATCH-1386 qdr_action_list_t could be replaced by an array of struct q

franz1981:
It could be improved by adding a configurable batch size, more tests and saving completly to allocate qdr_action_t

### DIFF
--- a/src/router_core/router_core.c
+++ b/src/router_core/router_core.c
@@ -40,6 +40,92 @@ ALLOC_DEFINE(qdr_connection_info_t);
 
 static void qdr_general_handler(void *context);
 
+uint64_t next_power_of_two(uint64_t v)
+{
+    assert(v > 0);
+    v--;
+    v |= v >> 1;
+    v |= v >> 2;
+    v |= v >> 4;
+    v |= v >> 8;
+    v |= v >> 16;
+    v++;
+    return v;
+}
+
+inline void qdr_action_q_init(qdr_action_q_t *actions, uint64_t capacity)
+{
+    capacity = next_power_of_two(capacity);
+    actions->c_seq = 0;
+    actions->p_seq = 0;
+    actions->capacity = capacity;
+    ALLOC_CACHE_ALIGNED(capacity * sizeof(qdr_action_t), actions->action);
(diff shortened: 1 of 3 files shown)


franz1981:
I need to take care of freeing up the action q at the end of the core thread lifecycle too

franz1981:
On my tests with 9 workers I've gotten an improvement from 64 K msg/sec -> 73 K msg/sec

franz1981:
I've just noticed that a similar improvement could be done on `qdr_general_work_list_t` too, probably with the same benfits. I think it would be better to abstract the work done on the action list into a proper separate data structure, to allow reuse it AMAP

src/router_core/router_core.c
```diff
@@ -101,7 +188,11 @@ void qdr_core_free(qdr_core_t *core)
     // Stop and join the thread
     //
     core->running = false;
-    sys_cond_signal(core->action_cond);
+    sys_mutex_lock(core->action_lock);
+    if (core->sleeping) {
```
franz1981: @kgiusti I forgot to mention that this PR contains an important optimization under load to reduce the signaling work of the worker threads ie to save signaling core if it isn't sleeping. 
This is beneficial from many point of views although it won't show up on benchmarks: please let me know if I should move it into a separate PR/jira/commit :+1:

franz1981:
@kgiusti @ted-ross I see that's sitting here from a long time, but in many tests I've performed last year I see this change to be quite beneficial.

Let me summarize the changes:

1. replaced the action list of pointers with a queue of structs (that can grow unbounded)
2. using a wait flag to saving some signaling 

It could be improved by using a proper lock-free multi-producer, single consumer queue (of structs) or (better) a mesh of single-producer, single-consumer queues: some tests from @mgoulish last year on the former approach has shown a significant improvement in both latency and throughput by using such approach.

This PR is a sort of "half" solution that won't introduce any lock-free queue. but is a bit more mechanical friendly with the dispatch allocator.

Let me know your decisision so I can either close or rebase by solving the conflicts for this one :+1:

PR #539, opened 2019-07-18
NO-JIRA - put items on the allocator free list stack to increase localty

franz1981:
I need yet to perform some benchmark


franz1981:
I can close this, because is not that important to be fixed ATM and with this :+1:

PR #541, opened 2019-07-22
DISPATCH-1389: optimize the STATE_IN_BODY codepath in qd_iterator_octet

kgiusti:
Minor change increases field parsing performance in general

Single hop (best of 3 runs)
master: 
    3.177: msgs sent=500000 msgs/sec=  157377.115
patch:
    2.470: msgs sent=500000 msgs/sec=  202455.051

Two hop:
master:
  8.170: msgs sent=500000 msgs/sec=   61198.177

patch:
  6.814: msgs sent=500000 msgs/sec=   73377.456

### DIFF
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -558,6 +558,32 @@ unsigned char qd_iterator_octet(qd_iterator_t *iter)
     if (!iter)
         return 0;
 
+    if (iter->state == STATE_IN_BODY) {
+        if (iter->view_pointer.remaining == 0)
+            return (unsigned char) 0;
+
+        unsigned char result = *(iter->view_pointer.cursor);
+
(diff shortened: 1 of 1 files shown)


src/iterator.c
```diff
@@ -555,8 +555,33 @@ void qd_iterator_annotate_space(qd_iterator_t *iter, const char* space, int spac
 
 unsigned char qd_iterator_octet(qd_iterator_t *iter)
 {
-    if (!iter)
-        return 0;
+    assert(iter);
```
ganeshmurthy: It is good to add this assert but can we still keep this line ?
    if (!iter)
        return 0;

It saves the router from a crash in production environment

kgiusti: yeah, was hoping to save a bit there, but look like the rest of the api verifies the iter pointer, so I'll add it back

PR #544, opened 2019-07-22
DISPATCH-1389: export qd_buffer_t access methods inline

kgiusti:
The qd_buffer_t structure is defined in the header file as a public
structure.  There are several trivial methods that are called
frequently enough that there is a benefit to performance by inlining
them.


ChugR:
+1 Ship it

PR #545, opened 2019-07-24
DISPATCH-1391 - Zero out pn_link references on all qd_links related t…

ganeshmurthy:
…o the session


src/container.c
```diff
@@ -541,14 +541,16 @@ void qd_container_handle_event(qd_container_t *container, pn_event_t *event,
                     qd_conn->n_sessions--;
                 }
 
-                pn_link = pn_link_head(conn, PN_LOCAL_ACTIVE | PN_REMOTE_CLOSED);
+                //Sweep thru every pn_link in this connection and a matching session and zero out the
+                // qd_link->pn_link reference. We do this in order to not miss any pn_links
+                pn_link = pn_link_head(conn, 0);
```
kgiusti: Suggestion:

Since you now have to check every link on the session why not combine this loop with the previous into a single loop?
Loop through all once using an if statement to check for PN_LOCAL_ACTIVE | PN_REMOTE_ACTIVE and do the detach handling  zeroing the pn_link entry.

PR #548, opened 2019-08-12
DISPATCH-1396: Add starting router procedure

bhardesty:


docs/books/_common/fragment-start-router-service-command.adoc
```diff
@@ -0,0 +1,23 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License
+////
+
+[source,bash,options="nowrap"]
+----
+$ systemctl start qdrouterd.service
```
jdanekrh: This requires `systemd`, meaning rhel 7+. Not sure if this is problem or not.

bhardesty: I generalized a bit here since the major distros (Fedora, openSUSE, Debian, and most Debian variants) boot systemd by default.

PR #553, opened 2019-08-21
NO-JIRA: update Travis with latest Proton (0.9.0)

kgiusti:
I also took the liberty to bump up the test timeout when running in travis - occasionally hitting the default timeout.


kgiusti:
I'll fix the bad version # in the commit log before merging!

PR #555, opened 2019-08-21
DISPATCH-1401: fix coverity locking false positive.

kgiusti:
Adds assert tests to enforce the locking convention.


kgiusti:
outdated

PR #557, opened 2019-08-28
Dispatch 1394 1404

kgiusti:
Pushing both fixes as they are dependent

### DIFF
--- a/include/qpid/dispatch/message.h
+++ b/include/qpid/dispatch/message.h
@@ -232,8 +232,17 @@ void qd_message_send(qd_message_t *msg, qd_link_t *link, bool strip_outbound_ann
 /**
  * Check that the message is well-formed up to a certain depth.  Any part of the message that is
  * beyond the specified depth is not checked for validity.
+ *
+ * Note: some message sections are optional - QD_MESSAGE_OK is returned if the
+ * optional section is not present, as that is valid.
  */
-int qd_message_check(qd_message_t *msg, qd_message_depth_t depth);
+typedef enum {
+    QD_MESSAGE_DEPTH_INVALID,     // corrupt or malformed message detected
+    QD_MESSAGE_DEPTH_OK,          // valid up to depth, including 'depth' if not optional
+    QD_MESSAGE_DEPTH_INCOMPLETE   // have not received up to 'depth', or partial depth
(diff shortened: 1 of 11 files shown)


include/qpid/dispatch/message.h
```diff
@@ -232,8 +232,17 @@ void qd_message_send(qd_message_t *msg, qd_link_t *link, bool strip_outbound_ann
...
+ *
+ * Note: some message sections are optional - QD_MESSAGE_OK is returned if the
+ * optional section is not present, as that is valid.
  */
-int qd_message_check(qd_message_t *msg, qd_message_depth_t depth);
+typedef enum {
+    QD_MESSAGE_INVALID,     // corrupt or malformed message
+    QD_MESSAGE_OK,          // valid up to depth, including 'depth' if not optional
+    QD_MESSAGE_INCOMPLETE   // have not received up to 'depth', or partial depth
+} qd_message_check_t;
```
ganeshmurthy: Should we call this qd_message_status_t and have the components as 

typedef enum {
    QD_MESSAGE_STATUS_INVALID,     // corrupt or malformed message
    QD_MESSAGE_STATUS_OK,          // valid up to depth, including 'depth' if not optional
    QD_MESSAGE_STATUS_INCOMPLETE   // have not received up to 'depth', or partial depth
} qd_message_check_t;

For example, I looked at this one - https://github.com/apache/qpid-dispatch/blob/master/src/router_core/router_core_private.h#L74

kgiusti: Since the API is checking a message up to a given depth, why not use "depth" in the name?
Something like:

`typedef enum {`
`QD_MESSAGE_DEPTH_INVALID, // corrupt or malformed message`
`QD_MESSAGE_DEPTH_OK, // valid up to depth, including 'depth' if not optional`
`QD_MESSAGE_DEPTH_INCOMPLETE // have not received up to 'depth', or partial depth`
`} qd_message_depth_status_t;`
` `
`qd_message_depth_status_t qd_message_check_depth(qd_message_t *msg, qd_message_depth_t depth);`

ganeshmurthy: Sounds good to me. Let's do this and ship it.

PR #567, opened 2019-09-19
DISPATCH-1428

grs:
Allow a route-container connection to be looked up by connector name, even if the container id is the same as a connection from a different connector.

### DIFF
--- a/src/router_core/route_control.c
+++ b/src/router_core/route_control.c
@@ -549,22 +549,8 @@ void qdr_route_del_auto_link_CT(qdr_core_t *core, qdr_auto_link_t *al)
     qdr_core_delete_auto_link(core, al);
 }
 
-
-void qdr_route_connection_opened_CT(qdr_core_t       *core,
-                                    qdr_connection_t *conn,
-                                    qdr_field_t      *container_field,
-                                    qdr_field_t      *connection_field)
+static void activate_route_connection(qdr_core_t *core, qdr_connection_t *conn, qdr_conn_identifier_t *cid)
 {
-    if (conn->role != QDR_ROLE_ROUTE_CONTAINER)
-        return;
-
-    qdr_conn_identifier_t *cid = qdr_route_declare_id_CT(core,
-            container_field?container_field->iterator:0, connection_field?connection_field->iterator:0);
-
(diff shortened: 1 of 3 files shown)


src/router_core/route_control.c
```diff
@@ -625,9 +595,65 @@ void qdr_route_connection_closed_CT(qdr_core_t *core, qdr_connection_t *conn)
...
+        return;
+
+    if (connection_field) {
+        qdr_conn_identifier_t *cid = qdr_route_declare_id_CT(core, 0, connection_field->iterator);
+        qdr_add_connection_ref(&cid->connection_refs, conn);
+        conn->conn_id = cid;
+        activate_route_connection(core, conn, conn->conn_id);
+        if (container_field) {
+            cid = qdr_route_declare_id_CT(core, container_field->iterator, 0);
+            qdr_add_connection_ref(&cid->connection_refs, conn);
```
kgiusti: Q: after adding the second connection_ref (616), if the conn_id and the cid are equal the new cid is not saved in the conn's alt_conn_id.

Does this leak a cid?

grs: It would result in an extra connection ref being added for the same connection, which would probably not be removed when the connection goes away as it would only remove the first ref, so the cid would indeed then not be eligible for removal at that point.

I'll move the qdr_add_connection_ref after the cid comparison.

tests/system_tests_link_routes.py
```diff
@@ -2052,6 +2052,132 @@ def on_message(self, event):
...
+        try:
+            p.teardown()
+        except Exception as e:
+            raise Exception("%s\n%s" % (e, out))
+        return out
+
+    def test_both_link_routes_active(self):
+        cmds = [
+            'CREATE --type=linkRoute name=foo prefix=foo direction=in connection=one',
+            'CREATE --type=linkRoute name=bar prefix=bar direction=in connection=two'
```
ganeshmurthy: can we please add a third CREATE linkRoute with a containerId and test if link route works ? Like so,
CREATE --type=linkRoute name=test prefix=test direction=in containerId=SendReceiveTestClient
Set the container id to abcd

grs: Done!

PR #571, opened 2019-09-20
DISPATCH-1422: Create global policy denial count

ChugR:
Policy status creates two global counts that get aggregated for
all links and connections:

* linksDenied - sum of all vhost sender and receiver denials
* totalDenials - sum of linksDenied, connectionsDenied, and
                 sessionsDenied

Implementation adds locks around changing shared static and
generating statistic reports for management.

Self test is added on to the PolicyVhostOverride as that test puts
its statistics into a known vhost. Otherwise stats could go into
system-specific vhosts like '0.0.0.0', '::0', 'localhost', etc.
adding unneeded complexity to the analysis.


ChugR:
Fixed at commit 69868bf

PR #573, opened 2019-09-24
DISPATCH-1426 - Added an abandoned deliveries handler so all abandone…

ganeshmurthy:
…d deliveries are handled in a single place, right before the qd_link is freed.

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -102,8 +102,8 @@ static qdr_delivery_t *qdr_node_delivery_qdr_from_pn(pn_delivery_t *dlv)
 }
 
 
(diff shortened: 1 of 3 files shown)


src/router_node.c
```diff
@@ -102,8 +102,8 @@ static qdr_delivery_t *qdr_node_delivery_qdr_from_pn(pn_delivery_t *dlv)
 }
 
 
-static void qdr_node_reap_abandoned_deliveries(qdr_core_t *core, qd_link_t *link)
-{
+void qd_link_abandoned_deliveries_handler(void *context, qd_link_t *link) {
```
kgiusti: Nit:  Add a newline after the parameter list: the opening brace should be on the next line by itself.

PR #578, opened 2019-10-01
DISPATCH-1434 - Added new attribute saslPasswordFile to the connector…

ganeshmurthy:
… entity. saslPassword entity has been deprecated.

### DIFF
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -1002,7 +1002,7 @@
                 "saslPassword": {
                     "type": "string",
                     "required": false,
-                    "description": "(DEPRECATED) The password that the connector is using to connect to a peer.",
+                    "description": "(DEPRECATED) The password that the connector is using to connect to a peer. This takes precedence over the saslPasswordFile if both are specified. This attribute has been deprecated because it is unsafe to store plain text passwords in config files. Use the saslPasswordFile instead",
                     "create": true,
                     "deprecated": true,
                     "hidden": true
@@ -1010,7 +1010,7 @@
                 "saslPasswordFile": {
                     "type": "string",
                     "required": false,
-                    "description": "The absolute path to the file that contains the password that the connector uses to connect to a peer.",
+                    "description": "The absolute path to the file that contains the password that the connector uses to connect to a peer. This file should be permission protected to limit access",
                     "create": true
                 },
                 "messageLoggingComponents": {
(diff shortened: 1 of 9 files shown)


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1002,10 +1002,17 @@
...
-                    "description": "The password that the connector is using to connect to a peer.",
+                    "description": "(DEPRECATED) The password that the connector is using to connect to a peer.",
                     "create": true,
+                    "deprecated": true,
                     "hidden": true
                 },
+                "saslPasswordFile": {
+                    "type": "string",
+                    "required": false,
+                    "description": "The full path to the file that contains the password that the connector uses to connect to a peer.",
```
bhardesty: Instead of "full path", use "absolute path".

bhardesty:
The now deprecated "saslPassword" is used in the doc in two places. They should be updated:
 
https://qpid.apache.org/releases/qpid-dispatch-1.9.0/user-guide/index.html#connecting-using-username-password-authentication-router

https://qpid.apache.org/releases/qpid-dispatch-1.9.0/user-guide/index.html#managing_sasl_encryption_and_authentication

src/connection_manager.c
```diff
@@ -327,6 +329,41 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
...
+                    buffer[i] = '\0';
+                    free(config->sasl_password);
+                    config->sasl_password = strdup(buffer);
+                }
+                fclose(file);
+            }
+        }
+        free(password_file);
+    }
+
```
kgiusti: This chunk of code appears to be a cut'n'paste of the ssl password file parsing.
Would it be possible to eliminate the redundancy an instead create a local (static) function that does this then have both the sasl and ssl codepaths call it?

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1002,10 +1002,17 @@
                 "saslPassword": {
                     "type": "string",
                     "required": false,
-                    "description": "The password that the connector is using to connect to a peer.",
+                    "description": "(DEPRECATED) The password that the connector is using to connect to a peer.",
```
kgiusti: Compare this to the text description for SSL's password.  SSL Password's description contains a bit more detail - can you update saslPassword's description to be consistent?

> This attribute has been deprecated because it is unsafe to store plain text passwords in config files. Use the passwordFile instead

src/connection_manager.c
ganeshmurthy: Done

python/qpid_dispatch/management/qdrouter.json
ganeshmurthy: Done

PR #581, opened 2019-10-04
DISPATCH-1442 - Added annotation field to the router entity.

ted-ross:


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -493,6 +493,12 @@
                     "create": true,
                     "required": false,
                     "default": "balanced"
+                },
+                "annotation": {
+                    "type": "string",
+                    "description": "Annotation text for this router, to be used by user or orchestration layers.",
```
bhardesty: It doesn't seem clear from the description why someone would add an annotation to a router. Is it just a free-form description of the router?

ted-ross: Suggestions are welcome...
How about "Optional annotation text for this router.  This may be used by the console to control the appearance of the router or it may be used by controlling orchestration software to annotate the router with useful information."
The field is optional and non-functional (it doesn't affect the operation of the router).  It's usefulness is completely dictated by the management/control layers over the router.

bhardesty: How about this (I made some assumptions about the primary use case for providing an annotation, which may or may not be accurate):

Optional annotation text to describe or label this router. This text does not affect the function of the router, but it can be used by external tools (such as the console or orchestration software) to control how the router is displayed.

ted-ross: That works for me.  I'll make the change

PR #587, opened 2019-10-15
DISPATCH-1445 - Modeled saslPassword to be more or less in line with …

ganeshmurthy:
…sslProfile's password field


tests/system_tests_sasl_plain.py
```diff
@@ -22,15 +22,31 @@
...
+import os, json
+from subprocess import PIPE, STDOUT, Popen
+from system_test import TestCase, Qdrouterd, main_module, DIR, TIMEOUT, SkipIfNeeded, Process
 from system_test import unittest
 from qpid_dispatch.management.client import Node
 from proton import SASL
 
 class RouterTestPlainSaslCommon(TestCase):
 
+    def run_qdmanage(self, cmd, input=None, expect=Process.EXIT_OK,
```
kgiusti: We've cut and pasted this code too many times.
See system_test.py - class QdManager().   It wraps qdmanage and exports operations as methods.  See system_tests_edge_router.py for an example.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1008,7 +1008,7 @@
                 "saslPassword": {
                     "type": "string",
                     "required": false,
-                    "description": "The password that the connector is using to connect to a peer.",
+                    "description": "The password that the connector is using to connect to a peer. Supports three prefixes namely - env:, file:, pass:. env:var obtains the password from the environment variable var. Since the environment of other processes is visible on certain platforms (e.g. ps under certain Unix OSes) this option should be used with caution. file:absolutepath obtains the password from the  absolute path of the file containing the password. This option is the safest since permissions can be set on the file. pass:password or password with no prefix is used to directly specify the password and should only be used where security is not important",
```
bhardesty: It's never easy to describe multiple values in paragraph format, but I think this works a bit better:

The password that the connector is using to connect to a peer. You can specify the password by specifying an environment variable that stores the password, a file that stores the password, or by entering the password in clear text. To use an environment variable, specify "saslPassword: env:<environment-variable>". Use this option with caution, because the environment of other processes is visible on certain platforms (for example, "ps" on certain Unix OSs). To use a file, specify "saslPassword: file:<absolute-path-to-file>". This option is the most secure, because permissions can be set on the file that contains the password. To specify the password in clear text, specify "saslPassword: pass:<password>" or "saslPassword: <password>". This option is insecure, so it should only be used if security is not a concern.

bhardesty:
In ../docs/books/user-guide/modules/connecting-using-username-password-authentication.adoc, we show an example that uses the saslPassword attribute. It's technically still correct, but it uses the non-recommended clear text option. It might be worth changing it to use the file option, since that seems most conducive to production scenarios.

PR #590, opened 2019-10-17
DISPATCH-1384 - Fix unit_tests

jdanekrh:
The inlining is not happening on Linux anyways, so disabling it should affect only macOS. I could not think of other way to make tests pass with the inlining, except doing some really crazy things that would have to affect what happens during normal execution too.

I have a StackOverflow question about it, btw ;)
https://stackoverflow.com/questions/56711149/how-do-i-replace-a-function-by-a-mock-implementation-when-unit-testing-a-shared

`-Wl,-export-dynamic` is Clangs equivalent of `-rdynamic` in GCC. I thought it should be used, but it does not appear to be necessary, so I am not adding that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,9 +191,13 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wno-gnu-statement-expression)
 endif()
 
-if (NOT (APPLE OR USE_SANITIZERS OR USE_TSAN))
+if (APPLE)
+  set(CATCH_UNDEFINED "-Wl,-flat_namespace")
+elseif(NOT (USE_SANITIZERS OR USE_TSAN))
   set(CATCH_UNDEFINED "-Wl,-z,defs")
-endif ()
+else(APPLE)
+  set(CATCH_UNDEFINED "")
+endif(APPLE)
 
 ##
 ## Header file installation
--- a/src/server.c
+++ b/src/server.c
@@ -1524,6 +1524,7 @@ bool qd_connector_decref(qd_connector_t* ct)
     return false;
 }
 
+__attribute__((noinline)) // permit replacement by dummy implementation in unit_tests
 void qd_server_timeout(qd_server_t *server, qd_duration_t duration) {
     pn_proactor_set_timeout(server->proactor, duration);
 }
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -191,9 +191,13 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wno-gnu-statement-expression)
 endif()
 
-if (NOT (APPLE OR USE_SANITIZERS OR USE_TSAN))
+if (APPLE)
```
kgiusti: Just a head's up - I'm refactoring the way we enable TSAN in cmake to follow the model used by proton. 
See [DISPATCH-1450](https://issues.apache.org/jira/browse/DISPATCH-1450)

I'll be creating a pull request for DISPATCH-1450 in the next day or so.

jdanekrh: Thanks for the head's up. Would you be OK if this gets merged before you get done? For APPLE, the important thing is never to set `"-Wl,-z,defs"`, because then the compile fails. No other considerations I know about, so far.

Btw, there is apparently special automagical setting to enable sanitizers with XCODE

```
cmake_minimum_required(VERSION 3.13)
set(CMAKE_XCODE_GENERATE_SCHEME ON)
set(CMAKE_XCODE_SCHEME_ADDRESS_SANITIZER ON)
set(CMAKE_XCODE_SCHEME_ADDRESS_SANITIZER_USE_AFTER_RETURN ON)
```

https://stackoverflow.com/questions/44320465/whats-the-proper-way-to-enable-addresssanitizer-in-cmake-that-works-in-xcode

Probably something that can be safely ignored for now.

I was wondering if "modern" CMake has special support for Valgrind, and so on, but I found nothing. It has e.g. support for cppcheck, though, https://cmake.org/cmake/help/v3.10/prop_tgt/LANG_CPPCHECK.html

kgiusti: yep, np.  I'll hold off until this patch lands.

jdanekrh: I'll consider comment https://github.com/jdanekrh/qpid-dispatch/commit/4c07a4d9f5e0bf6c12a3d5661f86adc6390b432a#r35553494 as sufficient review and merge this.

PR #591, opened 2019-10-17
DISPATCH-1384 - Fix assert in system_tests_bad_configuration

jdanekrh:
The expected Linux error is

    2019-10-17 10:14:09.319632 +0200 SERVER (info) [C1] Connection to unresolvable.host.name:amqp failed: proton:io Name or service not known - connect to  unresolvable.host.name:5672 (../src/server.c:974)

The expected macOS error is

    2019-06-21 21:00:34.125714 -0700 SERVER (info) [C4] Connection to unresolvable.host.name:amqp failed: proton:io unknown node or service - on connect resolving unresolvable.host.name:5672 (/Users/cloud-user/projects/qpid-dispatch/src/server.c:965)

The macOS error did not match the Linux specific regex used before.

### DIFF
--- a/tests/system_tests_bad_configuration.py
+++ b/tests/system_tests_bad_configuration.py
@@ -34,6 +34,7 @@ import re
 from subprocess import PIPE, STDOUT
 from system_test import TestCase, Qdrouterd, TIMEOUT, Process
 
+
 class RouterTestBadConfiguration(TestCase):
 
     """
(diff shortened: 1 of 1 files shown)


jdanekrh:
"**79% tests passed**, 13 tests failed out of 61"

jdanekrh:
There should not be anything "controversial" about this, and I've checked that the test is capable of failing (always concern when touching existing tests), so I'll just merge this. Thanks for review.

PR #592, opened 2019-10-17
DISPATCH-1363 - Move RAT licence check from travis builds into a sub build

jdanekrh:
Now easier to do when the `matrix` key is in place in `.travis.yml`.

Doing this also fixes the duplicated top level `cache` key in `.travis.yml` we had.

It appears that `matrix` is outdated syntax, and `jobs` should be preferred, https://github.com/travis-ci/docs-travis-ci-com/issues/1500. The doc page https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix uses `matrix.include` in the text and `jobs.include` in the yaml samples. Task for another time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,23 @@ matrix:
   allow_failures:
     - os: osx
   include:
+  - name: "apache-rat:check"
+    os: linux
+    sudo: false
+    env: []
+    language: java
+    addons:
+      apt:
+        packages: []
+    cache:
+      directories:
+        - $HOME/.m2/repository
+      before_cache:
+        - rm -rf $HOME/.m2/repository/org/apache/qpid/*
+    install: []
+    before_script: []
+    script:
+    - mvn apache-rat:check
   # prepending /usr/bin to PATH to avoid mismatched python interpreters in /opt
   - os: linux
     env:
@@ -61,8 +78,6 @@ addons:
     - python2.7-dev
     - sasl2-bin
     - swig
-    # For apache RAT tool
-    - maven
     # documentation
     - asciidoc
     - asciidoctor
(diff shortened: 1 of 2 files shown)


jdanekrh:
@ganeshmurthy Is this what you've had in mind?

ganeshmurthy:
@jdanekrh can you please rebase this PR with master and simply add a file with no license so that the RAT check fails? This is what I would like to see. If the RAT check fails, there should be a red X mark next to the checkin. When I click on the X mark I should be taken to a page which shows all builds passing but only the RAT failing (this is already happening).

jdanekrh:
Here, https://travis-ci.org/jdanekrh/qpid-dispatch/builds/599323593

jdanekrh:
Ahd the PR here https://github.com/apache/qpid-dispatch/pull/593

ganeshmurthy:
@jdanekrh ok thanks, I will wait for the builds to complete and check back in an hour or so to see how things have worked out. Will keep you posted.

jdanekrh:
@ganeshmurthy is Travis usually this slow during US work hours?

ganeshmurthy:
> @ganeshmurthy is Travis usually this slow during US work hours?

@jdanekrh It has been slow the past couple of days but usually it is pretty good.

gemmellr:
The Travis rsources under /apache are shared for the entire ASF, so it depends what other projects have done, e.g if anyone is overly [ab]using the resources at a given time.

Though these are only very small jobs by comparison to many, it still helps not to use branches in the main repo simply for testing variants of changes. You can enable and run Travis on your own repo fork (which is served from a far larger resource pool, and so is typically very fast to start building) and push to / pull request against the main repo only when you think its ready.

PR #595, opened 2019-10-18
DISPATCH-1450: enable thread sanitizer (TSAN) run time checking

kgiusti:
Currently all issues found by running the CTest suite with TSAN are
suppressed.  JIRAs will be opened to track each issue at a later date.


cmake/RuntimeChecks.cmake
```diff
@@ -17,41 +17,97 @@
...
+elseif(RUNTIME_CHECK STREQUAL "helgrind")
+  assert_has_valgrind()
+  message(STATUS "Runtime race checker: valgrind helgrind")
+  set(QDROUTERD_RUNNER "${VALGRIND_EXECUTABLE} --tool=helgrind ${VALGRIND_COMMON_ARGS}")
+
+#elseif(RUNTIME_CHECK STREQUAL "asan")
+#  assert_has_sanitizers()
+#  message(STATUS "Runtime memory checker: gcc/clang memory sanitizers")
+#  set(SANITIZE_FLAGS "-g -fno-omit-frame-pointer -fsanitize=address,undefined")
+#  set(TEST_WRAP_PREFIX "${CMAKE_SOURCE_DIR}/tests/preload_asan.sh $<TARGET_FILE:qpid-proton-core>")
```
jdanekrh: Why is this commented out? Besides asan, there is still one more sanitizer to be considered in the future, msan, `-fsanitize=memory`.

README
```diff
@@ -100,10 +100,62 @@ This does the following:
...
+
+GCC/Clang Thread Sanitizer (TSAN)
+---------------------------------
+This option turns on extra run time threading verification.
+
+Applicable only to GCC versions >= 4.8 and Clang versions >= 4.1.
+
+To enable the thread sanitizer set the RUNTIME_CHECK build flag to "tsan":
+
+cmake .. -DRUNTIME_CHECK=tsan
```
jdanekrh: It is recommended to use `-g`, and `-O1` or `-O2` with sanitizers (besides `-fno-omit-frame-pointer`, etc), precisely for the reason of speeding up the check. One of the reasons sanitizers can be faster than valgrind.

I am not sure how to do this in CMake. Just throw in `-O1` with other sanitizer options, if current build type is Debug? Note in readme? Require Release (ReleaseWithxxxSymbols) as build type when sanitizers are to be enabled?

cmake/RuntimeChecks.cmake
kgiusti: Simply due to prioritizing delivery of thread sanitizing as soon as possible.
We have Valgrind for memory checking already so there is coverage there, but nothing for threading at this point.

Feel free to open a JIRA upstream in dispatch to track implementing address and thread sanitizing as a separate patch.

README
kgiusti: Right now the level of optimization (-Ox) is determined by the build type.  So, by default (RelWithDebInfo) this will supply -O2.

the only target that does not enable optimization is Debug, which is by design.  Running tsan on a Debug build is not a priority - we should be running tsan on optimized targets only.

PR #596, opened 2019-10-18
DISPATCH-1455: assertRaises and mutually_exclusive groups not working as expected in rhel6

nicob87:
just to open the discussion, this is one way to fix it.
- mutually_exclusive_groups, not working as expected in python2.6, so manually
adding the functionality only for python_version < 2.7.
- assertRaises also works different in python2.6, it expects more arguments.

### DIFF
--- a/python/qpid_dispatch_internal/tools/command.py
+++ b/python/qpid_dispatch_internal/tools/command.py
@@ -25,7 +25,8 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import print_function
 
-import sys, json, argparse, os
+import sys, argparse, os
+from itertools import combinations
 try:
     from collections.abc import Mapping, Sequence
 except ImportError:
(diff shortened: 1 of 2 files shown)


nicob87:
in rhel6 this is gonna fail because the error message is not exactly like this:
```                                                                                                   
if "error: argument --all-entities: not allowed with argument -c/--connections" in str(e):                             
```
probably it will bit more like:
```
"error: argument --all-entities: not allowed with argument -c"
```

kgiusti:
I would simply skip the test if python is 2.6.  Would that be easier?

import sys

Then in the test
if sys.version_info[0]  == 2 and sys.version_info[1] < 7:
    self.skipTest("requires python >= 2.7")

taa daa!!!

ganeshmurthy:
Instead of skipping the test, we could test for the Python version and check for the appropriate message.

ganeshmurthy:
I have opened a JIRA to address this - https://issues.apache.org/jira/browse/DISPATCH-1455

nicob87:
Good! ok, based on the feedback, probably I will open a new PR with a different approach, just skipping, and/or testing the expected different behaviour and then we can compare and see what solution we like more.
PR linked with the Jira.

ganeshmurthy:
> Good! ok, based on the feedback, probably I will open a new PR with a different approach, just skipping, and/or testing the expected different behaviour and then we can compare and see what solution we like more.
> PR linked with the Jira.

Excellent. Yes, please close this PR and open a new one with what you said above. Thanks Nicolas.

PR #599, opened 2019-10-25
DISPATCH-1461: correct the logging of long addresses on attach

kgiusti:


src/router_core/terminus.c
```diff
@@ -76,30 +76,49 @@ void qdr_terminus_free(qdr_terminus_t *term)
...
+//
+static inline int safe_snprintf(char *str, size_t size, const char *format, ...)
+{
+    va_list ap;
+    va_start(ap, format);
+    int rc = vsnprintf(str, size, format, ap);
+    va_end(ap);
+
+    if (size && rc >= size)
+        return strlen(str);  // return actual # of bytes written
```
ted-ross: Shouldn't you just return size here?  The strlen won't count the closing null if there is one.

kgiusti: Hmmm.... I suppose size - 1 would work (snprintf return count excludes the null byte).
I'll give it a try.

src/router_core/terminus.c
```diff
@@ -76,30 +76,49 @@ void qdr_terminus_free(qdr_terminus_t *term)
 }
 
 
+// DISPATCH-1461: snprintf() is evil - it returns >= size on overflow.  This
+// wrapper will never return >= size, even if truncated.  This makes it safe to
+// do pointer & length arithmetic without overflowing the destination buffer in
+// qdr_terminus_format()
+//
+static inline int safe_snprintf(char *str, size_t size, const char *format, ...)
```
ganeshmurthy: should this function be in some place like src/aprinf.h so it can be used by everybody ?

kgiusti: eventually, yes - there are many places were snprintf is used, we should review them.
Until then, this fixes the immediate need.

src/router_core/terminus.c
kgiusti: Ok, patch updated.

PR #600, opened 2019-10-25
DISPATCH-1463 - Add detection for stuck deliveries

ted-ross:
The current delayed-delivery counts are limited in that they only count deliveries that have been settled.  These counters do not count deliveries that are currently "stuck" and have not been settled.

This feature provides visibility into the number of stuck deliveries per link and per router.

A new counter "deliveriesStuck" shall be added to the link and router entities for per-link and aggregated-per-router respectively.

A delivery is considered "stuck" when it has been in the undelivered or unsettled lists for more than ten seconds and is still not settled.  Once a stuck delivery is settled or otherwise removed, it is no longer considered stuck.

When a link transitions from zero to one stuck delivery, an INFO log message shall be emitted indicating that at least one delivery is stuck on that link.

### DIFF
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -570,6 +570,11 @@
                     "graph": true,
                     "description": "The total number of settled deliveries that were held in the router for more than 10 seconds."
                 },
+                "deliveriesStuck": {
+                    "type": "integer",
+                    "graph": true,
+                    "description": "The current number of deliveries that are unsettled and have been held in the router for more than 10 seconds."
+                },
                 "deliveriesIngress": {
                     "type": "integer",
                     "description":"Number of deliveries that were sent to it by a sender that is directly attached to the router.",
@@ -1523,6 +1528,11 @@
                     "type": "integer",
                     "graph": true,
                     "description": "The total number of settled deliveries that were held in the router for more than 10 seconds."
+                },
+                "deliveriesStuck": {
+                    "type": "integer",
+                    "graph": true,
+                    "description": "The current number of deliveries that are unsettled and have been held in the router for more than 10 seconds."
                 },
                 "settleRate": {
                     "type": "integer",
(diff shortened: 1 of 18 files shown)


ted-ross:
No tests have been added yet.  I'll update the PR with tests when they're ready.

src/router_core/router_core_thread.c
```diff
@@ -123,6 +123,33 @@ void qdr_modules_finalize(qdr_core_t *core)
...
+ * Process up to one available background action.
+ * Return true iff an action was processed.
+ */
+static bool router_core_process_background_action_LH(qdr_core_t *core)
+{
+    qdr_action_t *action = DEQ_HEAD(core->action_list_background);
+
+    if (!!action) {
+        DEQ_REMOVE_HEAD(core->action_list_background);
+        if (action->label)
```
kgiusti: Can the call to qd_log be moved to the unlocked section?  qd_log() takes locks and IIRC does sync I/O.

ted-ross: Yes, thanks.  I'll change that.

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -570,6 +570,11 @@
                     "graph": true,
                     "description": "The total number of settled deliveries that were held in the router for more than 10 seconds."
                 },
+                "deliveriesStuck": {
+                    "type": "integer",
+                    "graph": true,
+                    "description": "The current number of deliveries that are unsettled and have been held in the router for more than 10 seconds."
+                },
```
ChugR: Would you consider a 'peakDeliveriesStuck' stat? 
Picture a network where some clients go down and zillions of deliveries get stuck. Then those clients come back and the stuck number goes to zero. This would provide some indication of an issue even though the current state is fine.

ted-ross: Do we need this?  INFO logs are issued when a link has stuck deliveries and the count is available through the prometheus-format scraper.
Since this is more of a gauge than a counter, having a peak value might be appropriate.  Would it also require a way to be reset?

ChugR: Live stats is fine. Adding more becomes a slippery slope towards too much.

PR #602, opened 2019-10-28
DISPATCH-1453 - Fix the code so the correct error message shows up wh…

ganeshmurthy:
…en a tx sender attaches in the absence of a coordinated link route

### DIFF
--- a/tests/system_tests_one_router.py
+++ b/tests/system_tests_one_router.py
@@ -24,7 +24,7 @@ from __future__ import print_function
 
 from proton import Condition, Message, Delivery, Url, symbol, Timeout
 from system_test import TestCase, Qdrouterd, main_module, TIMEOUT, DIR, Process
-from system_test import unittest
+from system_test import unittest, QdManager
 from proton.handlers import MessagingHandler, TransactionHandler
 from proton.reactor import Container, AtMostOnce, AtLeastOnce, DynamicNodeProperties, LinkOption, ApplicationEvent, EventInjector
 from proton.utils import BlockingConnection, SyncRequestResponse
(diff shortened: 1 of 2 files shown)


ganeshmurthy:
@gemmellr Please take a look and approve if this works and you are satisfied. Thanks.

gemmellr:
@ganeshmurthy this seems to do the trick, with links to the coordinator rejected with same specific error condition+description regardless whether "defaultDistribution : unavailable" is configured or not.

tests/system_tests_one_router.py
```diff
@@ -3160,6 +3159,86 @@ def on_message(self, event):
...
+        self.assertTrue(test.passed)
+
+
+class RejectCoordinatorGoodMessageTest(RejectCoordinatorTest):
+    def __init__(self, url, link_route_present=False):
+        super(RejectCoordinatorGoodMessageTest, self).__init__(url)
+        self.link_route_present = link_route_present
+        self.error_with_link_route = "No route to the destination node"
+
+    def on_link_error(self, event):
```
kgiusti: Duplicate def of on_link_error (see next def below)

ganeshmurthy: Removed duplicate on_link_error. thanks

ganeshmurthy:
@gemmellr please approve this PR when you get a chance.

gemmellr:
The non-test code didnt seem to change since my previous comment, so I dont think theres anything needing me to try it out again is there? That being the case, still fine with me.

ganeshmurthy:
> The non-test code didnt seem to change since my previous comment, so I dont think theres anything needing me to try it out again is there? That being the case, still fine with me.

That is indeed the case. No need to try it out again. I will squash and merge this PR to master. 
Thanks @gemmellr and @kgiusti

PR #605, opened 2019-10-30
DISPATCH-1409 - Exposed available-credit and time-of-zero-credit as l…

ted-ross:
…ink attributes.  Added logging for links that have been stuck with no credit for more than ten seconds.

### DIFF
--- a/src/router_core/transfer.c
+++ b/src/router_core/transfer.c
@@ -262,6 +262,7 @@ void qdr_link_flow(qdr_core_t *core, qdr_link_t *link, int credit, bool drain_mo
     action->args.connection.drain  = drain_mode;
 
     qdr_action_enqueue(core, action);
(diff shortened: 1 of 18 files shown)


ted-ross:
I've put this up for review and testing to see if it provides the expected visibility into link credit.  I'll add a test shortly.

src/router_core/transfer.c
```diff
@@ -262,6 +262,7 @@ void qdr_link_flow(qdr_core_t *core, qdr_link_t *link, int credit, bool drain_mo
     action->args.connection.drain  = drain_mode;
 
     qdr_action_enqueue(core, action);
+    qdr_record_link_credit(core, link);
```
kgiusti: Curious - why after the action is enqueued, not before?

PR #607, opened 2019-11-01
DISPATCH-1466 - Initialize the host variable so it can be used anywhe…

ganeshmurthy:
…re inside the function


ganeshmurthy:
This problem was caused due to commit 76ef8420f8b87e4c3699462998d05b8d9156c1c2

PR #613, opened 2019-11-06
DISPATCH-1471: print strings that fail assert

ChugR:
This patch disables string truncation in all tests in the python suite derived from system_test/TestCase.
There's not a great way to write a test to see if this works.


ChugR:
Fixed at commit 3b7b8a1

PR #614, opened 2019-11-06
DISPATCH-1439 - Added two attributes to connection and one to router.

ganeshmurthy:
1. connection uptime - The number of seconds the connection has been up
2. connection last delivery seconds - The number of seconds since a delivery was sent on a connection
3. router uptime - the number of seconds since the router started

### DIFF
--- a/docs/man/qdstat.8.adoc
+++ b/docs/man/qdstat.8.adoc
@@ -73,6 +73,13 @@ The authentication method and user ID of the connection's authenticated user.
 tenant::
 If the connection is to a listener using multi-tenancy, this column displays the tenant namespace for the connection.
 
+last dlv::
+Time since the last delivery showed up on this connection. Displayed in the days:hours:minutes:seconds format.
+
+uptime::
+Time this connection has been up. Displayed in the days:hours:minutes:seconds format.
+
+
 qdstat -l
 ~~~~~~~~~
 type::
--- a/docs/man/qdstat.8.adoc
+++ b/docs/man/qdstat.8.adoc
@@ -74,7 +74,7 @@ tenant::
 If the connection is to a listener using multi-tenancy, this column displays the tenant namespace for the connection.
 
 last dlv::
-Time since the last delivery showed up on this connection. Displayed in the days:hours:minutes:seconds format.
+Time since the last delivery arrived on this connection. Displayed in the days:hours:minutes:seconds format.
 
 uptime::
 Time this connection has been up. Displayed in the days:hours:minutes:seconds format.
--- a/python/qpid_dispatch_internal/tools/display.py
+++ b/python/qpid_dispatch_internal/tools/display.py
@@ -45,7 +45,14 @@ def Commas(value):
     sval = left
 
 def TimeLong(value):
-  return strftime("%c", gmtime(value / 1000000000))
+  day = value // (24 * 3600)
+  time = value % (24 * 3600)
+  hour = time // 3600
+  time %= 3600
+  minutes = time // 60
+  time %= 60
+  seconds = time
+  return "%03d:%02d:%02d:%02d" % (day, hour, minutes, seconds)
 
 def TimeShort(value):
   return strftime("%X", gmtime(value / 1000000000))
--- a/src/router_core/agent_connection.c
+++ b/src/router_core/agent_connection.c
@@ -43,6 +43,8 @@
 #define QDR_CONNECTION_ACTIVE           18
 #define QDR_CONNECTION_ADMIN_STATUS     19
 #define QDR_CONNECTION_OPER_STATUS      20
+#define QDR_CONNECTION_UPTIME_SECONDS   21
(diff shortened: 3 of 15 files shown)


docs/man/qdstat.8.adoc
```diff
@@ -73,6 +73,13 @@ The authentication method and user ID of the connection's authenticated user.
 tenant::
 If the connection is to a listener using multi-tenancy, this column displays the tenant namespace for the connection.
 
+last dlv::
+Time since the last delivery showed up on this connection. Displayed in the days:hours:minutes:seconds format.
```
bhardesty: Maybe "arrived" instead of "showed up"?

src/router_core/agent_connection.c
```diff
@@ -240,6 +245,17 @@ static void qdr_connection_insert_column_CT(qdr_core_t *core, qdr_connection_t *
...
 
+    case QDR_CONNECTION_UPTIME_SECONDS:
+        qd_compose_insert_uint(body, core->uptime_ticks - conn->conn_uptime);
+        break;
+
+    case QDR_CONNECTION_LAST_DLV_SECONDS:
+        if (conn->is_delivery_sent)
+            qd_compose_insert_uint(body, core->uptime_ticks - conn->last_delivery_time);
+        else
+            qd_compose_insert_string(body, "-");
```
ted-ross: Adding hyphens to fields is the job of qdstat.  It should not be done here (I don't believe there is a precedent for it).  A better option is to insert a null value if there is no information to send.

src/router_core/agent_router.c
```diff
@@ -233,6 +235,13 @@ static void qdr_agent_write_column_CT(qd_composed_field_t *body, int col, qdr_co
         qd_compose_insert_uint(body, core->links_blocked);
         break;
 
+    case QDR_ROUTER_UPTIME_SECONDS:
+        if (core->uptime_ticks== 0)
+            qd_compose_insert_uint(body, 0);
```
ted-ross: This if-then-else doesn't seem to be any different from simply calling qs_compose_insert_uint.

src/router_core/router_core_private.h
```diff
@@ -659,6 +659,9 @@ struct qdr_connection_t {
     qdr_conn_admin_status_t     admin_status;
     qdr_error_t                *error;
     bool                        closed; // This bit is used in the case where a client is trying to force close this connection.
+    uint32_t                    conn_uptime; // The number of seconds that this connection has been up and running.
+    uint32_t                    last_delivery_time; // The number of seconds since the last delivery arrived on this connection.
```
ted-ross: Both of these comments are incorrect.  These are timestamps, not durations.

src/router_core/router_core_private.h
```diff
@@ -659,6 +659,9 @@ struct qdr_connection_t {
...
+    bool                        is_delivery_sent;
```
ted-ross: Couldn't this boolean be avoided by saying that last_delivery_time == 0 if no deliveries have been sent?

python/qpid_dispatch_internal/tools/display.py
```diff
@@ -45,7 +45,14 @@ def Commas(value):
...
 def TimeLong(value):
-  return strftime("%c", gmtime(value / 1000000000))
+  day = value // (24 * 3600)
+  time = value % (24 * 3600)
+  hour = time // 3600
+  time %= 3600
+  minutes = time // 60
+  time %= 60
+  seconds = time
+  return "%03d:%02d:%02d:%02d" % (day, hour, minutes, seconds)
```
ted-ross: I'm curious as to why this was done.  Does the strftime function no format the data properly?

ganeshmurthy: I wanted qdstat to display in days:hours:minutes:seconds (we have a 3 digit limit on the days). I assumed that the router (or a connection) will not be up for more than 999 days.

That being said, strftime does not directly provide what I want. It rolls the the day into the year after 365 days. Then I need to use the year to calculate the days which becomes difficult if a leap year is involved. I could do the following and display the output in years:days:hours:minutes:seconds: 

value=150000000
year_from_epoch = strftime("%Y", gmtime(value))
year=int(year_from_epoch) - 1970
print ("Output=%s:%s" % (str(year), strftime("%j:%X", gmtime(value))))

[gmurthy@localhost test]$ python strtime.py 
Output=4:276:02:40:00
[gmurthy@localhost test]$ 

With all this in mind, I decided to write a simple custom function myself.

src/router_core/router_core_private.h
ganeshmurthy: Yes, agreed

src/router_core/agent_connection.c
ganeshmurthy: Changed this to insert null instead of "-"

src/router_core/router_core_private.h
ganeshmurthy: Fixed comment

PR #615, opened 2019-11-07
DISPATCH-1473: python2.7 does not support mutually exclusive arguments

ChugR:


ChugR:
Fixed at commit faa7dd7

PR #617, opened 2019-11-12
DISPATCH-1472 - Add new section to qdstat man page for qdstat -g

bhardesty:


docs/man/qdstat.8.adoc
```diff
@@ -41,6 +41,89 @@ include::{generated_man}/qdstat_help.adoc[]
 OUTPUT COLUMNS
 --------------
 
+qdstat -g
+~~~~~~~~~
+
+Version::
+The version of Dispatch Router.
+
+Mode::
```
ganeshmurthy: The router also has a standalone mode 

https://github.com/apache/qpid-dispatch/blob/master/python/qpid_dispatch/management/qdrouter.json#L388

PR #618, opened 2019-11-13
DISPATCH-1481: Multicast links should block until a consumer is present

kgiusti:
This fixes an infinite loop bug which occurs when using multicast
addresses on inbound waypoint (broker) links.

### DIFF
--- a/docs/books/modules/user-guide/routing.adoc
+++ b/docs/books/modules/user-guide/routing.adoc
@@ -116,6 +116,14 @@ This involves configuring an address with a routing pattern. All messages sent t
 +
 This involves configuring a waypoint address to identify the broker queue and then connecting the router to the broker. All messages sent to the waypoint address will be routed to the broker queue.
 
+=== Message Routing Flow Control
+
+{RouterName} uses a _credit-based_ flow control mechanism to ensure that producers can only send messages to a router if at least one consumer is available to receive them. Because {RouterName} does not store messages, this credit-based flow control prevents producers from sending messages when there are no consumers present.
+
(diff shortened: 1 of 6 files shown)


docs/books/modules/user-guide/routing.adoc
```diff
@@ -116,6 +116,10 @@ This involves configuring an address with a routing pattern. All messages sent t
 +
 This involves configuring a waypoint address to identify the broker queue and then connecting the router to the broker. All messages sent to the waypoint address will be routed to the broker queue.
 
+=== Sending Messages to {RouterName}
+
+{RouterName} moves messages directly from publishers to consumers.  The router does not provide message storage as would a broker.  A message sent to the router must have at least one consumer present in order to properly route the message.  The router uses a _credit_ based flow control mechanism to prevent publishers from passing messages to the router when there are no consumers present.  A client wishing to publish a message to the router must wait until the router has provided it with credit.  Attempting to publish a message without credit available will cause the client to block.  Once credit is made available the client will unblock and the message will be sent to the router.  Most AMQP client libraries provide a means for determining the amount of credit available to a publisher.
+
 === Addresses
```
bhardesty: Took a stab at revising this:

=== Message Routing Flow Control

{RouterName} uses a \_credit-based\_ flow control mechanism to ensure that producers can only send messages to a router if at least one consumer is available to receive them. Because {RouterName} does not store messages, this credit-based flow control prevents producers from sending messages when there are no consumers present.

A client wishing to send a message to the router must wait until the router has provided it with credit. Attempting to publish a message without credit available will cause the client to block. Once credit is made available, the client will unblock, and the message will be sent to the router.

NOTE: Most AMQP client libraries enable you to determine the amount of credit available to a producer. For more information, consult your client's documentation.

kgiusti: thanks - I've updated the patch with your changes

PR #619, opened 2019-11-13
DISPATCH-1480: properly settle messages received by the edge address …

kgiusti:
…proxy

### DIFF
--- a/src/router_core/modules/edge_router/addr_proxy.c
+++ b/src/router_core/modules/edge_router/addr_proxy.c
@@ -424,6 +424,7 @@ static void on_transfer(void           *link_context,
                         qd_message_t   *msg)
 {
     qcm_edge_addr_proxy_t *ap = (qcm_edge_addr_proxy_t*) link_context;
+    uint64_t dispo = PN_ACCEPTED;
 
     //
     // Validate the message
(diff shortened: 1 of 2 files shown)


src/router_core/modules/edge_router/addr_proxy.c
```diff
@@ -460,8 +461,14 @@ static void on_transfer(void           *link_context,
 
         qd_parse_free(body);
         qd_iterator_free(iter);
+    } else {
+        qd_log(ap->core->log, QD_LOG_WARNING,
+               "Edge Address Proxy: received an invalid message body, rejecting");
+        dispo = PN_REJECTED;
     }
 
+    qdrc_endpoint_settle_CT(ap->core, dlv, dispo);
```
ganeshmurthy: Did Address Sanitizer find this bug during run time presumably when the addr_proxy's on_transfer() received an invalid message body? In other words, was an invalid message body sent to this function? If yes, in addition to this fix, should the caller also be fixed to send a valid message?

kgiusti: No invalid messages were involved.   All deliveries that end up received by this endpoint client simply were never settled.  
I'm not aware of any instance where invalid messages have been received by this code, however I want to add the extra logging IF at some point the code breaks and actually generates an invalid message.

However I think you raise a good point:  It would be best to make sure the sender can recover if a PN_REJECTED outcome is returned.  Let me check that before this patch lands.

kgiusti: Turns out that all messages sent to this client are PRE-settled, so setting the disposition to PN_REJECTED is harmless.

I think this patch is good to go.

PR #625, opened 2019-11-21
DISPATCH-1488: properly forward tranaction coordinator disposition data

kgiusti:
Will add unit test in a following patch.


gemmellr:
The promised test never materialised before this was resolved. I would like to stop finding similar basic issues over time that occurred purely because there are no tests for the related functionality.

PR #626, opened 2019-11-21
DISPATCH-1491: use sets rather than iterating over lists

grs:


python/qpid_dispatch_internal/router/mobile.py
```diff
@@ -54,43 +54,44 @@ def tick(self, now):
...
+            msg = MessageMAU(None, self.id, self.mobile_seq, list(self.added_addrs), list(self.deleted_addrs), _hints=hints)
 
             self.sent_deltas[self.mobile_seq] = msg
             if len(self.sent_deltas) > MAX_KEPT_DELTAS:
                 self.sent_deltas.pop(self.mobile_seq - MAX_KEPT_DELTAS)
 
             self.container.send('amqp:/_topo/0/all/qdrouter.ma', msg)
             self.container.log_ma(LOG_TRACE, "SENT: %r" % msg)
-            self.local_addrs.extend(self.added_addrs)
+            for addr in self.added_addrs:
```
ssorj: I think you could do "self.local_addrs.update(self.added_addrs)" here.  It may or may not make a difference to performance.

ssorj: I tried this out for fun: https://gist.github.com/ssorj/fe624ab4b3840db34d9ebbff5f04cb6b

grs: So altered.

PR #629, opened 2019-11-22
DISPATCH-1492 - Prevent router crash when senders connect to edge lis…

ganeshmurthy:
…teners on interior routers


ChugR:
+!
I can connect to the edge listener as a client and the system does not crash. The edge listener behaves  well as a normal client listener

PR #630, opened 2019-11-25
DISPATCH-1186: Add qdstat option to get CSV format for tables

ChugR:
Add a "--csv" switch to qdstat command line to apply to display tables.

|| CSV character  || value        ||
|  separator      |  comma        |
|  string quote   |  double quote |

All non-blank values in headings and tables are output as quoted strings.
Blank values in tables are not quoted. The output shall be consecutive
separator commas.

This option removes ambiguity for certain tables (like 'qdstat -l') where
columns (like peer and phs) have no entry, and headers (like 'conn id') have
spaces in their names.

### DIFF
--- a/python/qpid_dispatch_internal/tools/__init__.py
+++ b/python/qpid_dispatch_internal/tools/__init__.py
@@ -22,6 +22,6 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import print_function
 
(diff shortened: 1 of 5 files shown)


ganeshmurthy:
1. Tested all qdstat option combinations with --csv and looked at output in LibreOffice - looks good
2. All system tests pass with this PR.

Approving this PR. Ready to go into master.

ChugR:
Fixed at commit 7a10a4d

PR #631, opened 2019-11-26
DISPATCH-1475: prevent I/O thread from accessing link during deletion

kgiusti:
Fix includes consistently adding link to proper priority slot in
connections links_with_work array.


ChugR:
This commit works well. I've blasted my setup for a couple of hours,  250,000 reconnects, all with no issue. Ship it!

PR #636, opened 2019-12-02
DISPATCH-1496: Reset link variables after the drain is complete. Also…

ganeshmurthy:
… removed unused drain flag and related functionality

### DIFF
--- a/include/qpid/dispatch/router_core.h
+++ b/include/qpid/dispatch/router_core.h
@@ -674,6 +674,15 @@ int qdr_link_process_deliveries(qdr_core_t *core, qdr_link_t *link, int credit);
 
 void qdr_link_flow(qdr_core_t *core, qdr_link_t *link, int credit, bool drain_mode);
 
+/**
+ * The passed in link has been drained, it is no longer in drain mode. Sets the link's drain flag to false and sets credit to core to zero.
+ * Call this right after calling pn_link_drained
+ *
+ * @param core - router core
+ * @param link - the link that has been drained
+ */
+void qdr_link_drained(qdr_core_t *core, qdr_link_t *link);
+
 typedef void (*qdr_link_first_attach_t)  (void *context, qdr_connection_t *conn, qdr_link_t *link, 
                                           qdr_terminus_t *source, qdr_terminus_t *target);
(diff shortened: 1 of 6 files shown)


ganeshmurthy:
@gemmellr can you please test this and let me know if this works ? It fixes the additional credit problem and passes all existing drain tests in the test suite.
I will write a test for this tomorrow.

@ted-ross please take a look as well.

ganeshmurthy:
I added a test to verify the fix

gemmellr:
No idea on the code changes, but I gave it a try out and all seemed well, the router no longer granted extra credit itself.

src/router_node.c
```diff
@@ -1531,8 +1531,11 @@ static void CORE_link_drained(void *context, qdr_link_t *link)
 
     pn_link_t *plink = qd_link_pn(qlink);
 
-    if (plink)
+    if (plink) {
         pn_link_drained(plink);
+        link->drain_mode = false;
+        link->credit_to_core = 0;
```
ted-ross: Are these operations safe to do here?  This function is invoked on an IO thread.

include/qpid/dispatch/router_core.h
```diff
@@ -674,6 +674,15 @@ int qdr_link_process_deliveries(qdr_core_t *core, qdr_link_t *link, int credit);
...
 void qdr_link_flow(qdr_core_t *core, qdr_link_t *link, int credit, bool drain_mode);
 
+/**
+ * The passed in link has been drained, it is no longer in drain mode. Sets the link's drain flag to false and sets credit to core to zero.
+ * Call this right after calling pn_link_drained
+ *
+ * @param core - router core
+ * @param link - the link that has been drained
+ */
+void qdr_link_drained(qdr_core_t *core, qdr_link_t *link);
```
ssorj: I know this is the naming Proton has, but it's confusing.  'qdr_link_drained' looks like a predicate.  Something like 'qdr_link_mark_drained' or '_set_drained' would be a lot clearer IMO.

ganeshmurthy: I will change this to qdr_link_set_drained. Thanks.

PR #638, opened 2019-12-03
DISPATCH-1488: add test to verify transanction handling

kgiusti:


gemmellr:
Not so much with the test for now then after all, if I'm understanding the code/comments? :)

kgiusti:
At this point that's the best we can do.  I'm going to talk a bit further with the proton devs - either I'm doing it wrong or the python api needs to be enhanced.

PR #639, opened 2019-12-05
DISPATCH-1512 - Fixed variable name in for loop in unmap_all_addresses

ganeshmurthy:


grs:
Oops, sorry, I think that was me! Thanks for fixing @ganeshmurthy!

PR #641, opened 2019-12-10
DISPATCH-1514 - Dynamically turned on proton frame trace on existing …

ganeshmurthy:
…connections

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -515,6 +515,7 @@ qd_error_t qd_log_entity(qd_entity_t *entity) {
     char* module = 0;
     char *output = 0;
     char *enable = 0;
+    bool trace_enabled = false;
 
     do {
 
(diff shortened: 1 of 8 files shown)


src/server.c
```diff
@@ -1276,6 +1277,25 @@ void qd_server_set_container(qd_dispatch_t *qd, qd_container_t *container)
     qd->server->container = container;
 }
 
+void qd_server_trace_all_connections()
+{
+    qd_dispatch_t *qd = qd_dispatch_get_dispatch();
+    if (qd->server) {
+        qd_connection_list_t  conn_list = qd->server->conn_list;
+        qd_connection_t *conn = DEQ_HEAD(conn_list);
+        while(conn) {
```
kgiusti: the qd->server->conn_list is protected by qd->server->lock.  this loop will need to hold that lock while walking the list or it is possible the connection it is working on is removed (and possibly freed) by another thread.

src/log.c
```diff
@@ -541,7 +542,16 @@ qd_error_t qd_log_entity(qd_entity_t *entity) {
...
             enable = qd_entity_get_string(entity, "enable");
             QD_ERROR_BREAK();
+
             src->mask = enable_mask(enable);
+
+            //
+            // If trace logging is enabled, loop thru all connections in the router and call the pn_transport_set_tracer callback
+            // so proton frame trace can be output as part of the router trace log.
+            //
+            if (qd_log_enabled(src, QD_LOG_TRACE)) {
```
kgiusti: would it be possible to move this outside of the log_source_lock?  For example, just set a flag here then check the flag after dropping the lock and making the call to qd_server_trace_all_connections() then?

PR #645, opened 2019-12-11
DISPATCH-1516 - Modified log statements to include link identifier. A…

ganeshmurthy:
…lso logged delivery outcome

### DIFF
--- a/src/router_core/delivery.c
+++ b/src/router_core/delivery.c
@@ -376,7 +376,7 @@ void qdr_delivery_increment_counters_CT(qdr_core_t *core, qdr_delivery_t *delive
                 core->modified_deliveries++;
         }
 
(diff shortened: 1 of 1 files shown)


src/router_core/delivery.c
```diff
@@ -373,6 +376,8 @@ void qdr_delivery_increment_counters_CT(qdr_core_t *core, qdr_delivery_t *delive
                 core->modified_deliveries++;
         }
 
+        qd_log(core->log, QD_LOG_DEBUG, delivery->presettled?"Delivery outcome for pre-settled: dlv:%lx link:%"PRIu64" is %s": "Delivery outcome for: dlv:%lx link:%"PRIu64" is %s", (long) delivery,  link->identity, pn_disposition_type_name(outcome));
```
kgiusti: I had to read this a couple of times to figure out what's going on.
I'd avoid using a ?: to pick the format argument - that will work but it's.... clever.

Why not use a format string with a %s entry for the part of the output that varies, i.e.:
qd_log(..., ..., "Delivery outcome for%s: dlv:%lx link:<etc, etc, etc> is %s", 
            (delivery->presettled) ? " pre-settled" : "",
....)

Note the lack of space between "for" and the ":" in the format string and the leading space before "presettled"

PR #654, opened 2019-12-18
DISPATCH-1532 - Reimplemented mobile-address-synchronization as a cor…

ted-ross:
…e module.

https://issues.apache.org/jira/browse/DISPATCH-1532

No functionality was added in this PR.  I've done some manual testing to ensure backward compatibility with the former implementation.


src/router_core/route_tables.c
```diff
@@ -784,19 +852,19 @@ static void qdr_do_link_lost(qdr_core_t *core, qdr_general_work_t *work)
...
 
 
-void qdr_post_mobile_added_CT(qdr_core_t *core, const char *address_hash, qd_address_treatment_t treatment)
+void qdr_post_set_mobile_seq_CT(qdr_core_t *core, int router_maskbit, uint64_t mobile_seq)
 {
-    qdr_general_work_t *work = qdr_general_work(qdr_do_mobile_added);
-    work->field = qdr_field(address_hash);
-    work->treatment = treatment;
+    qdr_general_work_t *work = qdr_general_work(qdr_do_set_mobile_seq);
+    work->in_conn_id = mobile_seq;
```
kgiusti: I see what you did there.  Bad Ted: no soup for you!

ted-ross: Ok... You're right...  I fixed it.

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -0,0 +1,860 @@
...
+
+//
+// Address.sync_mask bit values
+//
+#define ADDR_SYNC_DELETION_WAS_BLOCKED  0x00000001
+#define ADDR_SYNC_IN_ADD_LIST           0x00000002
+#define ADDR_SYNC_IN_DEL_LIST           0x00000004
+#define ADDR_SYNC_TO_BE_DELETED         0x00000008
+
+#define BIT_SET(M,B)   M |= B
```
kgiusti: Any objection to moving these into ctools.h?
If so I'd recommend you put parenthesis around (B) in the assignments and around both M and B in the check.

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -0,0 +1,860 @@
...
+static void qcm_mobile_sync_compose_diff_addr_list(qdrm_mobile_sync_t *msync, qd_composed_field_t *field, bool is_added)
+{
+    qdr_address_list_t *list = is_added ? &msync->added_addrs : &msync->deleted_addrs;
+
+    qd_compose_start_list(field);
+    qdr_address_t *addr = DEQ_HEAD(*list);
+    while (addr) {
+        const char *hash_key = (const char*) qd_hash_key_by_handle(addr->hash_handle);
+        qd_compose_insert_string(field, hash_key);
+        if (is_added) {
```
kgiusti: Just a suggestion:  you _could_ move this check out of the address loop by setting the list name (SYNC_ADD/DEL) and bit in locals before the loop (like you set the list var).

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -0,0 +1,860 @@
...
+static const char *HINTS      = "hints";
+static const char *ADD        = "add";
+static const char *DEL        = "del";
+static const char *EXIST      = "exist";
+static const char *HAVE_SEQ   = "have_seq";
+
+//
+// Address.sync_mask bit values
+//
+#define ADDR_SYNC_DELETION_WAS_BLOCKED  0x00000001
```
kgiusti: Could this flag handling the blocking of the address deletion be better implemented by turning addr->block_deletion into a reference counter instead?

kgiusti: Not a pointer reference counter, to be clearer.   A counter that prevents deletion when incremented?  Call qdr_check_addr_CT when decrementing causes it to hit zero?

ChugR:
At commit 9198a4 this patch occasionally hangs during the first self test in system_tests_multicast. Scraping the logs after a hang timeout reveals that after a bunch of setup the test sends 10 messages with leading content :

> S-INT_A-1 -> 9: MCAST PRESETTLED LARGE RX DETACH

Then the code falls into a loop doing management queries:

> $amqp:/_edge/EA1/temp.85H3ENeh1E1t9hs 33 St name self type org.amqp.management operation QUERY entityType 'org.apache.qpid.dispatch.router.address

and

> (amqp:/_topo/0/INT.A/temp.zgRSn+05d+QH7t8 34 St name self type org.amqp.management operation QUERY entityType 'org.apache.qpid.dispatch.router.address 

The query stream is repeated 5000 or so times before the test times out after 600 seconds.

ted-ross:
> At commit 9198a4 this patch occasionally hangs during the first self test in system_tests_multicast. Scraping the logs after a hang timeout reveals that after a bunch of setup the test sends 10 messages with leading content :
> 
> > S-INT_A-1 -> 9: MCAST PRESETTLED LARGE RX DETACH
> 
> Then the code falls into a loop doing management queries:
> 
> > $amqp:/_edge/EA1/temp.85H3ENeh1E1t9hs 33 St name self type org.amqp.management operation QUERY entityType 'org.apache.qpid.dispatch.router.address
> 
> and
> 
> > (amqp:/_topo/0/INT.A/temp.zgRSn+05d+QH7t8 34 St name self type org.amqp.management operation QUERY entityType 'org.apache.qpid.dispatch.router.address
> 
> The query stream is repeated 5000 or so times before the test times out after 600 seconds.

This was caused by a bug in the MAU generation that caused routers to hold stale MA state from other routers.  The PR has been updated with a fix.

ted-ross:
I accidentally closed the PR... reopening.

src/router_core/modules/mobile_sync/mobile.c
kgiusti: I'm ok with leaving it as-is for now and moving it after the patch lands.

src/router_core/modules/mobile_sync/mobile.c
ted-ross: Thank you for this suggestion.  It has been incorporated in the latest squash/rebase.

PR #657, opened 2020-01-08
DISPATCH-1544: mask coverity false positive delivery use-after-free

kgiusti:
This patch just adds some debug asserts and extra comments to ensure the delivery reference counts are correct (match the expectations of the code).


ganeshmurthy:
This commit only adds asserts ?

kgiusti:
The patch also annotates the code so coverity won't report the use after free issue (see the /* coverity .... */ comments this patch adds).

The asserts are there to prevent future code changes from inadvertently messing up the reference counters - which _could_ introduce a real use-after-free bug which coverity will ignore due to the annotation.

PR #658, opened 2020-01-10
DISPATCH-1540 - Set the pre-settled flag appropriately on a delivery …

ganeshmurthy:
…and its peers on large streaming messages. This will allow the presettled counters to be updated correctly

### DIFF
--- a/src/router_core/delivery.c
+++ b/src/router_core/delivery.c
@@ -203,19 +203,13 @@ void qdr_delivery_remote_state_updated(qdr_core_t *core, qdr_delivery_t *deliver
 
 qdr_delivery_t *qdr_deliver_continue(qdr_core_t *core,qdr_delivery_t *in_dlv, bool settled)
 {
-    //
-    // If the delivery is already pre-settled, don't do anything with the pre-settled flag.
-    //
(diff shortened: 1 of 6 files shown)


src/router_core/delivery.c
```diff
@@ -201,8 +201,16 @@ void qdr_delivery_remote_state_updated(qdr_core_t *core, qdr_delivery_t *deliver
...
 
-qdr_delivery_t *qdr_deliver_continue(qdr_core_t *core,qdr_delivery_t *in_dlv)
+qdr_delivery_t *qdr_deliver_continue(qdr_core_t *core,qdr_delivery_t *in_dlv, bool settled)
 {
+    //
+    // If the delivery is already pre-settled, don't do anything with the pre-settled flag.
+    //
+    // If the in_delivery was not pre-settled, you can go to pre-settled.
+    if (! in_dlv->presettled && settled) {
+        in_dlv->presettled = settled;
```
kgiusti: This is (technically) a possible race condition.  This method runs under the I/O thread which should not be modifying qdr_delivery_t state directly.

It would be better to pass along the presettled flag in the action.   I suggest considering changing the existing code to use the action->args.delivery action structure (adding a pre-settled flag there) instead of the args.connection structure.

PR #662, opened 2020-01-15
DISPATCH-1547: add tests for router table updates

kgiusti:


tests/system_tests_topology.py
```diff
@@ -592,5 +594,182 @@ def run(self):
...
+        # wait for INT_A mobile addresses to be removed from INT_C
+        mgmt = INT_C.management
+        a_type = 'org.apache.qpid.dispatch.router.address'
+        rsp = mgmt.query(a_type).get_dicts()
+        while any(map(lambda a: a['name'].find('closest/on_A') != -1, rsp)):
+            time.sleep(0.25)
+            rsp = mgmt.query(a_type).get_dicts()
+
+        # cleanup
+        for c in consumers:
```
ganeshmurthy: Can we please move this loop to close the consumer right after the INT_A.teardown() call just to be clear that the consumers are gone forever and will not try to reconnect after the router is bounced.

kgiusti: done

ganeshmurthy:
Approved !

PR #668, opened 2020-01-20
DISPATCH-1509: free core agent timer on shutdown

kgiusti:
Refactored the core agent code by pulling out the agent fields from qdr_core_t and putting them in their own private structure with setup/cleanup methods.

### DIFF
--- a/src/router_core/agent.c
+++ b/src/router_core/agent.c
@@ -39,6 +39,18 @@ static void qdr_manage_update_CT(qdr_core_t *core, qdr_action_t *action, bool di
 ALLOC_DECLARE(qdr_query_t);
 ALLOC_DEFINE(qdr_query_t);
 
+
+struct qdr_agent_t {
+    qdr_query_list_t       outgoing_query_list;
+    sys_mutex_t           *query_lock;
+    qd_timer_t            *timer;
+    qdr_manage_response_t  response_handler;
+    qdr_subscription_t    *subscription_mobile;
+    qdr_subscription_t    *subscription_local;
+    qd_log_source_t       *log_source;
+};
+
+
 //==================================================================================
 // Internal Functions
 //==================================================================================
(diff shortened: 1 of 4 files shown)


src/router_core/agent.c
```diff
@@ -39,27 +39,40 @@ static void qdr_manage_update_CT(qdr_core_t *core, qdr_action_t *action, bool di
 ALLOC_DECLARE(qdr_query_t);
 ALLOC_DEFINE(qdr_query_t);
 
+
+struct qdr_agent_t {
```
ganeshmurthy: since the object is already called qdr_agent_t is it redundant to have the word agent for its members ? For example, agent->agent_response_handler can just be called agent->response_handler, same with agent_subscription_mobile, agent_subscription_mobile, agent_subscription_local and agent_log_source

src/router_core/agent.c
```diff
@@ -107,6 +122,54 @@ static void qdr_agent_set_columns(qdr_query_t *query, qd_parsed_field_t *attribu
...
+        qd_timer_free(agent->agent_timer);
+        if (agent->query_lock)
+            sys_mutex_free(agent->query_lock);
+
+        //we can't call qdr_core_unsubscribe on the subscriptions because the action processing thread has
+        //already been shut down. But, all the action would have done at this point is free the subscriptions
+        //so we just do that directly.
+        free(agent->agent_subscription_mobile);
+        free(agent->agent_subscription_local);
+
```
ganeshmurthy: agent->agent_log_source is not freed here. Is it freed elsewhere ?

src/router_core/router_core_private.h
```diff
@@ -800,15 +801,7 @@ struct qdr_core_t {
...
-    //
-    // Agent section
-    //
-    qdr_query_list_t       outgoing_query_list;
-    sys_mutex_t           *query_lock;
-    qd_timer_t            *agent_timer;
-    qdr_manage_response_t  agent_response_handler;
-    qdr_subscription_t    *agent_subscription_mobile;
-    qdr_subscription_t    *agent_subscription_local;
+    qdr_agent_t               *agent;
```
ganeshmurthy: I wonder if this should be mgmt_agent instead of just agent because it is used to respond exclusively to management requests coming in on $management

src/router_core/router_core_private.h
```diff
@@ -920,7 +913,9 @@ void *router_core_thread(void *arg);
 uint64_t qdr_identifier(qdr_core_t* core);
 void qdr_management_agent_on_message(void *context, qd_message_t *msg, int link_id, int cost, uint64_t in_conn_id);
 void  qdr_route_table_setup_CT(qdr_core_t *core);
-void  qdr_agent_setup_CT(qdr_core_t *core);
+qdr_agent_t *qdr_agent(qdr_core_t *core);
```
ganeshmurthy: Should this be called qdr_agent_CT instead?

src/router_core/agent.c
kgiusti: I'm all for shorter names - I'll update the patch.

src/router_core/agent.c
kgiusti: Turns out the log module frees all the log sources up on shutdown.

src/router_core/router_core_private.h
kgiusti: +1

src/router_core/router_core_private.h
kgiusti: No - it's called before the core thread is running.

PR #673, opened 2020-01-22
DISPATCH-1556: Doc - add step to set sasldb permissions

bhardesty:


docs/books/modules/user-guide/enabling-username-password-authentication.adoc
```diff
@@ -55,6 +55,19 @@ The full user name is the user name you entered plus the domain name (`__<user-n
 ====
 --
 
+. Ensure that the `qdrouterd` process can read the SASL database.
++
+--
+The `qdrouterd` process may run as an unprivileged user. Therefore, you may need to adjust the permissions or ownership of the SASL database so that the router can read it.
```
jbyrne-redhat: It might be worth rethinking the two uses of 'may' in this paragraph. 

Specifically, I wasn't sure if the first sentence is intended to imply _permission_ (that is, the `qdrouterd` process is _allowed to_ run as an unprivileged user) or _ambiguity_ (the `qdrouterd` process _might_ run as an unprivileged user).

The second sentence is more clear-cut. I would consider rewording to "Therefore, you might need to adjust..."

PR #676, opened 2020-01-31
DISPATCH-1518: Added ability to turn on protocol frame trace logging …

ganeshmurthy:
…for a particular connection

### DIFF
--- a/python/qpid_dispatch/management/client.py
+++ b/python/qpid_dispatch/management/client.py
@@ -76,6 +76,16 @@ class Entity(EntityBase):
 class Node(object):
     """Client proxy for an AMQP management node"""
 
+    def clean_attrs(self, attrs):
+        BOOL_VALUES = {"yes": 1, "true": 1, "on": 1, "no": 0, "false": 0,
+                       "off": 0}
+        if isinstance(attrs, dict):
+            for key in attrs.keys():
+                if attrs[key] in BOOL_VALUES.keys():
+                    attrs[key] = bool(BOOL_VALUES[attrs[key]])
+
+        return attrs
+
     @staticmethod
     def connection(url=None, router=None, timeout=10, ssl_domain=None, sasl=None):
         """Return a BlockingConnection suitable for connecting to a management node
@@ -332,7 +342,7 @@ class Node(object):
         identity = identity or attributes.get(u'identity')
         if name and identity: name = None # Only send one
         request = self.request(operation=U'UPDATE', type=type, name=name,
-                               identity=identity, body=attributes)
+                               identity=identity, body=self.clean_attrs(attributes))
         return Entity(self, self.call(request).body)
 
     def delete(self, type=None, name=None, identity=None):
--- a/python/qpid_dispatch/management/client.py
+++ b/python/qpid_dispatch/management/client.py
@@ -77,12 +77,16 @@ class Node(object):
     """Client proxy for an AMQP management node"""
 
     def clean_attrs(self, attrs):
-        BOOL_VALUES = {"yes": 1, "true": 1, "on": 1, "no": 0, "false": 0,
-                       "off": 0}
+        BOOL_VALUES = {"yes"  : True,
+                       "true" : True,
+                       "on"   : True,
+                       "no"   : False,
+                       "false": False,
+                       "off"  : False}
         if isinstance(attrs, dict):
             for key in attrs.keys():
                 if attrs[key] in BOOL_VALUES.keys():
-                    attrs[key] = bool(BOOL_VALUES[attrs[key]])
+                    attrs[key] = BOOL_VALUES[attrs[key]]
 
         return attrs
 
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -1754,6 +1754,7 @@
                     "description": "Indicates whether protocol level amqp frame trace logging is turned on or off for this connection.",
                     "type": "boolean",
                     "create": false,
+                    "update": true,
                     "default": false
                 },
                 "sasl": {
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -1750,6 +1750,12 @@
                     "description": "Indicates whether the connection content is encrypted.",
                     "type": "boolean"
                 },
+                "enableProtocolTrace": {
+                    "description": "Indicates whether protocol level amqp frame trace logging is turned on or off for this connection.",
+                    "type": "boolean",
+                    "create": false,
+                    "default": false
+                },
                 "sasl": {
                     "description": "SASL mechanism in effect for authentication.",
                     "type": "string"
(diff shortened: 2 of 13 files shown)


python/qpid_dispatch/management/client.py
```diff
@@ -76,6 +76,16 @@ def delete(self):
 class Node(object):
     """Client proxy for an AMQP management node"""
 
+    def clean_attrs(self, attrs):
+        BOOL_VALUES = {"yes": 1, "true": 1, "on": 1, "no": 0, "false": 0,
+                       "off": 0}
+        if isinstance(attrs, dict):
+            for key in attrs.keys():
+                if attrs[key] in BOOL_VALUES.keys():
+                    attrs[key] = bool(BOOL_VALUES[attrs[key]])
```
kgiusti: Instead of putting 1 or 0 as the values for BOOL_VALUES, why not use the python bools True and False?  Then you don't need to cast to bool()

python/qpid_dispatch/management/client.py
```diff
@@ -76,6 +76,16 @@ def delete(self):
 class Node(object):
     """Client proxy for an AMQP management node"""
 
+    def clean_attrs(self, attrs):
+        BOOL_VALUES = {"yes": 1, "true": 1, "on": 1, "no": 0, "false": 0,
```
kgiusti: What about case?  Is 'Yes' true or only 'yes'?

python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1750,6 +1750,12 @@
                     "description": "Indicates whether the connection content is encrypted.",
                     "type": "boolean"
                 },
+                "enableProtocolTrace": {
+                    "description": "Indicates whether protocol level amqp frame trace logging is turned on or off for this connection.",
+                    "type": "boolean",
+                    "create": false,
+                    "default": false
+                },
```
kgiusti: Should this include "update": true?

src/router_core/router_core_private.h
```diff
@@ -681,6 +684,7 @@ struct qdr_connection_t {
     bool                        closed; // This bit is used in the case where a client is trying to force close this connection.
     uint32_t                    conn_uptime; // Timestamp which can be used to calculate the number of seconds this connection has been up and running.
     uint32_t                    last_delivery_time; // Timestamp which can be used to calculate the number of seconds since the last delivery arrived on this connection.
+    bool                        enable_protocol_trace; // Has trace level logging been turned on for this connecvtion.
```
kgiusti: "connecvtion"?

tests/system_tests_log_level_update.py
```diff
@@ -120,6 +123,189 @@ def test_turn_on_protocol_trace(self):
...
+
+    @classmethod
+    def setUpClass(cls):
+        """Start a router and a messenger"""
+        super(EnableConnectionLevelInterRouterTraceTest, cls).setUpClass()
+
+        def router(name, connection):
+
+            config = [
+                # Use the deprecated attributes helloInterval, raInterval, raIntervalFlux, remoteLsMaxAge
```
kgiusti: Huh?  This comment seems like a cut&paste artifact.

tests/system_tests_log_level_update.py
```diff
@@ -120,6 +123,189 @@ def test_turn_on_protocol_trace(self):
         self.assertTrue(num_attaches == 0)
 
 
+class EnableConnectionLevelInterRouterTraceTest(TestCase):
+
+    inter_router_port = None
+
+    @classmethod
+    def setUpClass(cls):
+        """Start a router and a messenger"""
```
kgiusti: Stale comment?

src/router_core/agent_connection.c
```diff
@@ -563,47 +568,103 @@ void qdra_connection_update_CT(qdr_core_t      *core,
...
+                    // Policy on the connection that is requesting that some other connection be deleted does not allow
+                    // for the other connection to be deleted.Set the status to QD_AMQP_FORBIDDEN and just quit.
+                    //
+                    query->status = QD_AMQP_FORBIDDEN;
+                    query->status.description = "You are not allowed to perform this operation.";
+                    qd_compose_start_map(query->body);
+                    qd_compose_end_map(query->body);
+                    admin_status_bad_or_forbidden = true;
+                 }
+                else if (admin_state) { //admin state and trace are the only fields that can be updated via the update management request for type connection.
```
kgiusti: Isn't this always true since if (admin_state) is checked above on line 578?

src/router_core/agent_connection.c
```diff
@@ -563,47 +568,103 @@ void qdra_connection_update_CT(qdr_core_t      *core,
...
+                    if (identity) {
+                        conn = qdr_connection_find_by_identity_CT(core, identity);
```
kgiusti: Same here - identity was checked above in line 562

src/router_core/agent_connection.c
```diff
@@ -563,47 +568,103 @@ void qdra_connection_update_CT(qdr_core_t      *core,
...
-        if (!user_conn) {
-            // This is bad. The user connection (that was requesting that some
-            // other connection be dropped) is gone
-            query->status.description = "Parent connection no longer exists";
+        //
+        // The only two fields that can be updated on a connection is the enableProtocolTrace flag and the admin state.
+        // If both these fields are not there, this is a bad request
+        // For example, this qdmanage is a bad request - qdmanage update --type=connection identity=1
+        //
+        if (!trace_field && !admin_state) {
```
kgiusti: why not check this first?  You've already checked admin_state above so this really is confusing.

src/router_core/agent_connection.c
```diff
@@ -563,47 +568,103 @@ void qdra_connection_update_CT(qdr_core_t      *core,
...
             qdra_connection_set_bad_request(query);
+            qdr_agent_enqueue_response_CT(core, query);
+            return;
         }
 
-        else {
-            if (!user_conn->policy_allow_admin_status_update) {
+        if (!conn) {
+            if (identity) {
```
kgiusti: again, already checked above.

python/qpid_dispatch/management/client.py
ganeshmurthy: This is tough question. I think 'Yes' should be invalid, so should 'YES' or 'YEs' or 'NO' or 'No'
Please let me know what your thoughts on that are.

src/router_node.c
```diff
@@ -1432,6 +1432,30 @@ static void CORE_link_second_attach(void *context, qdr_link_t *link, qdr_terminu
...
+    if (!tport)
+        return;
+
+    if (trace) {
+        pn_transport_trace(tport, PN_TRACE_FRM);
+        pn_transport_set_tracer(tport, connection_transport_tracer);
+
+    }
+    else {
+        pn_transport_set_tracer(tport, 0);
```
ChugR: When pn_tracer_t tracer is zero the an assert triggers in the proton code.

ganeshmurthy:
@kgiusti @ChugR I have added a new commit to this PR which addresses your code review comments. Please take a look and comment

PR #682, opened 2020-02-13
DISPATCH-1564 - Modified the way ulong is displayed on the output of …

ganeshmurthy:
…qdstat

### DIFF
--- a/python/qpid_dispatch_internal/tools/display.py
+++ b/python/qpid_dispatch_internal/tools/display.py
@@ -68,6 +68,8 @@ class Header:
   TIME_SHORT = 6
   DURATION = 7
   COMMAS = 8
+  # This is a plain number, no formatting
+  PLAIN_NUM = 9
 
   def __init__(self, text, format=NONE):
     self.text = text
@@ -85,6 +87,8 @@ class Header:
         return ''
       if self.format == Header.NONE:
         return value
+      if self.format == Header.PLAIN_NUM:
+        return PlainNum(value)
       if self.format == Header.KMG:
         return self.num(value)
       if self.format == Header.YN:
@@ -127,6 +131,7 @@ class Header:
       return "%2.1f%c" % (fp, tag)
     return "%4d%c" % (value / 1000, tag)
 
+
   def num(self, value):
     if value < 1000:
       return "%4d" % value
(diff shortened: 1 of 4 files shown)


python/qpid_dispatch_internal/tools/display.py
```diff
@@ -138,6 +143,33 @@ def num(self, value):
     value /= 1000
     return self.numCell(value, 'g')
 
+def PlainNum(value):
```
kgiusti: Is the precision check even necessary since the return string will be the minimum size in any case?  Could PlainNum simply return "%d" % value ?

tests/system_tests_console.py
```diff
@@ -51,7 +51,7 @@ def is_cmd(name):
             devnull = open(os.devnull, "w")
             subprocess.Popen([name], stdout=devnull, stderr=devnull).communicate()
         except OSError as e:
-            if e.errno == os.errno.ENOENT:
+            if errno == errno.ENOENT:
```
kgiusti: Finally!

python/qpid_dispatch_internal/tools/display.py
kgiusti: Also, why "PlainNum" and not "Integer", since %d requires an integer argument?

python/qpid_dispatch_internal/tools/display.py
```diff
@@ -138,6 +143,33 @@ def num(self, value):
...
+    if value < 10:
+      return "%1d" % value
```
kgiusti: You might want to wrap this in a try/except block to catch formatting errors.   Simply return "%s" % value on error.

Try  "%d" % "HI GANESH" in python -> TypeError

python/qpid_dispatch_internal/tools/display.py
ganeshmurthy: Yes, it could simply return   "%d" % value. I added the precision because I thought the numbers were not left justified if  "%d" % value was used but I was wrong. I ran the tests with "%d" % value. and the values came out correctly.

ganeshmurthy: I went with PlainNum since no formatting was done, do you want me to call it PlainInteger or just Integer ?

python/qpid_dispatch_internal/tools/display.py
ganeshmurthy: Will do !

tests/system_tests_console.py
ganeshmurthy: Did you mean you want to use a finally clause instead of except ?

python/qpid_dispatch_internal/tools/display.py
ganeshmurthy: I am keeping it as PlainNum since you were ok with it

tests/system_tests_console.py
kgiusti: Clarification "Finally someone got around to fixing this!"

PR #683, opened 2020-02-13
DISPATCH-1562: Changed the names of prometheus metrics to be in line …

ganeshmurthy:
…with prometheus recommendation


ganeshmurthy:
Naming rules are here - https://prometheus.io/docs/practices/naming/

Please take a look and let me know if you don't agree with any names that I came up with. @lulf has approved these changes.

PR #684, opened 2020-02-13
DISPATCH-1568: using doctest 

nicob87:
- Demo using the library that Jiri Daněk suggested in the issue, it seems to be a very nice option for several reasons, it is an evolution of catch2, compile time significantly smaller, support nested "subcases" which simplifyes setup and teardown, it is a single header file, and others.
see: https://github.com/onqtam/doctest

- jira: https://issues.apache.org/jira/browse/DISPATCH-1568 

- To run the demo, just:

```
[nbrignon@localhost build (nb-doctest)]$ cmake .. -DCMAKE_BUILD_TYPE=Debug
...
[nbrignon@localhost build (nb-doctest)]$ make
...
```
- and then we can execute the binary directly:

```
[nbrignon@localhost build (nb-doctest)]$ ./tests/c_unittests/c_unittests
[doctest] doctest version is "2.3.6"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:      5 |      5 passed |      0 failed |      0 skipped
[doctest] assertions:     15 |     15 passed |      0 failed |
[doctest] Status: SUCCESS!
```
- or debug it:
```
[nbrignon@localhost build (nb-doctest)]$ gdb ./tests/c_unittests/c_unittests
...
(gdb) b safe_snprintf
Breakpoint 1 at 0x408330
(gdb) run
...
Breakpoint 1, safe_snprintf (str=str@entry=0x7fffffffcd00 " \313\371\367\377\177", size=size@entry=19, format=format@entry=0x4228d2 "something")
    at /home/nbrignon/qpid-dispatch/src/router_core/terminus.c:89
89      size_t safe_snprintf(char *str, size_t size, const char *format, ...) {
```

- or execute it through ctest:
```
nbrignon@localhost build (nb-doctest)]$ ctest -VV -R c_unittest
...
test 1
    Start 1: c_unittest

1: Test command: /home/nbrignon/qpid-dispatch/build/tests/c_unittest
1: Test timeout computed to be: 600
1: [doctest] doctest version is "2.3.6"
1: [doctest] run with "--help" for options
1: ===============================================================================
1: [doctest] test cases:      5 |      5 passed |      0 failed |      0 skipped
1: [doctest] assertions:     15 |     15 passed |      0 failed |
1: [doctest] Status: SUCCESS!
1/1 Test #1: c_unittest .......................   Passed    0.01 sec

The following tests passed:
        c_unittest

100% tests passed, 0 tests failed out of 1


Total Test time (real) =   0.01 sec
```


tests/c_unittests_main.cpp
```diff
@@ -0,0 +1,2 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include "qdr_doctest.h"
```
jdanekrh:
```suggestion
#include "doctest.h"
```

jdanekrh: aha, right. I misunderstood what you are doing.

src/router_core/terminus.c
```diff
@@ -75,25 +76,47 @@ void qdr_terminus_free(qdr_terminus_t *term)
...
+    // max size allowed must be INT_MAX (since vsnprintf reutrns an int)
+    if (size == 0 || size > INT_MAX) {
+        //TODO log a warning somewhere?
+        return 0;
+    }
+    int max_possible_return_value = (int)(size - 1);
     va_list ap;
     va_start(ap, format);
+#ifdef TESTING
+    int rc = mocked_vsnprintf(str, size, format, ap);
```
jdanekrh: I was thinking about this and I don't like this much. I understand that in C, there isn't many options to replace vsnprintf calls in tests, but this "pollutes" the production code a lot.

In the C++ world, they seem to call these techniques "seams", [1] [2] [3]. Dispatch already uses the linker seam in two places, to mock `qd_server_timeout` and `qd_timer_now`.

I generally like using the linker better, it seems cleaner to me.

[1] http://www.informit.com/articles/article.aspx?p=359417&seqNum=3
[2] https://accu.org/index.php/journals/1927 Refactoring Towards Seams in C++
[3] https://cute-test.com/guides/mocking-with-cute/

tests/test_terminus_mocks.c
```diff
@@ -0,0 +1,32 @@
+#include <strings.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdarg.h>
+#include <inttypes.h>
+#include <limits.h>
```
nicob87: only string.h, stdbool.h and stdarg.h are needed here

src/router_core/terminus.c
```diff
@@ -75,25 +76,47 @@ void qdr_terminus_free(qdr_terminus_t *term)
     free_qdr_terminus_t(term);
 }
 
+#ifdef TESTING
+__attribute__((weak))
```
nicob87: no need to make it weak since this goes to a shared library. __attribute__((noinline)) probably is a good idea

src/router_core/terminus.c
nicob87: Hi @jdanekrh, Awesome review! Thanks.
About the preprocessor, yes, I agree that code in this pr is polluted and it is possible to do it cleaner, with a preprocessor "seam". More over that I do not like using preprocessor at all. And also I do not like to compile twice just to be able to use the preprocessor. BUT I was looking for a way to do not pollute the production "binary". I didn't know a couple of tricks explained in your documentation.
I think it is possible to do something like what is explained in [2] (Run-time function interception) with the addition here we are mixing c and c++. It looks like it may work :-P.
Will give it a try tomorrow or one of this days.

Regards!

tests/test_terminus_mocks.c
```diff
@@ -0,0 +1,71 @@
...
+}
+void mocked_value(int v)
+{
+    mocked_rc = v;
+}
+
+/* Even without any "#include" compiler complaing about unmatching with stdio.h
+ * vsnprintf. */
+/*-Werror=builtin-declaration-mismatch*/
+#define ARGUMENTS char *str, size_t size, const char *format, ...
```
nicob87: this will change

jiridanek:
I've been thinking about this some:

Support for RHEL 6 / CentOS 6 by the vendor ends at the end of this year (2020) and it is possible to get a C++11 compatible compiler from Software Collections (there is gcc 4.8, if I recall correctly), even on CentOS 6. Therefore, I'd go for doctest 2.4.0, instead for the C++98 compatible branch.

I'd merge this in two stages. I am unsure about the mocking part, personally. So I'd merge the nonmocking part first, and add the rest later.

I was thinking about merging #294 and I was thinking that it deserves to have a test. Then I realized we still don't have the doctest PR merged.

nicob87:
> I've been thinking about this some:
> 
> Support for RHEL 6 / CentOS 6 by the vendor ends at the end of this year (2020) and it is possible to get a C++11 compatible compiler from Software Collections (there is gcc 4.8, if I recall correctly), even on CentOS 6. Therefore, I'd go for doctest 2.4.0, instead for the C++98 compatible branch.
> 
> I'd merge this in two stages. I am unsure about the mocking part, personally. So I'd merge the nonmocking part first, and add the rest later.
> 
> I was thinking about merging #294 and I was thinking that it deserves to have a test. Then I realized we still don't have the doctest PR merged.

Thanks Jiri, I will submit a pr (or update this one)  containing only the doctest library, and the test without mocks, since I think that is very straightforward.

jiridanek:
@nicob87 awesome, thanks

nicob87:
@jiridanek  updated, If I did no mistake, all "mocking/related" removed.

jiridanek:
I've changed it some more, see my commits

* DISPATCH-1568 Update RuntimeChecks to work with C++ code

Since we are not mocking, the tests can run under any kind of build, including ASan builds. That requires passing the asan compile args to C++ compiler.

* DISPATCH-1568 c-unittests: replace #define with C++ const, use SUBCASEs more

C++ is nice in that e.g. `const int size=5; int array[size]` is allowed. No need for #defines.

* DISPATCH-1568 Add terminus_private.h to make the `static` method testable

@kgiusti had some nice things to say about the _private.h convention, and I think it is a good match in this situation as well. Correct me if I am wrong.

* DISPATCH-1568 Remove qpid-dispatch-testing library (mocking proposal)

Since we aren't mocking (yet), we don't need to build a special version of the library for the tests.

jiridanek:
@kgiusti thank you, LICENSE added

gemmellr:
I think it would have been nice to squash the changes before pushing to the repo, 6 (well, I just made a 7th :P) commits for this relatively straightforward addition seems unecessary and complicates the history later for no real gain, with e.g some changes adding and removed in the same push, and the licence changes seperate from the changes that requires them.

If trying to preserve authorship then just 2 would have at least been simpler. The original commits would still have remained here on the PR if they were simply being 'saved' for future.

jiridanek:
I was thinking about authorship, about meaningful commit message (the squashed commit would have to be, essentially, "DISPATCH-xxx - Change all the things I think should be changed". I couldn't think of a specific commit message), and about `git blame` (if I look at the commit message associated with a changed line, it should explain change on that particular line).

gemmellr:
I find myself thinking of just deleting this instead of posting as it turned out stupidly long, but meh, typed it so...

I get the authorship bit to an extent, though if simply removing/undoing a bunch of stuff from the prior not-yet-pushed PR commit in the next ones I dont think the original really needs to go in the main repo in full as its arguably cluttering history at that point. Not all changes need to be accepted. E.g you could squash explicit removal bits and attribute the net added change to the original person if you arent overly caring that 'these are my changes undoing yours'. Or you could just ask them to undo the changes you think arent required.

Either way the full commits could still have been here to reference without need to complicate history on the master (you could even reference the PR having intermediate history in a squashing commit), which is what I think files with commits of  'adds blah' and then immediately 'removes blah' effect in the same push does. 

E.g to the earlier comment and also your mention of git blame, in this case the git blame will in some cases now be showing you essentially undoing stuff that effectively never existed on master before now, instead of just showing no-change or just what was net added change vs master/past-releases from before this PR. To understand the changes actually made to these parts later based on the blame you might now need to examine at least 2 or possible more updates from separate commits 'for the same change', which might end up explaining that 'nothing much happened here except lots of churn', vs it being direct in the blame showing what did or didnt happen as a unit.

To the description side of things, commit messages can have multiple lines and files can always be commented if its really necessary, plus JIRAs and PRs relating to the given commit(s) can also have explanatory detail in comments, which is a good part of why they exist and are referenced.

I do think having several commits toward an end can make a lot of sense in some cases, e.g when its complicated feature or fix including refactors etc composed of several largely independent changes that sit well on their own (often to the extent theyll get their own JIRAs) rather than bunched together. This just doesnt feel like one of those times to me, and so instead having the various intermediate commits for whats mostly a single effective change just makes it harder to see what the change actually is later, requiring looking at the ensire set.

jiridanek:
Hopefully I won't have to deal with many mixed-authorship PRs. When all the changes are mine, it makes aggressive commit squashing an easier decision.

PR #690, opened 2020-02-19
DISPATCH-1577: open separate inter-router sessions for data and control

kgiusti:


kgiusti:
Abandoned: to be integrated into DISPATCH-1579

PR #691, opened 2020-02-20
DISPATCH-1561: Added a test that makes sure that different modules ca…

ganeshmurthy:
…n write to different log files

### DIFF
--- a/tests/system_tests_log_level_update.py
+++ b/tests/system_tests_log_level_update.py
@@ -82,7 +82,7 @@ class ManyLogFilesTest(TestCase):
         server_log_found = True
         all_server_logs = True
         try:
-            with open('../setUpClass/test-router-server.log', 'r') as server_log:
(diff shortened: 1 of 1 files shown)


tests/system_tests_log_level_update.py
```diff
@@ -29,6 +29,102 @@
...
+
+        TEST_MSG = "LOGTEST"
+        msg = Message(body=TEST_MSG)
+        blocking_sender.send(msg)
+        received_message = blocking_receiver.receive()
+        self.assertEqual(TEST_MSG, received_message.body)
+        server_log_found = True
+        all_server_logs = True
+        try:
+            with open('../setUpClass/test-router-server.log', 'r') as server_log:
```
kgiusti: One totally optional suggestion - I *think* that the Qdrouter class stores the full path to the log files in the ".outdir" member [see the Process class in system_test.py].  Use that to access the logs rather than the hardcoded relative path (less brittle).

PR #692, opened 2020-02-21
DISPATCH-975: Implement max-message-size policy restrictions

ChugR:
* This commit has the basics of the implementation for initial review.
* This commit has tests for the size settings in the schema but no tests
  checking run-time enforcement.

Notes:

* Max message size is enforced on message/transfer ingress only.
  Once a message has entered the router network it is free to go to any
  destination.

* When a message exceeds max size then the ingress link is closed with an error
  Delivery copies of oversize denied messages are aborted.

* In message.c oversize messages are detected and then handled as
  discarded and aborted.

* In router_node.c oversize messages are logged, links are closed,
  and downstream deliveries are disconnected.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -1256,7 +1256,7 @@ qd_message_t *discard_receive(pn_delivery_t *delivery,
         } else if (rc == PN_EOS || rc < 0) {
             // end of message or error. Call the message complete
             msg->content->receive_complete = true;
-            msg->content->aborted = pn_delivery_aborted(delivery);
+            msg->content->aborted = pn_delivery_aborted(delivery) || msg->content->oversize;;
             qd_nullify_safe_ptr(&msg->content->input_link_sp);
 
             pn_record_t *record = pn_delivery_attachments(delivery);
@@ -1301,6 +1301,10 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
         msg->strip_annotations_in  = qd_connection_strip_annotations_in(qdc);
         pn_record_def(record, PN_DELIVERY_CTX, PN_WEAKREF);
         pn_record_set(record, PN_DELIVERY_CTX, (void*) msg);
+        msg->content->max_message_size = qd_connection_max_message_size(qdc);
+    } else {
+        // existing messages may detect oversize transitions only once
+        msg->content->oversize_detected = false;
     }
 
     //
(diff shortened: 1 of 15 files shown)


src/message.c
```diff
@@ -1301,6 +1301,10 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
         msg->strip_annotations_in  = qd_connection_strip_annotations_in(qdc);
         pn_record_def(record, PN_DELIVERY_CTX, PN_WEAKREF);
         pn_record_set(record, PN_DELIVERY_CTX, (void*) msg);
+        msg->content->max_message_size = qd_connection_max_message_size(qdc);
+    } else {
+        // existing messages may detect oversize transitions only once
+        msg->content->oversize_detected = false;
```
ganeshmurthy: Is setting msg->content->oversize_detected  to false necessary ? 
Calling ZERO (msg); already sets this flag to false. Also, once this flag becomes true it *should not* go to false?

src/message.c
```diff
@@ -1409,10 +1413,23 @@ qd_message_t *qd_message_receive(pn_delivery_t *delivery)
...
             //
             qd_buffer_insert(content->pending, rc);
+
+            // Handle maxMessageSize violations
+            if (content->max_message_size) {
+                content->bytes_received += rc;
+                if (content->bytes_received > content->max_message_size)
+                {
+                    content->oversize_detected = true;
+                    content->oversize = true;
```
ganeshmurthy: Can we do away with the content->oversize, just use the content->oversize_detected instead ?

src/router_node.c
```diff
@@ -329,12 +330,32 @@ static bool AMQP_rx_handler(void* context, qd_link_t *link)
     qdr_delivery_t *delivery = qdr_node_delivery_qdr_from_pn(pnd);
     bool       next_delivery = false;
 
-    //
-    // Receive the message into a local representation.
-    //
     qd_message_t   *msg   = qd_message_receive(pnd);
+    
+    bool is_oversize      = qd_message_exceeds_max_message_size(msg);
+    if (is_oversize) {
```
ganeshmurthy: Is there a case when is_oversize is true but qd_message_oversize_detected(msg) is false ?

src/message.c
ChugR: After further consideration here's another strategy:
Instead of two flags use a single counter: Upon going oversize the count will go from zero to one. When the count is zero then the message is not oversize. When the count is one that's when router_node.c takes "just went oversize" action. When the count is greater than one then router_node.c takes no special action.

This would probably be a performance improvement as the counter is modified only when oversize messages are being handled.

src/message_private.h
```diff
@@ -109,6 +109,8 @@ typedef struct {
     qd_parsed_field_t   *ma_pf_to_override;
     qd_parsed_field_t   *ma_pf_trace;
     int                  ma_int_phase;
+    int                  max_message_size;               // configured max; 0 if no max to enforce
+    int                  bytes_received;                 // returned by pn_link_recv()
```
kgiusti: I'd recommend making these both uint64_t.

src/policy.h
```diff
@@ -49,6 +50,7 @@ struct qd_policy__settings_s {
     int  maxSessions;
     int  maxSenders;
     int  maxReceivers;
+    int  maxMessageSize;
```
kgiusti: recommend uint64_t

src/policy.h
```diff
@@ -39,6 +39,7 @@ struct qd_policy_denial_counts_s {
     int sessionDenied;
     int senderDenied;
     int receiverDenied;
+    int maxSizeMessagesDenied;
```
kgiusti: In general don't we use unsigned values for counters elsewhere (like in the core code - all counters are uint64_t)

ChugR:
Review comments and further development to be included in a new, rebased pull request.

PR #693, opened 2020-02-22
DISPATCH-1579: open dedicated inter-router sessions for data and control links

kgiusti:


include/qpid/dispatch/container.h
```diff
@@ -68,6 +68,32 @@ typedef enum {
 } qd_detach_type_t;
 
 
+/**
+ * Session Class
+ *
+ * Used when creating new links from the router.  A connection maintains a set
+ * of sessions over which links can be created.  The session class indicates
```
ganeshmurthy: The session class is used to obtain the appropriate session from the array of sessions that are housed in the connection. The maximum number of sessions in an inter-router connection is QD_SSN_CLASS_COUNT.

src/router_node.c
```diff
@@ -786,18 +786,47 @@ static int AMQP_link_attach_handler(void* context, qd_link_t *link)
...
+    qd_router_t *router  = (qd_router_t*) context;
+    pn_link_t   *pnlink  = qd_link_pn(link);
+    pn_session_t *pn_ssn = pn_link_session(pnlink);
+
+    bool link_flowed = false;
+    if (pn_ssn) {
+        qd_session_t *qd_ssn = pn_session_get_context(pn_ssn);
+        if (qd_ssn && qd_session_is_q3_blocked(qd_ssn)) {
+            // Q3 blocked - have we drained enough outgoing bytes?
+            const size_t q3_lower = BUFFER_SIZE * QD_QLIMIT_Q3_LOWER;
```
ganeshmurthy: Can we please #define this q3_lower somewhere globally so we don't have to calculate it every time

src/router_node.c
```diff
@@ -786,18 +786,47 @@ static int AMQP_link_attach_handler(void* context, qd_link_t *link)
...
+                            link_flowed = true;  // original link flowed - do not flow again
+                    }
+                    blink = DEQ_HEAD(*blinks);
+                }
+            }
+        }
+    }
 
-    qdr_link_flow(router->router_core, rlink, pn_link_remote_credit(pnlink), pn_link_get_drain(pnlink));
+    if (!link_flowed) {
```
ganeshmurthy: Can we please invert the logic here to first send out the flow on the pertinent link and then do the loop after this is done

kgiusti: will do

src/router_node.c
kgiusti: This is tricky - BUFFER_SIZE is actually a variable not a constant.  The unit tests modify it to test various buffer sizes.  I can compute it at startup after the buffer code is done initializing.

PR #694, opened 2020-03-02

ChugR:
This commit has an updated implementation for further review.

Notes:

maxMessageSize enforcement.

size in the _max-message-size_ field of the Attach
frame. Qpid-dispatch ignores the _max-message-size_ field received in
incoming Attach frames.

inserted.

destination.

 * Disposition of _rejected_ is returned to the sender for that delivery.
 * Ingress link to qpid-dispatch is closed with an error.

In message.c oversize messages are detected and then handled as
oversize, discarded, and aborted.

In router_node.c oversize messages are logged, links are closed, and
downstream deliveries are disconnected.

ToDo:

Fix issues in DISPATCH-1581 where policy settings and counters are
_int_ and not _uint64_. That will be a separate work stream.

Fix all resource leaks. Certain patterns of client activity leak
qd_message, qd_link, and qd_disposition resources. Sometimes. More
testing required.

Characterize some issues that AMQP clients suffer when oversize
messages are denied. Clients fail in mysterious ways and some hints on
what is happening and what to do about it will speed client
development.

Improve single-router test to cover multiple routers.

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -27,6 +27,7 @@
 #include "entity_cache.h"
 #include "router_private.h"
 #include "delivery.h"
+#include "policy.h"
 #include <qpid/dispatch/router_core.h>
 #include <qpid/dispatch/proton_utils.h>
 #include <proton/sasl.h>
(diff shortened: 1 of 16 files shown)


src/router_node.c
```diff
@@ -329,12 +330,33 @@ static bool AMQP_rx_handler(void* context, qd_link_t *link)
...
 
-    //
-    // Receive the message into a local representation.
-    //
     qd_message_t   *msg   = qd_message_receive(pnd);
+    
+    int oversize = qd_message_exceeded_max_message_size(msg);
+    if (oversize > 0) {
+        // message is bigger than maxMessageSize
+        if (oversize == 1) {
```
kgiusti: Q: what if the message is large and was forwarded to the core, then the oversize condition is hit on a later transfer?  Won't there be a qdr_delivery_t outstanding on the core's qdr_link_t undelivered list?

src/router_node.c
```diff
@@ -329,12 +330,33 @@ static bool AMQP_rx_handler(void* context, qd_link_t *link)
...
+            // increment policy counters
+            qd_policy_count_max_size_event(pn_link, conn);
+            // reject delivery that went oversize
+            pn_delivery_update(pnd, PN_REJECTED);
+            // Initiate link close with error.
+            pn_condition_t * cond = pn_link_condition(pn_link);
+            (void) pn_condition_set_name(cond, QD_AMQP_COND_MESSAGE_SIZE_EXCEEDED);
+            pn_link_close(pn_link);
+            // Abort outbound deliveries that were receiving this incoming delivery
+            qdr_node_disconnect_deliveries(router->router_core, link, delivery, pnd);
```
kgiusti: If the message has not yet been forwarded to the core, then delivery here will be null.  qdr_node_disconnect_deliveries will crash in that case IIUC.

ChugR:
New PR coming soon

PR #696, opened 2020-03-06
DISPATCH-1586 - Added allocator metrics

fgiorgetti:
Needs to be reviewed and metric must have the correct suffixes, following
prometheus conventions (https://prometheus.io/docs/practices/naming/).

Please review and let me know if the approach below seems valid or not.

### DIFF
--- a/src/http-libwebsockets.c
+++ b/src/http-libwebsockets.c
@@ -460,6 +460,11 @@ typedef struct metric_definition {
     int_metric value;
 } metric_definition;
 
+typedef struct allocator_metric_definition {
+    const char* name;
+    qd_alloc_stats_t *(*fn)(void);
+} allocator_metric_definition;
+
 static int stats_get_connections(qdr_global_stats_t *stats) { return stats->connections; }
 static int stats_get_links(qdr_global_stats_t *stats) { return stats->links; }
 static int stats_get_addrs(qdr_global_stats_t *stats) { return stats->addrs; }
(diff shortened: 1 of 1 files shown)


src/http-libwebsockets.c
```diff
@@ -522,11 +611,35 @@ static bool write_stats(uint8_t **position, const uint8_t * const end, const cha
     }
 }
 
+static bool write_allocator_stats(uint8_t **position, const uint8_t * const end, const char* name, allocator_field field)
+{
+    //18 chars + 2*name + 2*(field+6) + 20 chars for int
+    size_t length = 18 + strlen(name)*2 + strlen(field.name)*2 + 20;
+    if (end - *position >= length) {
+        *position += lws_snprintf((char*) *position, end - *position, "# TYPE %s:%s_bytes gauge\n", name, field.name);
+        *position += lws_snprintf((char*) *position, end - *position, "%s:%s_bytes %lu\n", name, field.name, field.value);
```
jiridanek, line 620:
```suggestion
        *position += lws_snprintf((char*) *position, end - *position, "%s:%s_bytes %" PRIu64 "\n", name, field.name, field.value);
```

jiridanek: Not all platforms define uint64_t as unsigned long int, https://issues.apache.org/jira/browse/DISPATCH-1669

src/http-libwebsockets.c
```diff
@@ -537,7 +650,7 @@ static int callback_metrics(struct lws *wsi, enum lws_callback_reasons reason,
 {
     qd_http_server_t *hs = wsi_server(wsi);
     stats_t *stats = (stats_t*) user;
-    uint8_t buffer[LWS_PRE + 2048];
+    uint8_t buffer[LWS_PRE + 26326];
```
jiridanek, line 653: magic number is magic :D

src/http-libwebsockets.c
fgiorgetti: Ok!

jiridanek:
Can anyone please comment on the "Please review and let me know if the approach below seems valid or not." part? CC @grs

src/http-libwebsockets.c
jiridanek: please explain magic number, consider making it an int constant

grs:
Looks ok to me

src/http-libwebsockets.c
fgiorgetti: @jiridanek finally updated as suggested. Please feel free to review.

src/http-libwebsockets.c
```diff
@@ -573,11 +694,22 @@ static int callback_metrics(struct lws *wsi, enum lws_callback_reasons reason,
...
         }
-        int n = stats->current < metrics_length ? LWS_WRITE_HTTP : LWS_WRITE_HTTP_FINAL;
+
+        int alloc_cur = 0;
+        while (alloc_cur < allocator_metrics_length) {
+            if (write_allocator_metric(&position, end, &allocator_metrics[alloc_cur])) {
+                qd_log(hs->log, QD_LOG_DEBUG, "wrote allocator metric %lu of %lu", alloc_cur, allocator_metrics_length);
+                alloc_cur++;
+            } else {
+                qd_log(hs->log, QD_LOG_DEBUG, "insufficient space in buffer");
```
jiridanek, line 704: I'm wondering about the log levels, `QD_LOG_DEBUG` for success seems high to me (I'd go with TRACE) and `QD_LOG_DEBUG` for the failure here could be too low (I'd give it INFO). But I did not examine qdrouterd logging comprehensively, so maybe I am just no properly attuned.

src/http-libwebsockets.c
```diff
@@ -537,7 +657,8 @@ static int callback_metrics(struct lws *wsi, enum lws_callback_reasons reason,
 {
     qd_http_server_t *hs = wsi_server(wsi);
     stats_t *stats = (stats_t*) user;
-    uint8_t buffer[LWS_PRE + 2048];
+    // rationale for buffer size is explained at write_stats and write_allocator_stats
+    uint8_t buffer[LWS_PRE + 30618];
```
jiridanek, line 661: looks a bit fragile, but ok

jiridanek (approved):
This is tested by the existing test for metrics? Since it is only adding more metrics...

Do the new metrics follow Prometheus guidelines, as you mention in first comment, or that is still something to check?

lgtm, but this is something I have little experience with...

jiridanek:
One more thought. From the README

```
-DQD_MEMORY_STATS=
------------------

Dispatch will track memory pool usage statistics if this is enabled.```

So the statistics may not be always available, depending on the way Dispatch is compiled. Can that break your code, or is that then always going to report 0?

File: src/http-libwebsockets.c (hunk 3 above)
Comment (fgiorgetti) [reply src/http-libwebsockets.c:704]: Makes sense to me as well.

**Comment by fgiorgetti** [comment]
> One more thought. From the README
> 
> ```
> -DQD_MEMORY_STATS=
> ------------------
> 
> Dispatch will track memory pool usage statistics if this is enabled.
> ```
> 
> So the statistics may not be always available, depending on the way Dispatch is compiled. Can that break your code, or is that then always going to report 0?

Allocator metrics get reported as 0. I see that `qdmanage` works just fine, but `qdstat` is throwing an error in such case:

```
$ qdstat -m
2021-02-04 15:08:19.288818 UTC
Standalone_v++jCPCjYiQcGo3

TypeError: unsupported operand type(s) for *: 'int32' and 'NoneType'
```

ganeshmurthy:
Please do not merge this PR until after we start the vote for 1.15

jiridanek:
I'll merge and I'll create a Jira for broken qdstat

grs:
Personally I think having a broken qdstat on master is something to avoid. We also need to make sure that we still have compatability with older qdstat (and vice versa).

In my opinion it is better to get a solution to any consequent problems like this *before* merging.

jiridanek:
> In my opinion it is better to get a solution to any consequent problems like this _before_ merging.

My understanding was that qdstat was already broken on master (when compiled with that nondefault option) and the exception is not caused by this PR. Cc @fgiorgetti can you clarify?

jiridanek:
@grs I checked, the bug DISPATCH-1958 was already present, before this PR got merged. I tested at commit e78a18ae.

grs:
@jiridanek sorry, I misunderstood. If the matser is already broken no issue with the merge.

PR #698, opened 2020-03-06

ChugR:
This commit has an updated implementation for final review.

maxMessageSize enforcement.

size in the _max-message-size_ field of the Attach
frame. Qpid-dispatch ignores the _max-message-size_ field received in
incoming Attach frames.

inserted.

destination.

 * Disposition of _rejected_ is returned to the sender for that delivery.
 * Previous versions of this patch closed the sender's ingress link
   with an error. This ensured that a sender would know that the
   message did not get through and would clearly know why. However,
   spontaneous link closures cause a wide variety of serious problems
   for many clients. Link closure should not be enforced until, at a
   minimum, qpid-proton clients handle the closure cleanly.

edge routers. Tests try oversize messages with a variety of sender and

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -27,6 +27,7 @@
 #include "entity_cache.h"
 #include "router_private.h"
 #include "delivery.h"
+#include "policy.h"
 #include <qpid/dispatch/router_core.h>
 #include <qpid/dispatch/proton_utils.h>
 #include <proton/sasl.h>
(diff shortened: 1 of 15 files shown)


src/router_node.c
```diff
@@ -1884,3 +1885,12 @@ void qd_link_restart_rx(qd_link_t *in_link)
         qd_connection_invoke_deferred(in_conn, deferred_AMQP_rx_handler, safe_ptr);
     }
 }
+
+
+// prepend formatted connection and link info to policy denial text and log it
+void qd_connection_log_policy_denial(qd_link_t *link, const char *text)
+{
+    qdr_link_t *rlink = (qdr_link_t*) qd_link_get_context(link);
```
kgiusti: The qdr_link_t has a separate lifecycle from the qd_link_t.  IOW you'll want to check that rlink is not null here.

ChugR:
Closing this PR. Upcoming replacement includes review comment fixes, improved testing, and a rebase to latest master.

PR #702, opened 2020-03-12
NO-JIRA Improve printing of results in travis.yml (folded sections feature)

jiridanek:
* Print list files with unapproved license if any
* Fold output from `ctest`, print list of failures at the end

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,12 +36,14 @@ matrix:
     cache:
       directories:
         - $HOME/.m2/repository
-      before_cache:
-        - rm -rf $HOME/.m2/repository/org/apache/qpid/*
+    before_cache:
+      - rm -rf $HOME/.m2/repository/org/apache/qpid/*
     install: []
     before_script: []
     script:
     - mvn apache-rat:check
+    after_failure:
+    - cat target/rat.txt
   # prepending /usr/bin to PATH to avoid mismatched python interpreters in /opt
   - os: linux
     env:
@@ -114,7 +116,11 @@ script:
 - popd
 # Workaround on macOS for PROTON-808 Binaries have their library locations stripped
 - if [[ "${OSTYPE}" == "darwin"* ]]; then install_name_tool -add_rpath $PREFIX/lib/. $PREFIX/lib/proton/bindings/python/_cproton.so; fi
+# Undocumented Travis feature https://github.com/travis-ci/docs-travis-ci-com/issues/949 Can I add new log fold sections?
+- travis_fold start ctest  # [ctest
 - ctest -V && if [ "$BUILD_TYPE" = "Coverage" ]; then cmake --build . --target coverage; fi
+- travis_fold end ctest  # ctest]
+- FAILLOG=Testing/Temporary/LastTestsFailed.log; if [[ -e $FAILLOG ]]; then cat $FAILLOG && exit 1; else echo "Success!"; fi
 
 after_success:
 - cd ${TRAVIS_BUILD_DIR}/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,7 +120,7 @@ script:
 - travis_fold start ctest  # [ctest
 - ctest -V && if [ "$BUILD_TYPE" = "Coverage" ]; then cmake --build . --target coverage; fi
 - travis_fold end ctest  # ctest]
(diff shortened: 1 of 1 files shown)


jiridanek:
Run with a failure then looks like this
* https://travis-ci.org/github/jiridanek/qpid-dispatch/jobs/661558177

Run without any fails looks like this
* https://travis-ci.org/github/jiridanek/qpid-dispatch/jobs/661558175

IMO this is change for the better.

.travis.yml
```diff
@@ -36,12 +36,14 @@ matrix:
...
-      before_cache:
-        - rm -rf $HOME/.m2/repository/org/apache/qpid/*
+    before_cache:
+      - rm -rf $HOME/.m2/repository/org/apache/qpid/*
     install: []
     before_script: []
     script:
     - mvn apache-rat:check
+    after_failure:
+    - cat target/rat.txt
```
gemmellr: Depending on the failure, the file might not exist, worth guarding?

gemmellr:
If I'm looking at the job log its typically going to involve looking at that output, so I dont especially like it being collapsed by default personally. Being able to collapse it might be useful, but it being so by default less so.

Adding the additional output does seem useful regardless though.

gemmellr:
Well, the rat output one does. The other I personally find less obvious than the existing ctest failure list .which is more obvious in the output (to me anyway).

.travis.yml
jiridanek: If it does not exist, a nonzero return value from the last after_failure command does not really change anything. To some degree, missing target/rat.txt can be considered an additional failure, because of the expected way the rat CI job is supposed to fail.

I can add the check, it would make the job then future-proof against adding additional commands to after_failure.

jiridanek:
I cannot add a foldable section and let it be unfolded by default. The reason for adding the new ctest failure list afterwards is because I needed to show the list of failing tests even in the default folded state.

One thing I wanted but did not find was to only show output from the failed ctest targets by default, and allow to unfold the rest. That does not seem to be easily available.

I agree that when the failures are colored red, which the ctest stdout/stderr output does, it is more obvious. On the other hand, the way this PR does it, the output is much more friendly for a newcommer, I'd argue. It is easier to wrap ones head around a screenful of output, and then possibly start expanding sections, than to see thousands of lines by default from the start.

gemmellr:
I dont find it makes much difference other than requiring another step to see the output I am most likely there to look at. Even with it collapsed theres still enough output to flow off most screens. The initial 'result' of interest is usually at the bottom of a CI log so I tend to open the page and either hit end or use the UI to skip to the bottom, somthing thats still necessary. I dont think there are many people looking at the CI run output that havent already or arent about to run the tests themselves and see all the output.

The new failures output could at least be clearer if it printed an indication that it was emitting the failing tests, e.g the way the ctest output does. The highlighted red text not being the ctest command but the less obvious 'repeat failures output' command would still distract, but at least it would be far more obvious that the lines right above it are its interesting output signalling the failures.

jiridanek:
Thank you for the review. The ctest folding trully does not seem to be as good as I was hoping it might be. I don't have additional ideas how to improve it, except maybe hiding output from passing tests by default instead, and I don't know how to easily achieve that in ctest (so that I dont have any duplication in the fully expanded log).

One idea I had was to run ctest second time, only the tests that failed in the first run, assuming those were numbered less than say 5... That way, there would be less scrolling back up, and it would provide more clues about flakes.


I'll split out the rat change into separate PR and close this one PR.

.travis.yml
gemmellr: I wouldn't really see it as an 'additional failure', just a consequence of the actual failure, and so it seemed like any output reported for this failing is just distracting from seeing the actual failure. But I guess its not going to happen often and the output is fairly obvious, and unlike other jobs theres no additional checks here besides rat, so maybe its not as important as when it were part of the overall test run like before.

jiridanek: Gonna go with `|| true` there.

gemmellr:
As I say, I think the new output could easily be improved, to be more similar to the actual ctest ending summary and make it more obvious that it is reporting the failures. I dont think duplicating the 'these tests failed' summary is much of an issue, especially if one is collapsed normally. For me its just that collapsing the ctest output by default isnt an improvement since its usually what folks will be there to see.

I'd leave this PR open for now though, others may feel differently.

jiridanek:
The GitHub Action currently gives sufficient RAT output to figure out why the check failed. This PR does not look necessary any more, at least not in this form.

PR #705, opened 2020-03-13
DISPATCH-975: Enforce max message size on message ingress v3

ChugR:
This update incorporates review comments and a number of new features:

* All pn_delivery work is moved out of message.c and into router_node.c
* log_link_message identifies oversize messages
* Oversize message handling in router_node.c is separated into an
  isolated section of code making it easier to reason about how
  oversize messages are handled.
* Self tests are improved
** Each router has a different maxMessagesize
** Sixteen tests connect to each of four routers with senders and
   receivers.
** Oversize and Undersize messages are checked to prove that routers
   with smaller limits will successfully pass through larger messages.
** Interior routers with smaller limits than attached edge routers
   will block messages that the edge routers allow but are over the
   interior router limit.

TODO:

* This version still leaks resource (iterator, buffer, parsed_field,
message, message_content, and delivery) objects under some conditions.
* No tests yet for multicast or link route.

================================================================
Comments from previous pull requests
================================================================
This commit has an updated implementation for review (v2 PR#698)

maxMessageSize enforcement.

size in the max-message-size field of the Attach
frame. Qpid-dispatch ignores the max-message-size field received in
incoming Attach frames.

inserted.

destination.

 *  Disposition of rejected is returned to the sender for that delivery.
 *  Copies of the message being delivered through the router network are aborted.
 *  Previous versions of this patch closed the sender's ingress link
    with an error. This ensured that a sender would know that the
    message did not get through and would clearly know why. However,
    spontaneous link closures cause a wide variety of serious problems
    for many clients. Link closure should not be enforced until, at a
    minimum, qpid-proton clients handle the closure cleanly.

edge routers. Tests try oversize and undersize messages with a variety of sender and

### DIFF
--- a/include/qpid/dispatch/container.h
+++ b/include/qpid/dispatch/container.h
@@ -238,6 +238,8 @@ void qd_session_free(qd_session_t *qd_ssn);
 bool qd_session_is_q3_blocked(const qd_session_t *qd_ssn);
 qd_link_list_t *qd_session_q3_blocked_links(qd_session_t *qd_ssn);
 
+void qd_connection_log_policy_denial(qd_link_t *link, const char *text);
+
 
 // handy macros to get around PROTON-2184: pn_session_set_context aborts if
 // context==0  (can remove this once qdrouter requires >= proton 0.31.x)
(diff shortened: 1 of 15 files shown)


ChugR:
Updated PR coming

PR #708, opened 2020-03-23
DISPATCH-1490 - remove old-user-guide dir

bhardesty:


docs/books/CMakeLists.txt
```diff
@@ -34,16 +33,10 @@ if (ASCIIDOCTOR_EXE)
     execute_process (
       COMMAND ${CMAKE_COMMAND} -E copy_if_different
       ${CMAKE_CURRENT_SOURCE_DIR}/${file}
-      ${CMAKE_CURRENT_BINARY_DIR}/${file}
+      ${CMAKE_CURRENT_BINARY_DIR}/user-guide/${file}
```
bhardesty: @ganeshmurthy can you please verify this? Currently, if you run "make docs", the images in user-guide/index.html are all broken. This is because the images/ dir is created one level up from where index.html expects it to be. This just ensures that the images/ dir is created where index.html expects.

With this change, when I run "make docs", the images are in the correct location and they all show up in the doc as they should. However, I'm not sure if this change affects any of the make tooling for updating the website and/or creating new qpid-dispatch releases.

PR #711, opened 2020-03-27
Added debug-dump backtrace for leaked objects in pooled memory.

ted-ross:
If QD_MEMORY_DEBUG is defined, this stores a backtrace for each allocated pool object so that objects left unfreed can be better described at shutdown.

This is a fairly expensive feature.  We may want to consider providing an extra definable option so we can use QD_MEMORY_DEBUG without using this new feature.

### DIFF
--- a/src/alloc_pool.c
+++ b/src/alloc_pool.c
@@ -547,10 +547,11 @@ void qd_alloc_finalize(void)
             qd_alloc_item_t *item = DEQ_HEAD(qtype->allocated);
             while (item) {
                 size_t i;
-                char   **strings;
-                strings = backtrace_symbols (item->backtrace, item->backtrace_size);
+                char **strings;
+                strings = backtrace_symbols(item->backtrace, item->backtrace_size);
                 for (i = 0; i < item->backtrace_size; i++)
-                    fprintf (dump_file, "%s\n", strings[i]);
+                    fprintf(dump_file, "%s\n", strings[i]);
+                free(strings);
                 fprintf(dump_file, "\n");
                 item = DEQ_NEXT(item);
             }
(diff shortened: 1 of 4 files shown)


src/alloc_pool.c
```diff
@@ -501,11 +537,25 @@ void qd_alloc_finalize(void)
...
                     "alloc.c: Items of type '%s' remain allocated at shutdown: %"PRId64"\n",
                     desc->type_name,
                     desc->stats->total_alloc_from_heap - desc->stats->total_free_to_heap);
+#ifdef QD_MEMORY_DEBUG
+            qd_alloc_type_t *qtype = (qd_alloc_type_t*) desc->debug;
+            qd_alloc_item_t *item = DEQ_HEAD(qtype->allocated);
+            while (item) {
+                size_t i;
+                char   **strings;
+                strings = backtrace_symbols (item->backtrace, item->backtrace_size);
```
kgiusti: IIUC *strings needs to be freed manually

ted-ross: That would be good to know.  I'll check that.

ChugR:
It would also be valuable to have a timestamp on the allocation. Then it could be correlated with the log file accurately.

ganeshmurthy:
Please correct me if I am wrong here.

If I set the debugDumpFile attribute of router entity and have QD_MEMORY_STATS as ON and run the system_tests_one_router, I see the following on stdout

20: Router QDR debug dump file:
20: alloc.c: Items of type 'qd_policy_settings_t' remain allocated at shutdown: 14

It would be useful for me to see the alloaction backtrace of all the above 14 qd_policy_settings_t in the debugDumpFile, right ?

PR #717, opened 2020-04-08

ChugR:
maxMessageSize enforcement.

size in the max-message-size field of the Attach frame.

Qpid-dispatch ignores the max-message-size field received in
incoming Attach frames.

inserted.

Max message size is enforced on message/transfer ingress only. Interior
routers form a trusted domain where messages that have entered by one
interior router are allowed to any destination in the system. Edge
routers are not trusted; messages from an edge router must be checked
again by the attached interior router before they are forwarded to the
trusted network.

 * Transfer size is checked as a message enters an interior router.

A message that passes an interior router size check is free to go to any
destination on that router or any other interior or edge router.

 * Transfers size is checked as a message enters an edge router.

A message that passes an edge router size check is free to go to any
destination on that edge router. If the message is uplinked to an
interior router then that interior router may apply another max message
size check possibly using a different maximum message size setting.

 * Disposition of rejected is returned to the sender for that delivery.
 * The connection to the sender is closed with condition:

   "amqp:connection:message-size-exceeded"

edge routers. Tests try oversize and undersize messages with a variety of sender and

TODO:

* No tests yet for multicast or link route.

### DIFF
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -86,6 +86,7 @@ const char * const QD_AMQP_COND_RESOURCE_DELETED = "amqp:resource-deleted";
 const char * const QD_AMQP_COND_ILLEGAL_STATE = "amqp:illegal-state";
 const char * const QD_AMQP_COND_FRAME_SIZE_TOO_SMALL = "amqp:frame-size-too-small";
 const char * const QD_AMQP_COND_CONNECTION_FORCED = "amqp:connection:forced";
+const char * const QD_AMQP_COND_MESSAGE_SIZE_EXCEEDED = "amqp:connection:message-size-exceeded";
 
 const char * const QD_AMQP_PORT_STR = "5672";
 const char * const QD_AMQPS_PORT_STR = "5671";
--- a/include/qpid/dispatch/amqp.h
+++ b/include/qpid/dispatch/amqp.h
@@ -201,6 +201,7 @@ extern const char * const QD_AMQP_COND_PRECONDITION_FAILED;
 extern const char * const QD_AMQP_COND_RESOURCE_DELETED;
 extern const char * const QD_AMQP_COND_ILLEGAL_STATE;
 extern const char * const QD_AMQP_COND_FRAME_SIZE_TOO_SMALL;
+extern const char * const QD_AMQP_COND_MESSAGE_SIZE_EXCEEDED;
 
 extern const char * const QD_AMQP_COND_CONNECTION_FORCED;
 /// @};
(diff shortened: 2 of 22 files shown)


src/amqp.c
```diff
@@ -86,6 +86,7 @@ const char * const QD_AMQP_COND_RESOURCE_DELETED = "amqp:resource-deleted";
 const char * const QD_AMQP_COND_ILLEGAL_STATE = "amqp:illegal-state";
 const char * const QD_AMQP_COND_FRAME_SIZE_TOO_SMALL = "amqp:frame-size-too-small";
 const char * const QD_AMQP_COND_CONNECTION_FORCED = "amqp:connection:forced";
+const char * const QD_AMQP_COND_MESSAGE_SIZE_EXCEEDED = "amqp:connection:message-size-exceeded";
```
kgiusti, line 89: Not sure if this matters, but AMQP 1.0 defines a link-error for message-size-exceeded, not a connection-error:

[AMQP 1.0 link-error](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-link-error)

ChugR: Good point. Change the error to
    condition name: amqp:connection:forced
    condition description: message size exceeded

ChugR:
Fixed at commit a754e18

src/amqp.c
gemmellr: Why close the connection rather than the link?

ChugR: Earlier versions of this feature tried closing the link with an error. This did not work well with any of the proton clients, in particular the python client used by dispatch self tests. 

See https://issues.apache.org/jira/browse/PROTON-2182

ChugR: My tests tried only the proton-c clients. I didn't try any proton-j clients. Sorry for the mischaracterization of "any of the proton clients".

gemmellr: To be clearer, I was talking about the sending link, vs sending connection. In this case the example having both the sender and receiver of the message on the same connection probably confuses matters a touch.

From the error and stacktrace that appears to be more on the receiver side. Perhaps a general issue with handling aborted deliveries? Unclear it relates to choice of whether to close the sending link vs connection, other than the fact both are on the same connection.

gemmellr: Its arguably not that big a deal to the sending client, as anyone sending over the limit is going to have issues either way.

I was more curious given the commentry seems to suggest inter-router connections could be getting killed too in certain cases, which could affect other things besides the one doing the send? Or is that not the case? Though I guess that effect may still be true of link closure too.

PR #719, opened 2020-04-15
DISPATCH-1622: Called validate_add() before calling entity.create() w…

ganeshmurthy:
…hich prevents entity.create() being called if the validate_add() fails. Also added a null check in qd_bitmask_first_set().


ganeshmurthy:
Additional tests for this PR will be coming as part of another PR DISPATCH-1617

PR #721, opened 2020-04-17
Debug CI system that never returns on_sendable - do not merge

ChugR:
test only

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -327,6 +327,10 @@ static bool AMQP_rx_handler(void* context, qd_link_t *link)
     if (!pnd)
         return false;
 
+    if (pn_link_state(pn_link) & PN_LOCAL_CLOSED) {
+        return false;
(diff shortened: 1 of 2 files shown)


gemmellr:
You dont need to open pull requests for doing test runs like this, you can log in to Travis with github and enable Travis on your repo fork, then it will just run the same job whenever you push to your fork. It will typically do so more quickly too, as it wont be using the specific executor resources shared within the Apache org, but rather the much larger Travis-wide 'open source' resources pool.

ChugR:
Thanks Robbie. I set up the chain to build directly and the test passes on travis-ci. Master failed on the last run and I sure wish my debug tracing was turned on so I could see more info than just "your test timed out".

My real issue is with some rhel8 system somewhere where master never passes. I was hoping that the branch would get pulled in and tested but no luck so far.

PR #725, opened 2020-04-23
DISPATCH-1626 - Modified AMQP_rx_handler to call pn_delivery_update()…

ganeshmurthy:
… only for aborted deliveries. This ensures that only one dispostion is sent for a pn_delivery"

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -347,20 +347,6 @@ static bool AMQP_rx_handler(void* context, qd_link_t *link)
             //
             pn_link_advance(pn_link);
             next_delivery = pn_link_current(pn_link) != 0;
-
-            uint64_t local_disp = qdr_delivery_disposition(delivery);
-            //
-            // Call pn_delivery_update only if the local disposition is different than the pn_delivery's local disposition.
-            // This will make sure we call pn_delivery_update only when necessary.
-            //
-            if (local_disp != 0 && local_disp != pn_delivery_local_state(pnd)) {
-                //
-                // DISPATCH-1626 - This enables pn_delivery_update() and pn_delivery_settle() to be called back to back in the same function call.
-                // CORE_delivery_update() will handle most of the other cases where we need to call pn_delivery_update() followed by pn_delivery_settle().
-                //
-                if (qd_message_is_discard(msg))
-                    pn_delivery_update(pnd, local_disp);
-            }
         }
 
         if (qd_message_is_discard(msg)) {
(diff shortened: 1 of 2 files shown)


src/router_node.c
```diff
@@ -1829,11 +1838,16 @@ static void CORE_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t di
         // If the delivery is still arriving, don't push out the disposition change yet.
         //
         if (qd_message_receive_complete(msg)) {
-            pn_delivery_update(pnd, disp);
+            if (disp != pn_delivery_local_state(pnd)) {
+                pn_delivery_update(pnd, disp);
+                if (qdr_delivery_disposition(dlv) != disp)
```
kgiusti, line 1843: This should not be necessary (and it's a race): The disp passed into this function comes from the qdr_delivery_t so this check and set is a no-op.  I would remove these.

src/router_node.c
```diff
@@ -1829,11 +1838,16 @@ static void CORE_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t di
...
+                    qdr_delivery_set_disposition(dlv, disp);
+            }
         } else {
             // just update the local disposition for now - AMQP_rx_handler will
             // write this to proton once the message is fully received.
-            qdr_delivery_set_disposition(dlv, disp);
+            if (qdr_delivery_disposition(dlv) != disp)
```
kgiusti, line 1849: Same here - this should not be necessary.  You can replace these with an assert(qdr_delivery_disposition(dlv) == disp) to ensure this routine isn't called inappropriately just to be safe.

src/router_node.c
```diff
@@ -349,8 +349,17 @@ static bool AMQP_rx_handler(void* context, qd_link_t *link)
...
+            //
+            // Call pn_delivery_update only if the local disposition is different than the pn_delivery's local disposition.
+            // This will make sure we call pn_delivery_update only when necessary.
+            //
+            if (local_disp != 0 && local_disp != pn_delivery_local_state(pnd)) {
+                //
+                // DISPATCH-1626 - This enables pn_delivery_update() and pn_delivery_settle() to be called back to back in the same function call.
+                // CORE_delivery_update() will handle most of the other cases where we need to call pn_delivery_update() followed by pn_delivery_settle().
+                //
+                if (qd_message_is_discard(msg))
```
kgiusti, line 361: If the only point where the pn_delivery_update() needs to be done is if qd_message_is_discard(), then why not move all this dispo logic into the qd_message_is_discard(msg) logic right before the qd_delivery_settle() is done (linke #383)

PR #730, opened 2020-04-30
DISPATCH-1636: extract the peer router's version from the Open frame

kgiusti:


kgiusti:
I'll add a couple of version test comparison macros && tests...

tests/system_tests_edge_router.py
```diff
@@ -1640,6 +1640,51 @@ def _test_traffic(self, sender, receiver, address, count=5):
             tr.queue.get(timeout=TIMEOUT)
         tr.stop()
 
+    def test_000_check_peer_version_info(self):
+        """
+        Not a link proxy test - ensure router correctly parses peer version
+        numbers advertised in the incoming @open frame
+        """
+        lines = None
+        with open("../setUpClass/INT.A.log") as inta_log:
```
ganeshmurthy, line 1649: Can we use with open(self.router.outdir + '/INT.A.log, 'r') as inta_log:

include/qpid/dispatch/router_core.h
```diff
@@ -851,6 +851,30 @@ void qdr_query_free(qdr_query_t *query);
 typedef void (*qdr_manage_response_t) (void *context, const qd_amqp_error_t *status, bool more);
 void qdr_manage_handler(qdr_core_t *core, qdr_manage_response_t response_handler);
 
+typedef struct {
+    uint16_t major;
+    uint16_t minor;
+    uint16_t patch;
+    uint16_t flags;
+#define QDR_ROUTER_VERSION_SNAPSHOT 0x0100
+#define QDR_ROUTER_VERSION_RC       0x0200  // lower byte == RC #
```
ganeshmurthy, line 860: Are these used in the code anywhere?

tests/system_tests_edge_router.py
kgiusti: Ah - good catch, thanks!  (fixed)

include/qpid/dispatch/router_core.h
kgiusti: Nope - not yet.  Reserved.  Since we _do_ use a snapshot (dev builds), I wanted to add an additional flag to the version, which - since it's 2 bytes - aligns the version structure to a nice round 8 bytes and gives us a few extra bits to support RC numbers if we ever choose to do so in the future.

PR #731, opened 2020-05-05
DISPATCH-781 - Work in Progress - Do not merge.

ted-ross:
Updated to delay the activation of addresses for routing until credit has been issued from receivers.  Previously, the ATTACH of the outgoing link triggered routing for an address.  With this change, FLOW is the trigger.
Attached links that have not yet been issued credit are stored in the new addr->pending_rlinks list.
This is a building block for the complete end-to-end flow-control solution for DISPATCH-781.


ted-ross:
Note that there is a new test for end-to-end flow control that fails with this patch.  The test won't pass until the whole solution is completed.

ted-ross:
> One suggestion: The logic that adds/removes the link_ref to the qdr_address_t (QDR_LINK_CLASS_ADDRESS) by checking the initial_credit_received value is repeated across a couple of files here.
> 
> I'd suggest moving that logic into its own two functions and calling them instead.

Thanks.  I'll look into making this change.

PR #732, opened 2020-05-06
DISPATCH-1637 - Added a hashtable to store names of link routes, auto…

ganeshmurthy:
… links and address configs

### DIFF
--- a/src/router_core/route_control.c
+++ b/src/router_core/route_control.c
@@ -378,9 +378,22 @@ qdr_link_route_t *qdr_route_add_link_route_CT(qdr_core_t             *core,
     }
 
     //
+    // If a name was provided, use that as the key to insert the this link route name into the hashtable so
+    // we can quickly find it later.
+    //
+    if (name) {
+        qd_iterator_view_t iter_view = qd_iterator_get_view(name);
+        qd_iterator_reset_view(name, ITER_VIEW_ADDRESS_HASH);
+        qd_hash_insert(core->addr_lr_al_hash, name, lr, &lr->hash_handle);
+        qd_iterator_reset_view(name, iter_view);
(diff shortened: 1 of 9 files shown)


src/router_core/route_control.c
```diff
@@ -441,6 +454,9 @@ void qdr_route_del_link_route_CT(qdr_core_t *core, qdr_link_route_t *lr)
         }
     }
 
+    if (lr->hash_handle)
+        qd_hash_remove_by_handle(core->addr_lr_al_hash, lr->hash_handle);
```
kgiusti, line 458: Do we need to keep the hash handle around after this?
If not consider freeing it now and zero the lr->hash_handle.

src/router_core/route_control.c
```diff
@@ -543,6 +570,9 @@ void qdr_route_del_auto_link_CT(qdr_core_t *core, qdr_auto_link_t *al)
         }
     }
 
+    if (al->hash_handle)
+        qd_hash_remove_by_handle(core->addr_lr_al_hash, al->hash_handle);
```
kgiusti, line 574: ditto

src/router_core/router_core.c
```diff
@@ -678,6 +684,8 @@ void qdr_core_remove_address_config(qdr_core_t *core, qdr_address_config_t *addr
     qd_iterator_t *pattern = qd_iterator_string(addr->pattern, ITER_VIEW_ALL);
 
     // Remove the address from the list and the parse tree
+    if (addr->hash_handle)
+        qd_hash_remove_by_handle(core->addr_lr_al_hash, addr->hash_handle);
```
kgiusti, line 688: ditto

PR #738, opened 2020-05-15
DISPATCH-1614 - Raised and handled QDRC_EVENT_LINK_IN_DETACHED and QD…

ganeshmurthy:
…RC_EVENT_LINK_OUT_DETACHED events where the addr's edge_inlink and edge_outlink can be zeroed out

### DIFF
--- a/src/router_core/modules/edge_router/addr_proxy.c
+++ b/src/router_core/modules/edge_router/addr_proxy.c
@@ -229,7 +229,7 @@ static void on_link_event(void *context, qdrc_event_t event, qdr_link_t *link)
     switch (event) {
         case QDRC_EVENT_LINK_OUT_DETACHED: {
             qdr_address_t *addr = link->owning_addr;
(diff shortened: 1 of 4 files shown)


src/router_core/modules/edge_router/addr_proxy.c
```diff
@@ -215,6 +215,50 @@ static void del_outlink(qcm_edge_addr_proxy_t *ap, qdr_address_t *addr)
...
+    //
+    // We only care if the link event is on an edge connection.
+    //
+    if (link->conn->role != QDR_ROLE_EDGE_CONNECTION)
+            return;
+
+    switch (event) {
+        case QDRC_EVENT_LINK_OUT_DETACHED: {
+            qdr_address_t *addr = link->owning_addr;
+            if (addr && addr->edge_outlink && link == addr->edge_outlink) {
```
ted-ross, line 232: The second condition in the if is technically not necessary.  edge_outlink is never dereferenced, only compared.

PR #748, opened 2020-05-26
DISPATCH-1646 - Additional fix. Set USE_LIBWEBSOCKETS only if require…

ganeshmurthy:
…d version is found


kgiusti:
Take a look at [this](https://gist.github.com/kgiusti/b984e742d7e5056efeb5ca7664c071d7)
The "find_package" command allows a minimum version to be given.  Our hacked up FindLibWebSockets.cmake needs to handle that.
In addition cmake provides a macro for making a build flag (USE_LIBWEBSOCKETS) conditionally available if another flag (LIBWEBSOCKETS_FOUND) is true.

I think that's the cmake-y way of doing this....

PR #749, opened 2020-05-27
DISPATCH-1642: Change parse tree insert to fail on duplicate entry

kgiusti:
This avoids having to look up the pattern first in order to avoid duplication (which is expensive as it searches the entire tree).


kgiusti:
Merged:
https://gitbox.apache.org/repos/asf?p=qpid-dispatch.git;h=ee82acd

PR #755, opened 2020-06-08
Dispatch 1682: Augment parse tree lookup with a hash table

kgiusti:


include/qpid/dispatch/iterator.h
```diff
@@ -329,6 +329,11 @@ int qd_iterator_ncopy(qd_iterator_t *iter, unsigned char* buffer, int n);
  */
 unsigned char *qd_iterator_copy(qd_iterator_t *iter);
 
+/**
```
ted-ross, line 332: Is there any reason that the const version shouldn't be the only version?  i.e. add const to the argument of the original function?

kgiusti: The original is documented as advancing the iterator past the data copied out.  I'm assuming that some codepath relies on that behavior. 

I'll need to audit the existing use cases before we can be sure we can safely make that change.

ted-ross: In that case, don't make the change.

kgiusti:
**Memory Impact**
At 40K configured addresses the memory footprint as given via qdstat -m:

**master (pre patch)**

  VmSize    Pooled
  464 MiB   66.7 MiB

**DISPATCH-1682**

 VmSize     Pooled
  478 MiB   69.0 MiB

478 - 464 = +14MiB
14/40K = ~+350 bytes per configured address

tests/hash_test.c
```diff
@@ -0,0 +1,257 @@
...
+    return 0;
+}
+
+
+// test lookup and remove failures using strings
+//
+static char *test_str_bad(void *context)
+{
+
+    qd_hash_handle_t *handles[key_count];
```
jiridanek, line 200: This does not compile with clang 11, https://travis-ci.org/github/apache/qpid-dispatch/jobs/696018472#L1704

```/Users/travis/build/apache/qpid-dispatch/tests/hash_test.c:249:31: error: 
      variable length array folded to constant array as an extension
      [-Werror,-Wgnu-folding-constant]
    qd_hash_handle_t *handles[key_count];
                              ^~~~~~~~~```

PR #757, opened 2020-06-09
DISPATCH-1685: add heap allocation API that aborts on failure

kgiusti:


ganeshmurthy:
Should the existing code be modified to use these new functions or is that another commit ?

kgiusti:
> Should the existing code be modified to use these new functions or is that another commit ?

I'm not planning any further commits at present.   I just wanted to propose the new feature and get feedback on the idea of abort() on malloc failure in the router.
If the consensus is that this is A Good Idea then we can either phase these in over time (easy) or big bang (huge amount of work).  I'll leave that on the table for further discussion.

include/qpid/dispatch/ctools.h
```diff
@@ -217,4 +218,41 @@ do {                                    \
...
 
+//
+// Heap allocation with abort() on failure
+//
+
+static inline void *qd_malloc(size_t size)
+{
+    void *ptr = malloc(size);
+    if (!ptr)
+        abort();
```
jiridanek, line 229: maybe some error message for the abort? on the other hand, that abort will probably never be called in practice, so it does not matter for UX...

include/qpid/dispatch/ctools.h
```diff
@@ -217,4 +218,41 @@ do {                                    \
 #define MIN(a,b) (((a)<(b))?(a):(b))
 #define MAX(a,b) (((a)>(b))?(a):(b))
 
+//
+// Heap allocation with abort() on failure
+//
+
+static inline void *qd_malloc(size_t size)
```
jiridanek, line 225: can it ever happen in dispatch that size==0 in calling this?

include/qpid/dispatch/ctools.h
kgiusti: I was thinking about that, but decided that if a hard memory limit is hit it is unlikely that doing any I/O would succeed.

But at this point it probably cannot hurt to call perror before aborting...  at best it will indicate the failure was due to low memory.   I'll update the patch.

include/qpid/dispatch/ctools.h
kgiusti: Ah - good point.   I'm pretty sure there's no zero size use cases, but I'll add an assert() to enforce that contract.
If size == 0 the caller shouldn't make the call at all, or fall back to using malloc directly if they're doing something... weird.

kgiusti:
In general, malloc et. al. will fail (return null) in the case where the process hits RLIMIT_AS/DATA.  I believe it can also happen when running in a container that has a total memory limit configured.

Otherwise linux will return a non-zero value even if there is _no memory currently available_.  When the pointer is de-referenced the process hits a page fault and is suspended until more memory becomes available (or it is OOM killed).  

So if abort() occurs it is a hint to a Carbon-Based Lifeform that the provisioned memory is insufficient for the use case.

Or there's a bug.... there's always bugs.....

PR #760, opened 2020-06-11
DISPATCH-1679: Add sys_thread_self()

kgiusti:
sys_thread_self() provides an interface for accessing the sys_thread_t
pointer for the currently running thread.  This replaces the old
sys_thread_id() function which required the thread to already have its
sys_thread_t pointer, which basically defeats the purpose.  In
addition the return value of sys_thread_id() was pthread specific and
should only be compared using pthread_equal().  sys_thread_self()
return value is a plain pointer that can be compared in line.

### DIFF
--- a/src/posix/threading.c
+++ b/src/posix/threading.c
@@ -150,18 +150,37 @@ void sys_rwlock_unlock(sys_rwlock_t *lock)
 
 struct sys_thread_t {
     pthread_t thread;
+    void *(*f)(void *);
+    void *arg;
 };
+
+static __thread sys_thread_t *_self;
+
+
+// bootstrap _self before calling main thread function
+//
+static void *_thread_init(void *arg)
+{
+    _self = (sys_thread_t *)arg;
+    return _self->f(_self->arg);
+}
+
 
 sys_thread_t *sys_thread(void *(*run_function) (void *), void *arg)
 {
     sys_thread_t *thread = NEW(sys_thread_t);
-    pthread_create(&(thread->thread), 0, run_function, arg);
+    thread->f = run_function;
+    thread->arg = arg;
+    pthread_create(&(thread->thread), 0, _thread_init, (void *)thread);
     return thread;
 }
 
-long sys_thread_id(sys_thread_t *thread) {
-    return (long) thread->thread;
+
+sys_thread_t *sys_thread_self()
+{
+    return _self;
 }
+
 
 void sys_thread_free(sys_thread_t *thread)
 {
--- a/src/posix/threading.c
+++ b/src/posix/threading.c
@@ -161,7 +161,7 @@ static __thread sys_thread_t *_self;
 //
 static void *_thread_init(void *arg)
 {
-    _self = (sys_thread_t *)arg;
+    _self = (sys_thread_t*) arg;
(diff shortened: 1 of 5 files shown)


src/posix/threading.c
```diff
@@ -150,19 +150,38 @@ void sys_rwlock_unlock(sys_rwlock_t *lock)
...
-    pthread_create(&(thread->thread), 0, run_function, arg);
+    thread->f = run_function;
+    thread->arg = arg;
+    pthread_create(&(thread->thread), 0, _thread_init, (void *)thread);
     return thread;
 }
 
-long sys_thread_id(sys_thread_t *thread) {
-    return (long) thread->thread;
+
```
ChugR, line 178: Is sys_thread_id() now gone? Is the address returned by sys_thread_self now sufficient to be the "id"?

jiridanek:
N.B. I am responsible for the death of previous incarnation of `sys_thread_self` in https://github.com/apache/qpid-dispatch/commit/3db622312ecaf6b872af01c258f4687fdb5d9720. The problem with previous implementation was that it assumed `pthread_self()` returns (typedef'd alias of) `long`, which is not always true. It's not true on macOS. Nobody was using that function then, so it quietly died.

src/posix/threading.c
kgiusti: Yes and Yes!
There were no calls to sys_thread_id() in the code... so zero risk in removing it.
Each thread is allocated a unique sys_thread_t pointer that exists in shared memory so every thread's pointer will be globally unique.  One of the new unit tests verifies that is the case.

kgiusti:
> N.B. I am responsible for the death of previous incarnation of `sys_thread_self` in [3db6223](https://github.com/apache/qpid-dispatch/commit/3db622312ecaf6b872af01c258f4687fdb5d9720). The problem with previous implementation was that it assumed `pthread_self()` returns (typedef'd alias of) `long`, which is not always true. It's not true on macOS. Nobody was using that function then, so it quietly died.

I'm hoping the new sys_thread_self() call will work on macOS as it avoids exporting the pthread_t type.

src/posix/threading.c
```diff
@@ -150,19 +150,38 @@ void sys_rwlock_unlock(sys_rwlock_t *lock)
...
 };
 
+static __thread sys_thread_t *_self;
+
+
+// bootstrap _self before calling main thread function
+//
+static void *_thread_init(void *arg)
+{
+    _self = (sys_thread_t *)arg;
```
ted-ross, line 164: Style:
`(sys_thread_t*) arg;`

src/posix/threading.c
```diff
@@ -150,19 +150,38 @@ void sys_rwlock_unlock(sys_rwlock_t *lock)
...
+}
+
+
 sys_thread_t *sys_thread(void *(*run_function) (void *), void *arg)
 {
     sys_thread_t *thread = NEW(sys_thread_t);
-    pthread_create(&(thread->thread), 0, run_function, arg);
+    thread->f = run_function;
+    thread->arg = arg;
+    pthread_create(&(thread->thread), 0, _thread_init, (void *)thread);
```
ted-ross, line 174: Style:
`(void*) thread`

PR #761, opened 2020-06-11
DISPATCH-1657: DISPATCH-1545: check for un-initialized router connect…

kgiusti:
…ion masks

There can be a delay between when a new remote router is discovered
and the connection over which that router is accessed is identified.
If a message is forwarded before this process completes the connection
lookup fails.  The patch checks if the failure occurs and avoids
forwarding the message.

### DIFF
--- a/src/router_core/forwarder.c
+++ b/src/router_core/forwarder.c
@@ -660,23 +660,25 @@ int qdr_forward_closest_CT(qdr_core_t      *core,
                 qd_bitmask_first_set(addr->closest_remotes, &addr->next_remote);
 
             // get the inter-router connection associated with path to rnode:
-            int conn_bit = (rnode->next_hop) ? rnode->next_hop->conn_mask_bit : rnode->conn_mask_bit;
-            qdr_link_t *out_link;
-            if (control) {
-                out_link = peer_router_control_link(core, conn_bit);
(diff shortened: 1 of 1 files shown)


src/router_core/forwarder.c
```diff
@@ -660,23 +660,25 @@ int qdr_forward_closest_CT(qdr_core_t      *core,
...
-            qdr_link_t *out_link;
-            if (control) {
-                out_link = peer_router_control_link(core, conn_bit);
-            } else if (!receive_complete) {
-                out_link = get_outgoing_streaming_link(core, core->rnode_conns_by_mask_bit[conn_bit]);
-            } else {
-                out_link = peer_router_data_link(core, conn_bit, qdr_forward_effective_priority(msg, addr));
-            }
+            const int conn_bit = (rnode->next_hop) ? rnode->next_hop->conn_mask_bit : rnode->conn_mask_bit;
+            if (conn_bit >= 0) {
```
ted-ross, line 664: Did you also check the equivalent logic in qdr_forward_balanced_CT?  I think this is a good update.

kgiusti: Yeah I checked each access to node->conn_mask_bit and use of the conn bit in that file and elsewhere.  I'm pretty sure I've covered all the bases.

PR #756, opened 2020-06-08
Dispatch 1684 SwitchoverTest opens fallback receiver on primary connection and not on fallback connection

ChugR:
DISPATCH-1684: fallback dest test improvement reveals router bug. Work In Progress - do not merge

PR #765, opened 2020-06-19
DISPATCH-1693: Don't raise link event on fallback when primary closes

ChugR:
QDRC_EVENT_ADDR_ONE_SOURCE event on fallback link provokes a link
closure.
This commit fixes an issue exposed by DISPATCH-1684 (https://github.com/apache/qpid-dispatch/pull/756)
The test suite passes with this fix but are there other cases where closing the fallback link is desirable?


ChugR:
Fixed at commit bb89cc70

PR #767, opened 2020-06-25
DISPATCH-1660: Avoid assert in proton by not settling delivery

ChugR:
The qpid-proton assertion is 'delivery must not be local.settled'.
This commit avoids setting local.settled. The connection is never
again read and no new deliveries on this link or connection are
created.


ChugR:
Before this patch a system may run the test more than two thousand times before seeing the failure. The sequence of events that are special to trigger the assert remain a mystery.

ChugR:
This change causes failures in system_tests_policy_oversize_compound and is hereby withdrawn.

PR #771, opened 2020-07-01
Add a short note about the multi-tenancy flag

ctron:


docs/books/modules/user-guide/listening-client-connections.adoc
```diff
@@ -59,6 +60,7 @@ listener {
 +
 For more information, see xref:understanding-connection-failover-{context}[].
 --
+`multiTenant` (optional):: If true, apply multi-tenancy to endpoints connected at this listener.  The address space is defined by the virtual host (hostname field in the Open).
```
jiridanek, line 63: Afaik there isn't any other documentation for multi-tenancy yet. Adding just this note will not help those who want to learn about it (what Open?, it is not clear that stands for the AMQP Open performative).

Adding just this will not help those who want to use it and it be confusing for people who do not need this feature.

ctron: Well, this is a property that is part of the router. And the description is the exact same description from the schema, like for all the other properties.

I agree that having a full blow explanation of what this is doing would be great. But I also think that we should at least mention that this feature exists. Otherwise it is difficult to explain other why we are using an undocumented feature.

pwright:
@ctron you might be interested to know that I recently created https://issues.apache.org/jira/browse/DISPATCH-1710 to address the larger issue.  AFAIK there was not much use of this feature, sounds like that's changed?
I'm confused by "The address space is defined by the virtual host (hostname field in the Open)."... is that Enmasse specific?

docs/books/modules/user-guide/listening-client-connections.adoc
jiridanek: You're completely right. If this is merged, it would be the third copy of this same paragraph in dispatch ;)

It's not undocumented feature, even now. You can see it in the config file docs at https://qpid.apache.org/releases/qpid-dispatch-1.12.0/man/qdrouterd.conf.html#_listener that are autogenerated from the schema.  Afaik it is not a good idea to have the `listener` configuration example show every conceivable configuration option, just because it was a cool and new configuration option at some point in time.

jiridanek:
> "The address space is defined by the virtual host (hostname field in the Open)."... is that Enmasse specific?

The hostname field in Open performative is regular AMQP, http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-open. The idea of a virtual host is afaik a dispatch concept. I am unsure about address space. That sounds to me as an Enmasse specific concept, which seems to be slowly leaking into dispatch https://github.com/apache/qpid-dispatch/search?q=%22address+space%22&unscoped_q=%22address+space%22.

Tbh, I am not personally clear what address space actually is. I am guessing that it is a namespace for addresses, scoped to tenants. Two endpoints, each in different tenant, can listen on the same address, but messages sent to that address don't intermix.

PR #773, opened 2020-07-03
DISPATCH-1703: correctly propagate disposition data

kgiusti:


gemmellr:
I tried out the changes with the HelloWorld reproducer and some variations, they looked to resolve the issue.

fgiorgetti:
Tried it using an existing reproducer that performs commits and rollbacks and it worked fine this time.

PR #775, opened 2020-07-10
Dispatch 1708 Fallback dest test_28 receives phase 0 messages on fallback link.

ChugR:
Don't commit this until the bug[1] is fixed.

This PR releases phase 0 messages received on fallback link.

This PR also includes housekeeping changes:

- mnemonic names for the router ports to use in each test
- unique payload for every transfer: improves scraper analysis
- link names based on test and link role

[1] with this patch test_28 fails hard with EA1 never getting notified of the loss of the primary receiver on EB1.

### DIFF
--- a/tests/system_tests_fallback_dest.py
+++ b/tests/system_tests_fallback_dest.py
@@ -589,9 +589,11 @@ class SwitchoverTest(MessagingHandler):
         self.n_rel          = 0
         self.phase          = 0
         self.tx_seq         = 0
+        self.local_rel      = 0
 
     def timeout(self):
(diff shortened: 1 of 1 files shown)


tests/system_tests_fallback_dest.py
```diff
@@ -81,356 +81,370 @@ def router(name, mode, connection, extra=None):
         cls.routers[0].wait_router_connected('INT.B')
         cls.routers[1].wait_router_connected('INT.A')
 
+        cls.ROUTER_INTA = cls.routers[0].addresses[0]
```
ted-ross, line 84: Nice!  This is a major improvement.

PR #777, opened 2020-07-19
DISPATCH-1716 Don't set -Wl,-z,defs in -fsanitize= builds with Clang

jiridanek:


jiridanek:
This is genuinely funny. Both the default `virtualenv` version (https://github.com/pypa/virtualenv/issues/1740) and the latest one (https://github.com/pypa/virtualenv/issues/1873) contain a bug which breaks `tox` on Focal Fossa.

jiridanek:
I don't have any significantly better ideas how to parameterize the `-Wl,-z,defs` option, so I'll leave it at this.

I'm going to merge the PR eod today, unless somebody complains.

PR #779, opened 2020-07-21
DISPATCH-1720 Add GitHub Action to run console tests through npm

jiridanek:
CC @gemmellr @kpvdr my first real GH Action (not counting the rat check)

### DIFF
--- a/.github/workflows/console.yaml
+++ b/.github/workflows/console.yaml
@@ -1,0 +1,29 @@
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: cd console/react && npm install
+
+      - name: Run Tests
+        run: cd console/react && npm run test
+
+  eslint:
+    name: ESLint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: cd console/react && npm install
+
+      - name: Install Typescript
+        run: cd console/react && npm install typescript
+
+      - name: Run ESLint
+        run: cd console/react && npx eslint --max-warnings 0 .
--- a/.github/workflows/console.yaml
+++ b/.github/workflows/console.yaml
@@ -1,3 +1,22 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


gemmellr:
Is the javascript action doing anything? It doesnt appear to in the log, and its repo makes it seem more like a base action others would build their own Actions upon than anything else. 

I think it would be nice to give the jobs+steps names, rather than just having the commands show up in the overview etc. I prepared a tweaked version that drops the javascript action and adds names etc:
```
on: [push, pull_request]

jobs:
  test:
    name: Tests
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2

      - name: Install Console
        run: cd console/react && npm install

      - name: Run Tests
        run: cd console/react && npm run test

  eslint:
    name: ESLint
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2

      - name: Install Console
        run: cd console/react && npm install

      - name: Install Typescript
        run: cd console/react && npm install typescript

      - name: Run ESLint
        run: cd console/react && npx eslint --max-warnings 0 .
```



Either version could perhaps be expanded to run the tests on different NodeJS versions, e.g see: https://docs.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions. Note that if doing that, I would also set the strategy config for "fail-fast: false" otherwise it aborts the entire matrix by default if any of them fail.

P.S. maybe squash commits like these before raising the PR ;)

gemmellr:
Also, youll need a exclude  to fix the RAT check failing on the new GitHub Actions config, maybe something like:
```
diff --git a/pom.xml b/pom.xml
index 14b623c5..698e510c 100644
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,7 @@
                 <artifactId>apache-rat-plugin</artifactId>
                 <configuration>
                     <excludes>
+                      <exclude>.github/workflows/*.yml</exclude>
                       <exclude>.travis.yml</exclude>
                       <exclude>CMakeCache.txt</exclude>
                       <exclude>.gitmodules</exclude>
```

jiridanek:
Yeah, forgot about rat, again.

The action seems to work just fine without the using: javascript, so I'll leave it out.

I don't want to do the matrix, really. The javascript under test is running in the browser when the console is in operation. There is little point in testing it under various nodejs versions. Maybe later, or if somebody more knowledgeable of javascript think it is helpful thing to do.

Btw, `npm install` is only fetching project dependencies from the internet, it is not at all like `mvn install`

gemmellr:
I believe the javascript action is something people would use to construct other actions a workflow might use, e.g say like the checkout action or others. I dont think its one people use directly.

Fair enough about the NodeJS versions. I have no idea what running npm run test is actually doing, so I just thought I'd point it in case it were relevant what its running on.

Hehe re: npm install. I dont do really do any JS so I couldn't have told you that. To be fair, maven does also do that so its not that disimilar, it just also installs without needing extra flags.

jiridanek:
> npm install

I don't do js either, so that's why it's why these details are a big deal for me. One significant difference is that Maven eagerly fetches things from the internet when it needs them. NPM doesn't. So running `npm run test` without doing `install` first will fail.

> npm run test

Just fyi, it is user specified alias. It runs whatever command line is specified in `package.json` that it should run. The Console configured it to run some unittests, which makes sense. It seems to be using https://jestjs.io/, which is probably the default runner/framework choice for react. Jest does not believe in testing in browser; it is apparently somehow possible to use it for that, but at Facebook they don't do it, so it's all in nodejs, for speed and isolation, and these thing. And they have this feature they call "snapshot testing", https://jestjs.io/docs/en/snapshot-testing, which is, ... interesting idea prone to much misuse, IMO.

PR #780, opened 2020-07-22
DISPATCH-1721: Qdstat table vhostgroups: too wide; missing default settings

ChugR:
At a minimum the table is about 200 columns wide. If large lists of user
names and/or sources and targets are specified the table is so wide
that it becomes unreadable.

 * Break up groups display into single table for boolean and integer
   settings and individual tables for groups strings that might be large.

Entity management does not return appropriate defaults for table
vhostUserGroupSettings.

 * Fetch the schema from the target router and use those defaults.

 * Enforce a 'false' default on booleans that are not required and have
   no default. The fetched schema does not include the 'false' default.

### DIFF
--- a/python/qpid_dispatch/management/client.py
+++ b/python/qpid_dispatch/management/client.py
@@ -384,3 +384,6 @@ class Node(object):
 
     def get_log(self, limit=None, type=None):
         return self.call(self.node_request(operation=u"GET-LOG", entityType=type, limit=limit)).body
+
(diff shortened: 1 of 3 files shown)


jiridanek:
I am not sure if this would've been worth the effort, but one thing that would make this PR much easier to review would be a test which prints-out few dummy entries formatted the way qdstat output is formatted, and asserts it is equal to some predefined string `""" this-is-what-it-looks-like """`.

PR #782, opened 2020-07-23
DISPATCH-1722 - Fixed qdstat to display correct value of Worker Threa…

ganeshmurthy:
…ds. Also, used regex to match output of qdstat instead of using string compare

### DIFF
--- a/tests/system_tests_qdstat.py
+++ b/tests/system_tests_qdstat.py
@@ -66,24 +66,33 @@ class QdstatTest(system_test.TestCase):
 
     def test_general(self):
         out = self.run_qdstat(['--general'], r'(?s)Router Statistics.*Mode\s*Standalone')
-        self.assertTrue("Connections                      1" in out)
-        self.assertTrue("Nodes                            0" in out)
-        self.assertTrue("Auto Links                       0" in out)
-        self.assertTrue("Link Routes                      0" in out)
-        self.assertTrue("Worker Threads                   1" in out)
-        self.assertTrue("Router Id                        QDR.A" in out)
-        self.assertTrue("Mode                             standalone" in out)
+
+        self.assertTrue(re.match(r"(.*)\bConnections\b[ \t]+\b1\b(.*)",
+                                 out, flags=re.DOTALL) is not None, out)
+        self.assertTrue(re.match(r"(.*)\bNodes\b[ \t]+\b0\b(.*)",
+                                 out, flags=re.DOTALL) is not None, out)
+        self.assertTrue(re.match(r"(.*)\bAuto Links\b[ \t]+\b0\b(.*)",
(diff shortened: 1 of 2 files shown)


tests/system_tests_qdstat.py
```diff
@@ -66,24 +66,33 @@ def test_help(self):
...
-        self.assertTrue("Connections                      1" in out)
-        self.assertTrue("Nodes                            0" in out)
-        self.assertTrue("Auto Links                       0" in out)
-        self.assertTrue("Link Routes                      0" in out)
-        self.assertTrue("Worker Threads                   1" in out)
-        self.assertTrue("Router Id                        QDR.A" in out)
-        self.assertTrue("Mode                             standalone" in out)
+
+        self.assertTrue(re.match(r"(.*)\bConnections\b[ \t]+\b1\b(.*)",
+                                 out, flags=re.DOTALL) is not None, out)
```
jiridanek, line 71: There's `assertRegexpMatches` in Python 2.7 and unittest2. It does `re.search()` which is a better fit here (avoids having to put `.*` at the beginning and the end) and should give better assertion failure messages. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex

tests/system_tests_qdstat.py
```diff
@@ -66,24 +66,33 @@ def test_help(self):
...
     def test_general_csv(self):
         out = self.run_qdstat(['--general', '--csv'], r'(?s)Router Statistics.*Mode","Standalone')
-        self.assertTrue("Connections","1" in out)
-        self.assertTrue("Worker Threads", "1" in out)
-        self.assertTrue("Nodes","0" in out)
-        self.assertTrue("Auto Links","0" in out)
-        self.assertTrue("Link Routes","0" in out)
-        self.assertTrue("Router Id","QDR.A" in out)
-        self.assertTrue("Mode","standalone" in out)
+        self.assertTrue('"Connections","1"' in out)
```
jiridanek, line 89: `assertIn`, again, gives better error message if the assert fails.

ganeshmurthy: Thanks I used assertIn instead of assertTrue. I was not expecting it to work on RHEL 6 but it did

tests/system_tests_qdstat.py
ganeshmurthy: assertRegexpMatches gives a deprecation warning on Python 3.8 since it is replaced by assertRegex. I cannot use assertRegex because it fails on RHEL 6.

jiridanek: That deprecation warning IMO deserves a targetted suppression there, but that's just my view.

I never really used the 'warnings.filterwarnings(action, ...', but it looks viable.

PR #783, opened 2020-07-26
DISPATCH-1725 Fix some Jest Web Console unittests

jiridanek:
CC @ErnieAllen, thanks.

### DIFF
--- a/console/react/package.json
+++ b/console/react/package.json
@@ -48,6 +48,7 @@
     "@react-mock/localstorage": "^0.1.2",
     "@testing-library/jest-dom": "^4.2.3",
     "@testing-library/react": "^9.3.2",
+    "@types/jest": "^24.9.0",
     "body-parser": "^1.19.0",
     "browserslist": "^4.7.3",
     "jest-axe": "^3.2.0",
--- a/console/react/src/details/detailsTablePage.js
+++ b/console/react/src/details/detailsTablePage.js
@@ -42,7 +42,7 @@ import { dataMap } from "../overview/entityData";
 import { dataMap as detailsDataMap, defaultData } from "./entityData";
 import Updated from "../common/updated";
 
-class DetailTablesPage extends React.Component {
+class DetailsTablePage extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
@@ -58,7 +58,7 @@ class DetailTablesPage extends React.Component {
       redirect: false,
       redirectState: { page: 1 },
       redirectPath: "/dashboard",
-      lastUpdated: new Date()
+      lastUpdated: new Date(),
     };
     // if we get to this page and we don't have a props.location.state.entity
     // then redirect back to the dashboard.
@@ -105,7 +105,7 @@ class DetailTablesPage extends React.Component {
         if (!this.unmounted)
           this.setState({ rows, lastUpdated: new Date() }, () => {
             if (this.props.details) {
-              this.props.lastUpdated(this.state.lastUpdated);
+                this.props.lastUpdated(this.state.lastUpdated);
             }
           });
       },
@@ -239,4 +239,4 @@ class DetailTablesPage extends React.Component {
   }
 }
 
-export default DetailTablesPage;
+export default DetailsTablePage;
--- a/console/react/src/details/detailsTablePage.js
+++ b/console/react/src/details/detailsTablePage.js
@@ -105,7 +105,7 @@ class DetailsTablePage extends React.Component {
         if (!this.unmounted)
           this.setState({ rows, lastUpdated: new Date() }, () => {
             if (this.props.details) {
-                this.props.lastUpdated(this.state.lastUpdated);
+              this.props.lastUpdated(this.state.lastUpdated);
             }
           });
       },
--- a/console/react/src/details/detailsTablePage.test.js
+++ b/console/react/src/details/detailsTablePage.test.js
@@ -39,7 +39,8 @@ it("renders the DetailTablesPage", () => {
       },
       details: true,
       service,
-      schema: service.schema
+      schema: service.schema,
+      lastUpdated: (date) => {},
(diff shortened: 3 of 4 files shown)


console/react/src/details/detailsTablePage.js
```diff
@@ -42,7 +42,7 @@ import { dataMap } from "../overview/entityData";
 import { dataMap as detailsDataMap, defaultData } from "./entityData";
 import Updated from "../common/updated";
 
-class DetailTablesPage extends React.Component {
+class DetailsTablePage extends React.Component {
```
jiridanek, line 45: Just aligning the name with the rest of the app (I got confused by this before, I grepped for the class name ans so it nowhere used.)

console/react/src/details/detailsTablePage.js
```diff
@@ -105,7 +105,7 @@ class DetailTablesPage extends React.Component {
         if (!this.unmounted)
           this.setState({ rows, lastUpdated: new Date() }, () => {
             if (this.props.details) {
-              this.props.lastUpdated(this.state.lastUpdated);
+                this.props.lastUpdated(this.state.lastUpdated);
```
jiridanek, line 108:
```suggestion
              this.props.lastUpdated(this.state.lastUpdated);
```

console/react/src/details/detailsTablePage.test.js
```diff
@@ -39,7 +39,8 @@ it("renders the DetailTablesPage", () => {
       },
       details: true,
       service,
-      schema: service.schema
+      schema: service.schema,
+      lastUpdated: (date) => {},
```
jiridanek, line 43: Fix for `TypeError: this.props.lastUpdated is not a function`

console/react/test_data/qdrService.mock.js
```diff
@@ -37,19 +37,21 @@ export const mockService = ({ onSendMethod }) => {
   const cbSendMethod = onSendMethod ? onSendMethod : () => {};
   return {
     management: {
+      schema: () => schema,
```
jiridanek, line 40: Fix for `TypeError: this.props.service.management.schema is not a function`

console/react/test_data/qdrService.mock.js
```diff
@@ -60,17 +62,13 @@ export const mockService = ({ onSendMethod }) => {
             resolve();
           });
         },
-        startUpdating: () => {
-          return new Promise(resolve => {
-            resolve();
-          });
-        },
```
jiridanek, line 67: Fix for `63:9  warning  Duplicate key 'startUpdating'  no-dupe-keys`

console/react/test_data/qdrService.mock.js
```diff
@@ -37,19 +37,21 @@ export const mockService = ({ onSendMethod }) => {
...
+      schema: () => schema,
       connection: {
         sendMethod: () => {
           cbSendMethod();
           return Promise.resolve(methodResults);
         },
         is_connected: () => true,
-        setReconnect: () => {}
+        setReconnect: () => {},
+        getReceiverAddress: () => 'mockReceiverAddress',
```
jiridanek, line 48: Fix for `TypeError: this.service.management.connection.getReceiverAddress is not a function`

console/react/package.json
```diff
@@ -48,6 +48,7 @@
     "@react-mock/localstorage": "^0.1.2",
     "@testing-library/jest-dom": "^4.2.3",
     "@testing-library/react": "^9.3.2",
+    "@types/jest": "^24.9.0",
```
jiridanek, line 51: Adds type hinting for jest in IDE.

console/react/test_data/qdrService.mock.js
ErnieAllen: It's probably better to remove the startUpdating on line 52 and leave this one here. By returning a promise and then calling resolve(), this will allow the code that calls startUpdating().then(...) to run the code contained in the then(...).

jiridanek: That's what I've done, except I've organized it that startUpdating and stopUpdating are next to each other. The new line 52 reads `startUpdating: () => new Promise(resolve => resolve()),` which is exactly what was here, except written as oneliner.

jiridanek: ... uh, no, it isn't exactly the same, the difference is between `return resolve()` and just `resolve()`. But it may not matter...

ErnieAllen: return resolve() is odd, but it should still work.

console/react/test_data/qdrService.mock.js
ErnieAllen: Probably better to return 'amqp:/_topo/0/routerName/' instead of 'mockReceiverAddress'. The code that calls getReceiverAddress() will attempt to split the returned string on '/', add '$management' to the resulting array, and then contruct a router id from the array. 
However, that code would only be called if there was only 1 router in the network that is being tested. The qdrMockService.js file sets up 2 routers, so that code is never hit in tests that use qdrMockService.js anyway. But if that ever changes, it would be good to return an actual router ID.

console/react/src/details/detailsTablePage.js
ErnieAllen: Good catch. Since the class name is the default export, it can be imported as any name. And indeed, it is imported as DetailsTablePage in layout.js. But for sanity sake, it is definitely better to have the export and import name match.

console/react/test_data/qdrService.mock.js
jiridanek, line 48:
```suggestion
        getReceiverAddress: () => 'amqp:/_topo/0/routerName/',
```

jiridanek: I'll change it per your suggestion.

PR #786, opened 2020-07-27
NO-JIRA Configure .github/dependabot.yml

jiridanek:


jiridanek:
I won't merge this (yet) because I am unsure what this is going to do.

My goal is to make it easy to keep this file in my GitHub fork, so I get the dependency updates myself.

Adding the file into the apache/qpid-dispatch repository is either going to do nothing, or it will enable the automated dependency update PRs. Even if it enabled the automated PRs, I think it is survivable. In `version: 2`, the PRs are not automatically updated once created, so if they are left alone, they do not keep producing noise. And anyways, keeping components up-to-date is actually a good thing to do, if there is the bandwidth to keep up with the updates, esp. regarding testing them.

Again, I'll just leave this here for the time being, without merging it.

jiridanek:
There's probably nothing to worry about. Again, the worst case is the one-time storm of autogenerated PRs, which we may actually want to merge, at some point or another; or possibly use as inspiration to do the updating manually.

PR #803, opened 2020-07-30
DISPATCH-1733 - Free leaking qdr_connection_t when the router core ac…

ganeshmurthy:
…tion has been discarded

### DIFF
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -1381,6 +1381,10 @@ static void qdr_connection_opened_CT(qdr_core_t *core, qdr_action_t *action, boo
         qdrc_event_conn_raise(core, QDRC_EVENT_CONN_OPENED, conn);
     }
 
+    if (discard && conn) {
+        qdr_connection_free(conn);
+    }
+
     qdr_field_free(action->args.connection.connection_label);
     qdr_field_free(action->args.connection.container_id);
 }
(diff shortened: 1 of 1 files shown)


src/router_core/connections.c
```diff
@@ -1381,6 +1381,10 @@ static void qdr_connection_opened_CT(qdr_core_t *core, qdr_action_t *action, boo
         qdrc_event_conn_raise(core, QDRC_EVENT_CONN_OPENED, conn);
     }
 
+    if (discard && conn) {
```
kgiusti, line 1384: Code nit - why stuff this way down on the tail?   Why not do an immediate check for discard at the start of the function?  That would be consistent with how the discard flag is handled w.r.t. other action routines:
if (discard) {
   if (conn) qd_connection_free(conn);
   return;
}

Added bonus - drops a whole level of indentation.

ganeshmurthy: Thanks @kgiusti I have added code to check conn and discard at the top of the functions like other action routines and also dropped a level of indentation. The diff looks weird because of dropping one level of  indentation

src/router_core/connections.c
```diff
@@ -1312,78 +1312,81 @@ void qdr_check_addr_CT(qdr_core_t *core, qdr_address_t *addr)
...
 static void qdr_connection_opened_CT(qdr_core_t *core, qdr_action_t *action, bool discard)
 {
-
     qdr_connection_t *conn = safe_deref_qdr_connection_t(action->args.connection.conn);
-    if (!discard && conn) {
 
-        do {
-            DEQ_ITEM_INIT(conn);
-            DEQ_INSERT_TAIL(core->open_connections, conn);
+    if (!conn)
```
kgiusti, line 1317: Oh wait - if the conn has been freed whilst the action was queued conn will be null but connection_label and container_id will still need to be freed, right?

ganeshmurthy: Agreed, new commit pushed to take care of that case. thanks much

PR #804, opened 2020-07-31
DISPATCH-960 Use getservbyname on macOS (getservbyname_r not available)

jiridanek:


jiridanek:
This takes care of the macOS issue. I want to get macOS CI running again quickly. I am disheartened to see that additional tests started to fail on macOS while the build there was broken.

PR #805, opened 2020-07-31
Proposal for run-time lock validation

kgiusti:


docs/notes/RIPs/lock-validation.adoc
```diff
@@ -0,0 +1,101 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License
+////
+
+= Mutex Locking Validation
+
```
jiridanek, line 21: I propose a header for RIPs, consisting of the following metadata.

> Authors: Jane Doe
> Status: **Submitted** | Superseded | Implemented
> Last updated: 2020-04-05

(I am taking this from https://docs.google.com/document/d/1ef7_drjTl4NnzfAmGPzC4_R9vGHkpIzUrlGhxPqTbwg/edit)

jiridanek, line 21: There should be a `README.adoc` that explains what the RIPs are. I think that the following phrase (or a variant of it) should be part of it

> It’s possible that designs change as they are implemented in practice. The published design documents capture the initial design, and not the ongoing changes as designs are implemented.
>
> Always go to the documentation for descriptions of current Qpid Dispatch functionality.

(this is from https://bazel.build/designs/)

docs/notes/RIPs/lock-validation.adoc
```diff
@@ -0,0 +1,101 @@
...
+Example: a priority 5 lock may be taken when the thread is either not
+holding any locks or is holding lock(s) with priority < 5.  Attempting
+to take a priority 2 lock while holding the priority 5 lock is an
+error.
+
+In order to track lock nesting within a thread each thread will
+maintain a per-thread stack of active locks.  Taking a new lock will
+result in that lock being pushed onto the stack.  Attempting to push a
+lock in the wrong priority order will result in calling abort().
```
jiridanek, line 91: Do you intend to check that the lock being unlocked is the one on top of the lock stack?

docs/notes/RIPs/lock-validation.adoc
```diff
@@ -0,0 +1,101 @@
...
+A violation of any of the above rules will result in a call to abort().
+
+== Implementation Notes:
+
+Every lock will be assigned a fixed priority on creation.  The
+priority will be a simple positive integer.
+
+A lock with a priority of zero is an exclusive lock.  A thread must
+never hold another lock when taking a priority 0 lock. While holding a
+priority 0 lock it is an error to take any other lock.
```
jiridanek, line 73: Do you have any example of a lock that would get priority 0? Is this actually good for anything?

I was thinking of possibly using MAX_INT instead of 0, to avoid a special case. But there still would be a special case, because more than one lock may need to have this priority (and the rest of the proposal afaik disallows having multiple locks with the same priority).

docs/notes/RIPs/lock-validation.adoc
jiridanek, line 21: That `README.adoc` should explain what RIP stands for ;) I actually don't know. (But it sounds really cool.)

docs/notes/RIPs/lock-validation.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+In general lock nesting (e.g. taking a lock while holding another) is
+not permitted.
+
+However in a few instances locks will nest. One example of permitted lock
+nesting is the Python lock (qd_python_lock()).  This lock must be held
+while a thread is running within the Python interpreter.  Occasionally
+the Python interpreter will invoke native C functions.  These
+functions may take addional locks, such as the global logging lock
+(log_lock).
```
jiridanek, line 42: offtopic: is it necessary to have logging lock? There is lock-free logging in Java, so maybe C has it too?

docs/notes/RIPs/lock-validation.adoc
jiridanek, line 21: The proposal files could be numbered, so possibly `001-lock-validation.adoc` or maybe dated, `2020-07-31-lock-validation.adoc`.

docs/notes/RIPs/lock-validation.adoc
```diff
@@ -0,0 +1,101 @@
...
+locking is well behaved.
+
+== Introduction
+
+The Qpid Dispatch Router is a multi-threaded application.  These
+threads share state and use locks to ensure the consistency of that
+state.  The locking API is defined in
+include/qpid/dispatch/threading.h
+
+In general lock nesting (e.g. taking a lock while holding another) is
```
ted-ross, line 34: I think this statement is a bit strong.  There are a number of cases where locks are nested in the code base.

docs/notes/RIPs/lock-validation.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+* Unlocked paths: a mechanisim to ensure that a thread is not currently holding any locks during a particular code path.
+
+A violation of any of the above rules will result in a call to abort().
+
+== Implementation Notes:
+
+Every lock will be assigned a fixed priority on creation.  The
+priority will be a simple positive integer.
+
```
ted-ross, line 70: Are you proposing that the sys_mutex function be modified to include a lock priority?  Alternatively, are you planning to add an alternate constructor (sys_mutex_priority or similar) and having the existing one default to a particular value?  If so, what's the default?

docs/notes/RIPs/lock-validation.adoc
ted-ross: I agree that this feels inverted and it will be difficult to choose an appropriate priority for your lock.

Instead of saying 0 can't be acquired while another lock is held, could we say that when a 0 is held, no other locks may be acquired?  This would suggest a default priority of 0, which would apply to locks that are acquired and released in-line while data structures are manipulated and no functions are called.

This would mean that a lock's priority would be based on the priorities of other locks that may be recursively acquired, establishing a lock hierarchy.  Locks like the Python lock should have a relatively high priority since they are held over long-running operations.

docs/notes/RIPs/lock-validation.adoc
ChugR: Log locking is used to keep the written log in strict chronological order. gettimeofday, vsnprintf, and write_log are the items that need to be locked but the code locks only a little more than that.
It would be bad to have the logs out of order with respect to time of day.

docs/notes/RIPs/lock-validation.adoc
ChugR: An example would be really helpful. Assign some priorities to existing locks (connection work lock, log lock, python lock, etc.) and see what and what isn't allowed.

ssorj:
I propose putting these under /docs/rips/ (as a peer to notes/, man/, and so on).

jiridanek:
Somebody built a tool that does a lot of what's proposed here http://manpages.ubuntu.com/manpages/bionic/man8/deadlock_detector-bpfcc.8.html. TSan is of course another such tool.

docs/notes/RIPs/lock-validation.adoc
jiridanek: I meant something like http://logging.apache.org/log4j/2.x/manual/async.html, this keeps a queue of in-flight log messages and adding to the queue does not require holding a lock.

PR #806, opened 2020-07-31
DISPATCH-1736 - Removed skip_test flag that was wrongly added in comm…

ganeshmurthy:
…it 4a173e1ec0f7e0e8f9d7d37b9549d4f9c9d37fb0. This test should run all arches, 32 or 64 bit

### DIFF
--- a/tests/system_tests_protocol_settings.py
+++ b/tests/system_tests_protocol_settings.py
@@ -26,15 +26,6 @@ from system_test import TestCase, Qdrouterd, main_module
 from system_test import unittest
 from proton.utils import BlockingConnection
 import subprocess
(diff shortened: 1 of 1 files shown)


tests/system_tests_protocol_settings.py
```diff
@@ -245,7 +234,6 @@ def test_max_session_frames_default(self):
             # if frame size not set then a default is used
             self.assertTrue(" max-frame-size=16384" in open_lines[0])
             begin_lines = [s for s in log_lines if "-> @begin" in s]
-            # incoming-window is from the config
```
jiridanek, line 248: I'd instead explain where is the number coming from. Maybe

```# incoming-window is the Proton default```

PR #809, opened 2020-08-02
DISPATCH-1739 Add GitHub Action w/ sharding and bubblewrap

jiridanek:
This is a CI job triggered by pushes and pull requests sent to GitHub.

Output artifacts include parts of the workspace before running tests; this is
what the machines running test shards get. At the end, each shard outputs
a XML CTest report.

Note that `ctest -j2` results in test log lines being mixed up in stdout.
Refer to the output XML report or use grep (grep for lines starting with
the test number) to get understandable logs.

* Using python 3.6 as this is the version that works on all platforms tested.
* Using four shards, as adding more does not help much.
* Using ctest -j2, mostly just to exercise bubblewrap. Bwrap is more useful on Travis

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,199 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  compile:
+    name: "Compile (${{matrix.os}}, ${{matrix.runtimeCheck}})"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
+        buildType: [RelWithDebInfo]
+        runtimeCheck: [asan]
+    env:
+      BuildType: ${{matrix.buildType}}
+      ProtonBuildDir: ${{github.workspace}}/qpid-proton/build
+      DispatchBuildDir: ${{github.workspace}}/qpid-dispatch/build
+      InstallPrefix: ${{github.workspace}}/install
+
+      ProtonCMakeExtraArgs: '-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DBUILD_BINDINGS=python -DRUNTIME_CHECK=${{matrix.runtimeCheck}}'
+      DispatchCMakeExtraArgs: '-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCONSOLE_INSTALL=OFF -DCMAKE_C_FLAGS=-DQD_MEMORY_DEBUG -DRUNTIME_CHECK=${{matrix.runtimeCheck}}'
+
+      CCACHE_BASEDIR: ${{github.workspace}}
+      CCACHE_DIR: ${{github.workspace}}/.ccache
+      CCACHE_COMPRESS: 'true'
+      CCACHE_MAXSIZE: '400MB'
+    steps:
+
+      - name: Show environment (Linux)
+        if: ${{ always() && runner.os == 'Linux' }}
+        run: env -0 | sort -z | tr '\0' '\n'
+
+      - uses: actions/checkout@v2
+        with:
+          repository: 'apache/qpid-proton'
+          ref: 'master'
+          path: 'qpid-proton'
+
+      - uses: actions/checkout@v2
+        with:
+          path: 'qpid-dispatch'
+
+      # https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
+      - name: Prepare ccache timestamp
+        id: ccache_cache_timestamp
+        shell: cmake -P {0}
+        run: |
+          string(TIMESTAMP current_date "%Y-%m-%d-%H;%M;%S" UTC)
+          message("::set-output name=timestamp::${current_date}")
+
+      - uses: actions/cache@v2
+        env:
+          cache-name: cache-ccache
+        with:
+          path: .ccache
+          key: ${{ matrix.os }}-${{ matrix.runtimeCheck }}-${{ env.cache-name }}-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.runtimeCheck }}-${{ env.cache-name }}
+
+      - name: Create Build and Install directories
+        run: mkdir -p "${ProtonBuildDir}" "${DispatchBuildDir}" "{InstallPrefix}"
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+          architecture: x64
+
+      - name: Install python dependencies
+        run: python -m pip install setuptools wheel tox
+
+      - name: Install Linux dependencies
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          sudo apt install -y swig libpython3-dev libsasl2-dev libjsoncpp-dev libwebsockets-dev ccache ninja-build pixz
+
+      - name: Zero ccache stats
+        run: ccache -z
+
+      - name: qpid-proton cmake configure
+        working-directory: ${{env.ProtonBuildDir}}
+        run: cmake "${{github.workspace}}/qpid-proton" "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" "-DCMAKE_BUILD_TYPE=${BuildType}" "-DBUILD_TESTING=OFF" "-DENABLE_FUZZ_TESTING=OFF" "-GNinja" ${ProtonCMakeExtraArgs}
+
+      - name: qpid-proton cmake build/install
+        run: cmake --build "${ProtonBuildDir}" --config ${BuildType} -t install
+
+      - name: Display ccache stats
+        run: ccache -s
+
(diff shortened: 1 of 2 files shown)


jiridanek:
One issue I know of. For some reason, dispatch skips SASL tests, even though I believe I installed all that's needed

```
53: test_ssl_sasl_client_invalid (system_tests_ssl.RouterTestSslClient) ... skipped 'Cyrus library not available. skipping test'
53: test_ssl_sasl_client_valid (system_tests_ssl.RouterTestSslClient) ... skipped 'Cyrus library not available. skipping test'
```

I also couldn't find a way to use here the artifacts that the Proton pipeline is building.

CC @gemmellr

gemmellr:
> Note that ctest -j2 results in test log lines being mixed up in stdout.
Refer to the output XML report or use grep (grep for lines starting with
the test number) to get understandable logs.

Or not use -j2? How much of a difference does it make? You commented later its mainly for excercising bwrap and is more useful on Travis.

> Using four shards, as adding more does not help much.

Presumably due to all the setup costs relative to the time spent on tests, or the fact 1 shard is clearly longer than others and maybe still ends up that way?

Also, did you try less than 4? I see one of them has a test run around twice as long as most others, so unless all the long-running tests are perfectly spaced to fall in that shard, is it possible having less shards might actually distribute the test time more evenly such that it isnt much slower overall (if at all; it could even be faster if the tests group better), while giving less different output areas to look at (not personally a huge fan of the sharding, significantly complicates the build config and results, may point to some inefficient tests that it seems needed to begin with)

> I also couldn't find a way to use here the artifacts that the Proton pipeline is building.

I think thats fine, possibly even preferable. Its not that lengthy a build, and its likely to eventually want different versions as e.g the Travis tests use.

jiridanek:
> Or not use -j2?

Probably. I'll leave -DUSE_BWRAP=ON and remove the -j2.

> Also, did you try less than 4?

Yes, there is about 1 min improvement in total time between 3 and 4. I did not try 2.

There is the Python tox test, that is the most lengthy one. It is essentially running the other tests on py27 and then 36. I was thinking of giving it a shard of its own, but it seemed to me overly elaborate.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,165 @@
...
+          path: ${{env.DispatchBuildDir}}/Testing/**/*.xml
+      - name: Upload core files
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: cores_${{matrix.os}}_${{matrix.buildType}}_${{matrix.shard}}
+          path: |
+            **/core
+      - name: Environment (Linux)
+        if: ${{ always() && runner.os != 'macOS' }}
```
gemmellr, line 163: Other bits use "runner.os == 'Linux'", that seems simpler?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,165 @@
...
+          python -m pip install --upgrade pip
+          python -m pip install tox
+      - name: Install Linux dependencies
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          sudo apt install -y bubblewrap swig ninja-build libsasl2-2 sasl2-bin libjsoncpp1 libwebsockets15 xz-utils
+      - name: Install MacOS dependencies
+        if: ${{ runner.os == 'macOS' }}
+        run: |
+          brew install cmake ninja libuv swig pkgconfig jsoncpp openssl libwebsockets
```
gemmellr, line 137: A little forward-looking perhaps given the job doesnt run there yet? :)

.github/workflows/build.yaml
```diff
@@ -0,0 +1,165 @@
...
+          name: Test_Results_${{matrix.os}}_${{matrix.buildType}}_${{matrix.shard}}
+          path: ${{env.DispatchBuildDir}}/Testing/**/*.xml
+      - name: Upload core files
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: cores_${{matrix.os}}_${{matrix.buildType}}_${{matrix.shard}}
+          path: |
+            **/core
+      - name: Environment (Linux)
```
gemmellr, line 162: Would this be as well done at the start / before testing, rather than the end?

jiridanek: Proton runs this at the end. It probably doesn't matter that much if it's at the end or the beginning, given that the tests will be running really fast ;)

.github/workflows/build.yaml
jiridanek: I'll drop it. I started doing both Linux and macOS, and then reduced the scope.

.github/workflows/build.yaml
gemmellr: Same comment would apply to proton if I'd thought of it at the time :) Just seems more usual to care about the env at the start rather than end, e.g all the stuff about the node etc is usually at the start.

gemmellr:
The test times look a fair bit longer in the latest version, presumably thats as much/more due to removing -j2? Did you also try doing that leaving it with 4 shards first, to see the difference each change made? Maybe using 4 (or more :o) would be better if not using the -j2.

(I dislike the complexity sharding adds, but if having it then it should at least result in much faster runs to offset that. Right now its probbly barely halving the test time, so the overall job still looks well over half the time of a basic singular run)

jiridanek:
I am now thinking of these opportunities for speedup

* fix Proton build to not build test and example programs (either there isn't an option for it, or it does not work, not sure now)
* set Dispatch build not to build the console, as that takes about 2 minutes, caching npm does not help much, and is already tested on Travis and in Console GH Action
* run the Action with all reasonable combinations of shard cound and -j, to see what the run times are
* shard the tests smarter, based on the knowledge of their past duration, with some provision not to forget running newly added tests

Having a target goal for the test runtime would help. I am now thinking of aiming for <10 minutes for the ASan instrumented run.

Using larger -j values is risky as it increases the probability of spurious failures in the tests, caused by timeouts.

Long term, it would help to have more efficient tests. A lot of what's there now is something a QE team would traditionally write; orchestrated end-to-end system tests, except it is better integrated with the build process, and therefore somewhat easier to debug than a stand-alone test suite (possibly running on multiple machines or docker containers) would be.

gemmellr:
> fix Proton build to not build test and example programs (either there isn't an option for it, or it does not work, not sure now)

Seems reasonable.

>    set Dispatch build not to build the console, as that takes about 2 minutes, caching npm does not help much, and is already tested on Travis and in Console GH Action

Yep, I had just noticed the main build was caching there, and was actually thinking of suggesting maybe the workflows should be combined. Alternatively, ensuring the build isnt duplicated would make sense if its that slow.

>    shard the tests smarter, based on the knowledge of their past duration, with some provision not to forget running newly added tests
I'd try to avoid the CI being overly smart here, it tends to lead to hassle or test holes later. I think looking at seperating overly long running tests up in the build itself, or finding if they are burning unecessary time and elmiminating [some of] it, such that there is a more uniform spread in general might make most sense.

> Having a target goal for the test runtime would help. I am now thinking of aiming for <10 minutes for the ASan instrumented run.

Hard to say. In part it depends how often folks are going to use it and actually look at it. Thus far I think most folks rely on their local testing and dont much care about the CI. Though maybe thats as it is slower. It might help if the main developers actually chipped in here.

> Using larger -j values is risky as it increases the probability of spurious failures in the tests, caused by timeouts.
Yeah, I'd avoid those personally simply due to making the simple logs unreadable :)

> edit: I like splitting the compile and test phases into separate GHA jobs. I think the complexity from that is actually outweighed by the clear separation of the two phases and the knowledge of what precisely is needed to compile, and what files are needed to run the tests, and the ability to fetch the workspace, replicate environment (There is PPA repository with Python3.6 for Focal) and run what tests I want on my machine from sources compiled in Travis.

Fair enough. I lean towards simpler CI thats easier to maintain and more like what people using the source do. I'll admit working on mostly Java stuff means I have things easier, generally having little need to consider replicating environments or using compiled-elsewhere bits.

jiridanek:
Caching npm is not helpful, but caching compiles with ccache is. The Build step in this PR takes under 2 minutes when the cache is ready.

I've checked few combinations of shard count and thread count. I certainly don't want to go higher than -j2, due to spurious failures on timeouts that happen often with more threads. Keeping the shard count low is good for ergonomic, to have small number of PR jobs. So I am going with two shards and two threads, in the end. It runs for about 12 minutes (build + the slower of the shards), which should be OK for now. The equivalent job on Travis takes 30 minutes.

Skipping Proton tests/examples won't help much, when ccache is already deployed. Skipping the Console build for ASan GH Action is helpful, it saves almost two minutes of the compile time. Partitioning the tests into shards in an intelligent manner (using knowledge of their past duration) would help a lot, but that adds complications. Maybe next time.

.github/workflows/build.yaml
jiridanek: Moved to the beginning.

jiridanek:
Table of test duration for combinations of shard count and -j count. Red numbers mean that at least one shard had at least one failed test.

image

.github/workflows/build.yaml
```diff
@@ -0,0 +1,199 @@
...
+        working-directory: ${{env.DispatchBuildDir}}
+        run: |
+          ulimit -c unlimited
+          ctest -C ${BuildType} -V -T Test --no-compress-output -I ${{matrix.shard}},,${{matrix.shards}} -j2
+
+      - name: Upload test results
+        uses: actions/upload-artifact@v2
+        if: ${{ ! cancelled() }}
+        with:
+          name: Test_Results_${{matrix.os}}_${{matrix.buildType}}_${{matrix.shard}}
```
gemmellr, line 190: If there are to be different runtimeCheck values these will clash. If there arent, maybe drop the related matrix element.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,199 @@
...
+          path: ${{env.DispatchBuildDir}}/Testing/**/*.xml
+
+      - name: Upload core files (if any)
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: cores_${{matrix.os}}_${{matrix.buildType}}_${{matrix.shard}}
```
gemmellr, line 197: Ditto.

gemmellr:
I just included the runtimeCheck value to the test names, consistent with the workspace filename used, and pushed the change.

jiridanek:
GH Action on master is done, while Travis has not even started with rat check ;) Beating Travis at speed is not much of an accomplishment, but I'm happy with the result anyways.

gemmellr:
Give it time ;)

Everyone will flock to GHA, the apache org might have not have a shared quota big enough to avoid huge projects from causing queues for the rest, and then Travis/other could become more responsive. Repeat cycle.

It happens every time something becomes the shiny new CI Env hehe. The ASF Jenkins servers long used to be uselessly overloaded - now they too easily beat Travis / Appveyor most of the time as enough bigger jobs moved away from using Jenkins to using those.

Hence why I like having CI jobs in all the envs. Invariably at any given point at least one of them is backlogged, down, or walking wounded, but others are in a better state.

PR #813, opened 2020-08-06
DISPATCH-1720, DISPATCH-1739: move console jobs, add RAT check

gemmellr:
Combines things into a single build workflow, as they console bits arent independent and the jobs always run at the same time, so its annoying moving between the two workflows to see the full build results.

Adds a RAT check job.


gemmellr:
@jiridanek ^

gemmellr:
Thats true, but they dont group on the Actions tab at all (which is how I often look at them), and they seperate when you click through into their detail from various other routes so you end up hitting back a bunch for little reason. Given you dont care, joined up it is :)

jiridanek:
The dispatch team will be surprised by what they see when they open GitHub the next time ;) Assuming they don't do all the work using the fancy Git client for Emacs, ignoring the web UI.

PR #815, opened 2020-08-06
DISPATCH-1744: refactor common HTTP code

kgiusti:
Just a simple code reorg to move some of the configuration logic that is common to both versions of HTTP.
Do not bother with reviewing any of the http1_* files - these files are fairly old.  I'll follow up with a PR as I flesh out the adaptor some.

### DIFF
--- a/python/qpid_dispatch/management/qdrouter.json
+++ b/python/qpid_dispatch/management/qdrouter.json
@@ -1103,6 +1103,16 @@
                     "required": false,
                     "description": "Name of the sslProfile..",
                     "create": true
+                },
+                "protocolVersion": {
+                    "description": "The version of the HTTP protocol supported by this listener.",
+                    "type": [
+                        "HTTP/1.x",
+                        "HTTP/2.0"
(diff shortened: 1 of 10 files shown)


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -1103,6 +1103,16 @@
...
                     "description": "Name of the sslProfile..",
                     "create": true
+                },
+                "protocolVersion": {
+                    "description": "The version of the HTTP protocol supported by this listener.",
+                    "type": [
+                        "HTTP/1.x",
+                        "HTTP/2.0"
+                    ],
+                    "default": "HTTP/2.0",
```
ganeshmurthy, line 1113: Please default to HTTP/1.x

ganeshmurthy:
I have looked at the common part of this code and it looks good to me.

kgiusti:
commit c20b29a1f70a507fd04a26b2d1692aa522fdb9b1

PR #817, opened 2020-08-07
DISPATCH-1738 - Zero out the router_id references on core and router …

ganeshmurthy:
…so they will be no use after free error

### DIFF
--- a/src/dispatch.c
+++ b/src/dispatch.c
@@ -358,19 +358,19 @@ static void qd_dispatch_set_router_area(qd_dispatch_t *qd, char *_area) {
 void qd_dispatch_free(qd_dispatch_t *qd)
 {
     if (!qd) return;
-    qd_router_free(qd->router);
-    qd_dispatch_set_router_id(qd, NULL);
-    qd_dispatch_set_router_area(qd, NULL);
     free(qd->sasl_config_path);
     free(qd->sasl_config_name);
     qd_connection_manager_free(qd->connection_manager);
     qd_policy_free(qd->policy);
     Py_XDECREF((PyObject*) qd->agent);
+    qd_router_free(qd->router);
     qd_container_free(qd->container);
     qd_server_free(qd->server);
     qd_log_finalize();
     qd_alloc_finalize();
     qd_python_finalize();
+    qd_dispatch_set_router_id(qd, NULL);
+    qd_dispatch_set_router_area(qd, NULL);
 }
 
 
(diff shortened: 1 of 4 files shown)


src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -206,6 +206,15 @@ static void qcm_mobile_sync_compose_diff_hint_list(qdrm_mobile_sync_t *msync, qd
...
 {
+    //
+    // DISPATCH-1738: If the router is shutting down, the qd_dispatch_free() in dispatch.c frees the qd->router_id and
+    // sets the core->router_id to zero. When the core is shutting down, the core->router_id will be zero
+    // and we don't want to proceed with this function. We return here in order to prevent the use after free error on
+    // msync->core->router_id down below in the function.
+    //
+    if (! msync->core->router_id)
+        return 0;
+
```
ted-ross, line 217: Would it not be better to ensure that the core thread is shut down before freeing the router_id?  If the core thread is running while basic state is being freed, there must be many other potential issues in the code.

ted-ross: Can you blank the ID and area at the end of qd_dispatch_free rather than the beginning?

ganeshmurthy: ok fair enough. I moved the code that frees the core before the router_id is freed. But I still want to keep the if (! msync->core->router_id)

ted-ross: Why?  That's what I'm trying to get rid of.

src/dispatch.c
```diff
@@ -358,14 +358,14 @@ static void qd_dispatch_set_router_area(qd_dispatch_t *qd, char *_area) {
 void qd_dispatch_free(qd_dispatch_t *qd)
 {
     if (!qd) return;
+    qd_router_free(qd->router);
```
ted-ross, line 361: You've made a significant change to the shutdown order here.  I think it would be less risky to just move the qd_dispatch_set_router_* from the beginning to the end.  Those are just freeing allocated memory slots.

src/router_core/modules/mobile_sync/mobile.c
ganeshmurthy: To future proof it if somebody tries to change the order of shutdown just like I did :)

src/dispatch.c
ganeshmurthy: ok will do

src/router_core/modules/mobile_sync/mobile.c
ted-ross: This and the one on line 399 are spurious checks that should not be added.  If our standard is to check everything that we _know_ is correct but might someday get changed, then there are about 400 additional checks that you need to put in as well.

PR #827, opened 2020-08-18
DISPATCH-1749: Improve test framwork router.wait_ready function

ChugR:
This patch adds new check for testing that a router is ready.

  * Optional. Enabled by default.
  * Locates "SERVER (notice) Operational" in a log file.
  * Fails test if log settings don't support seeing log message.
  * Fails wait_ready if message never seen.
  * Additional test runs before existing port and connector probes.

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -218,6 +218,22 @@ def wait_ports(ports, **retry_kwargs):
     for port, protocol_family in dict_iteritems(ports):
         wait_port(port=port, protocol_family=protocol_family, **retry_kwargs)
 
+def wait_server_ready(logfile, search_pattern, **retry_kwargs):
+    """Wait for up to timeout time for server log to indicate router is fully functional.
+    Takes same keyword arguments as retry to control the timeout"""
+    def peek_at_log():
+        with open(logfile, 'r') as fin:
+            lines = fin.readlines()
+        for line in lines:
+            if search_pattern in line:
+                return # success
+        raise Exception("Server not operational yet. Logfile: %s" % logfile)
+
+    try:
+        retry_exception(peek_at_log, **retry_kwargs)
+    except Exception as e:
+        raise Exception("wait_server_ready timeout waiting for server ready: %s" % (e))
+
 def message(**properties):
     """Convenience to create a proton.Message with properties set"""
     m = Message()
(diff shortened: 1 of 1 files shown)


tests/system_test.py
```diff
@@ -421,29 +437,71 @@ def attributes(e, level):
...
             return "".join(["%s {\n%s}\n"%(n, attributes(p, 1)) for n, p in self])
 
-    def __init__(self, name=None, config=Config(), pyinclude=None, wait=True,
+    def __init__(self, name=None, config=Config(), pyinclude=None, wait=True, wait_for_server_log=True,
                  perform_teardown=True, cl_args=None, expect=Process.RUNNING):
         """
         @param name: name used for for output files, default to id from config.
         @param config: router configuration
         @keyword wait: wait for router to be ready (call self.wait_ready())
+        @param wait_for_server_log Use server log check for "wait ready"
```
jiridanek, line 446:
```suggestion
        @param wait_for_server_log: use server log check for "wait ready"
```

tests/system_test.py
```diff
@@ -421,29 +437,71 @@ def attributes(e, level):
...
+        @param perform_teardown
+        @param cli_args
+        @param expect
```
jiridanek, line 449: Is this good for anything? mentioning the parameters but not giving any description?

tests/system_test.py
```diff
@@ -421,29 +437,71 @@ def attributes(e, level):
             # top level list of tuples ('section-name', dict)
             return "".join(["%s {\n%s}\n"%(n, attributes(p, 1)) for n, p in self])
 
-    def __init__(self, name=None, config=Config(), pyinclude=None, wait=True,
+    def __init__(self, name=None, config=Config(), pyinclude=None, wait=True, wait_for_server_log=True,
```
jiridanek, line 440: There are now three ways of waiting for server readiness. Wait for open port, wait for `/healtz` endpoint (if libwebsockets is enabled for compilation) or wait for special log line. It feels kind of confusing, what is the difference, when should I do what, and so on.

jiridanek:
Typo in Jira name in commit message.

tests/system_test.py
```diff
@@ -421,29 +437,71 @@ def attributes(e, level):
...
         # setup log and debug dump files
         self.dumpfile = os.path.abspath('%s-qddebug.txt' % name)
         self.config.sections('router')[0]['debugDumpFile'] = self.dumpfile
+
+        # setup default logging
+        # sense where server info logs are found
+        # fail if server polling log ready and log setup is insufficient
+        self.logfile = None
+        self.server_logfile = None
+        self.server_search_pattern = "SERVER (notice) Operational"
```
jiridanek, line 466: I searched and the `QD_LOG_NOTICE, "Operational, %d Threads Running (process ID %ld)",` line seems to be the first thing the router prints after it registers signal handlers in `static void main_process()`. The sigterm handler catches that signal and essentially converts that to a clean exit.

Makes sense. The "crash" is detected by test code if SIGTERM is delivered before dispatch sets a handler for it.

ChugR:
Thanks for the comments @jiridanek but I'm withdrawing this PR. The new code presented here is itself open to problems as yet unseen.

PR #836, opened 2020-08-27
DISPATCH-1680 add openproperties doc

pwright:
@kgiusti could you take a look please

### DIFF
--- a/docs/books/modules/user-guide/adding-metadata-to-connections.adoc
+++ b/docs/books/modules/user-guide/adding-metadata-to-connections.adoc
@@ -53,8 +53,6 @@ Note the following restrictions on the JSON entries:
  
 * ASCII characters only for keys
 * The following keys are not allowed:
-** qd.*
-** x-opt-qd.*
 ** product
 ** version
 ** failover-server-list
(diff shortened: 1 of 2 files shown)


docs/books/modules/user-guide/adding-metadata-to-connections.adoc
```diff
@@ -0,0 +1,90 @@
...
+[options="nowrap",subs="+quotes"]
+----
+connector {
+   name: broker
+   role: route-container
+   host: 127.0.0.1
+   port: 22180
+   saslMechanisms: ANONYMOUS
+   openProperties: {
+      "label": "green"
```
jiridanek, line 47: missing brace
```suggestion
      "label": "green"
  }
```

jiridanek, line 47: missing `}`
```suggestion
   openProperties: {
      "label": "green"
   }
```

pwright: ah, thanks

docs/books/modules/user-guide/adding-metadata-to-connections.adoc
```diff
@@ -0,0 +1,91 @@
...
+   }
+}
+----
+
+Note the following restrictions on the JSON entries:
+ 
+* ASCII characters only for keys
+* The following keys are not allowed:
+** qd.*
```
kgiusti, line 56: On re-reading this I'm concerned a reader may take the asterisk literally, like "avoid 'qd.*' but 'qd.foobar' is just fine."

Should probably state explicitly that keys starting with the prefix "qd." and "x-opt-qd." are reserved and should not be used.

PR #837, opened 2020-08-27
DISPATCH-1764: At boot time configure connectors and then listeners last

ChugR:
Give the router a stable initial configuration before opening any
external I/O.

### DIFF
--- a/python/qpid_dispatch_internal/management/config.py
+++ b/python/qpid_dispatch_internal/management/config.py
@@ -278,7 +278,7 @@ def configure_dispatch(dispatch, lib_handle, filename):
     dispatch = qd.qd_dispatch_p(dispatch)
     config = Config(filename)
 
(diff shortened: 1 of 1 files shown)


python/qpid_dispatch_internal/management/config.py
```diff
@@ -331,7 +331,7 @@ def configure(attributes):
                     ssl_profile_name = a.get('name')
                     displayname_service.add(ssl_profile_name, display_file_name)
 
-    # Configure remaining singleton types; don't configure listeners
+    # Configure remaining singleton types
     for e in config.entities:
         if not e['type'] == 'org.apache.qpid.dispatch.listener':
```
ted-ross, line 336: Should there be a check for connectors here as well?

ganeshmurthy (approved):

ChugR:
Fixed at Commit 913df20c

PR #843, opened 2020-09-02
DISPATCH-1762: Connector ssl config errors must prohibit connections

ChugR:
Errors seen passing ssl protocol settings to Proton were logged but
the connector was allowed to connect anyway.

The initial complaint in DISPATCH-1762 was about how a bad ca cert
file name allowed the connection to proceed without using any ca cert
file. A self test is added to check that this particular case is fixed.

This patch prevents all connector connection attempts if there are
any configuration errors reported by Proton.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1254,7 +1254,7 @@ static bool setup_ssl_sasl_and_open(qd_connection_t *ctx)
         if (config->verify_host_name) {
             if (pn_ssl_domain_set_peer_authentication(domain, PN_SSL_VERIFY_PEER_NAME, NULL)) {
                 qd_log(ct->server->log_source, QD_LOG_ERROR,
-                        "SSL peer host name configuration failed for connection [C%"PRIu64"] to %s:%s",
+                        "SSL peer host name verification configuration failed for connection [C%"PRIu64"] to %s:%s",
                         ctx->connection_id, config->host, config->port);
                 failed = true;
             }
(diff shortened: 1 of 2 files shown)


tests/system_tests_ssl.py
```diff
@@ -780,5 +781,148 @@ def test_connected_tls_sasl_routers(self):
...
+            if verified:
+                break;
+            time.sleep(sleep_time)
+        self.assertTrue(verified, "Log line containing '%s' not seen in QDR.B log" % pattern)
+
+        # Show that router A does not have router B in its network
+        router_nodes = self.get_router_nodes()
+        self.assertTrue(router_nodes)
+        node = "QDR.B"
+        self.assertTrue(node not in router_nodes, "%s should not be connected" % node)
```
ganeshmurthy, line 924: Use assertNotIn instead of assertTrue
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertNotIn
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertNotIn

src/server.c
```diff
@@ -1219,6 +1222,8 @@ static void setup_ssl_sasl_and_open(qd_connection_t *ctx)
                 qd_log(ct->server->log_source, QD_LOG_ERROR,
                        "SSL local configuration failed for %s:%s",
                        config->host, config->port);
+                pn_ssl_domain_free(domain);
+                return;
```
ganeshmurthy, line 1226: could we populate a boolean inside these if statements and check it finally and then return ? That would keep the code behaving as it was earlier.

tests/system_tests_ssl.py
ChugR: fixed

src/server.c
ChugR: yes. fixed.

ganeshmurthy (approved):

ganeshmurthy:
Chuck, this looks good. Please free to squash and commit to master. Thank you

PR #847, opened 2020-09-19
DISPATCH-1751: Rework how AMQP session incoming-window is derived

ChugR:
AMQP Open defines a 'max-frame-size'
AMQP Begin defines an 'incoming-window'

Proton allows specification of a session 'capacity'. Initially the
incoming-window will be (capacity / max-frame-size), defining at most
how many max-size transfers will definitely be accepted on the session.

Dispatch listener config defines a maxFrameSize and a maxSessionFrames
the product of which is equal to the capacity to be configured in Proton.

Dispatch vhostUserGroup policy defines a maxFrameSize and a
maxSessionWindow. The maxSessionWindow is passed directly to Proton
as the capacity.

Proton has some rules about values are allowed for max-frame-size and
capacity.

 * max-frame-size has a minimum max-value (512 octets)
 * on 32-bit systems the capacity cannot exceed 2^31-1
 * on 64-bit systems the capacity cannot exceed 2^63-1

This patch

 * Defines an AMQP_MAX_WINDOW_SIZE (2^31-1) as used by Proton.
 * Cleans up the calculation of Proton session capacity from
   the various Dispatch configuration settings.
 * If maxSessionFrames is not set, meaning unlimited frames, then
   the session capacity sent to Proton is zero. In this case Proton
   never calculates an actual incoming-window value and simply
   sends AMQP_MAX_WINDOW_SIZE as a hard-coded value for incoming-window.
   This is for both 32-bit and 64-bit systems.
 * Adjusts some self tests to allow the default settings for 32-bit and
   64-bit systems without detecting which bit width is in effect. If
   the test detects one or the other value then it calls the test a
   success.

### DIFF
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -96,7 +96,7 @@ const char * const QD_AMQPS_PORT_STR = "5671";
 
 const char * const QD_AMQP_DFLT_PROTO = "tcp";
 
-const int QD_AMQP_MAX_WINDOW_SIZE = 2147483647;
+const int PN_AMQP_MAX_SESSION_CAPACITY_32BIT = 2147483647;
 
 /// Wrapper for getservbyname/getservbyname_r macOS compatibility.
 /// Needed because getservbyname is thread safe on macOS, and getservbyname_r is not defined there.
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -96,8 +96,6 @@ const char * const QD_AMQPS_PORT_STR = "5671";
 
 const char * const QD_AMQP_DFLT_PROTO = "tcp";
 
-const int PN_AMQP_MAX_SESSION_CAPACITY_32BIT = 2147483647;
-
 /// Wrapper for getservbyname/getservbyname_r macOS compatibility.
 /// Needed because getservbyname is thread safe on macOS, and getservbyname_r is not defined there.
 static inline int qd_getservbyname(const char *name, const char *proto);
--- a/src/amqp.c
+++ b/src/amqp.c
@@ -96,6 +96,8 @@ const char * const QD_AMQPS_PORT_STR = "5671";
 
 const char * const QD_AMQP_DFLT_PROTO = "tcp";
 
+const int QD_AMQP_MAX_WINDOW_SIZE = 2147483647;
+
 /// Wrapper for getservbyname/getservbyname_r macOS compatibility.
 /// Needed because getservbyname is thread safe on macOS, and getservbyname_r is not defined there.
 static inline int qd_getservbyname(const char *name, const char *proto);
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -440,8 +440,7 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
     //
     // Given session frame count and max frame size, compute session incoming_capacity
     //   On 64-bit systems the capacity has no limit.
-    //   On 32-bit systems the largest capacity is PN_AMQP_MAX_SESSION_CAPACITY_32BIT
-    //     due to a size_t limit on the capacity setting parameter.
+    //   On 32-bit systems the largest capacity is defined as half the process address space.
     //
     if (ssn_frames != 0) {
         // Limited incoming frames.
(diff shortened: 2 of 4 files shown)


src/connection_manager.c
```diff
@@ -438,25 +438,37 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
...
-        config->incoming_capacity = (sizeof(size_t) < 8) ? 0x7FFFFFFFLL : 0x7FFFFFFFLL * config->max_frame_size;
-    else {
-        uint64_t mfs      = (uint64_t) config->max_frame_size;
-        uint64_t trial_ic = ssn_frames * mfs;
-        uint64_t limit    = (sizeof(size_t) < 8) ? (1ll << 31) - 1 : 0;
-        if (limit == 0 || trial_ic < limit) {
-            // Silently promote incoming capacity of zero to one
-            config->incoming_capacity = 
-                (trial_ic < QD_AMQP_MIN_MAX_FRAME_SIZE ? QD_AMQP_MIN_MAX_FRAME_SIZE : trial_ic);
+    if (ssn_frames == 0) {
```
ganeshmurthy, line 445: Can we remove this empty if (ssn_frames == 0) check and instead just say if (ssn_frames != 0)

ganeshmurthy: Also can we change policy.c to only call pn_session_set_incoming_capacity if the capacity is non-zero. I understand that it is okay to call pn_session_set_incoming_capacity with a zero capacity but not calling it seems clearer. Your call

gemmellr: I'd agree with the above being clearer.

src/connection_manager.c
```diff
@@ -438,25 +438,37 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
...
+        // Unlimited incoming frames.
+        // config->incoming_capacity is zero. Proton incoming_window is always AMQP_MAX_WINDOW_SIZE
+    } else {
+        // Limited incoming frames.
+        // Specify this to proton by setting capacity to be
+        // the product (max_frame_size * ssn_frames).
+
+        size_t capacity = config->max_frame_size * ssn_frames;
+        assert(capacity >= (size_t)QD_AMQP_MIN_MAX_FRAME_SIZE);
```
gemmellr, line 454: The assert seems redundant if there is any verification of the actual max frame size setting..and if there isnt, maybe there should be? The ssn_frames is known to be non zero, and if its >=1 the assert cant trigger unless the frame size is illegal.

Ah, unless the assert  is also trying to handle overflow in the capacity calculation? In which case it doesnt really do that very completely, just checks it isnt under max frame size.

gemmellr (commented):
> Defines an AMQP_MAX_WINDOW_SIZE (2^31-1) as used by Proton.

Not sure this is entirely equivalent. Proton uses that value as an effectively-unlimited number of transfer frames for the AMQP session window, if you dont set Protons 'session capacity' in bytes (and have a max frame size). The issue is the use of size_t in the API for session capacity bytes providing a limit on what can be set there for the 32bit system, and so what frames window you can get if anything is set. That value is not directly used in liminting the 'session capacity' anywhere I dont think, which is how it seems to be used here? Which I think could also be a little confusing since its clearly not the max for the 64bit case. Maybe just rename the constant to clarify a bit?

Also though, if I'm remembering the last discussion of this, the value is unsigned and so wouldnt that make the configurable limit here 2^32 -1 instead really?

>   Adjusts some self tests to allow the default settings for 32-bit and 64-bit systems without detecting which bit width is in effect. If the test detects one or the other value then it calls the test a success.

If it isnt difficult I think having the tests assert the single specific value expected would be better, otherwise this could be broken again in certain ways without detection.

src/connection_manager.c
ganeshmurthy: > Also can we change policy.c to only call pn_session_set_incoming_capacity if the capacity is non-zero. I understand that it is okay to call pn_session_set_incoming_capacity with a zero capacity but not calling it seems clearer. Your call

@ChugR what do you think of the above comment ?

src/amqp.c
```diff
@@ -96,6 +96,8 @@ const char * const QD_AMQPS_PORT_STR = "5671";
 
 const char * const QD_AMQP_DFLT_PROTO = "tcp";
 
+const int PN_AMQP_MAX_SESSION_CAPACITY_32BIT = 2147483647;
```
gemmellr, line 99: Was it checked if this 2^31-1 value is actually the the limit? I commented before that prior discussion suggested it might not be though it needed verified:

> Also though, if I'm remembering the last discussion of this, the value is unsigned and so wouldnt that make the configurable limit here 2^32 -1 instead really?

Seperately, should it be PN_<foo> if its being defined by Dispatch? It may be about a Proton setting, but its still a Dispatch constant. Also nothing else here or in the include file seems to use anything except QD_

ChugR: 2^31-1 is not a limit imposed by Proton or by AMQP. It is half of a 32-bit address space and the chances of it being reached are about zero.

I've renamed the constant and reduced its scope in another patch.

PR #853, opened 2020-09-24
WIP: DISPATCH-1783 testing safe_snprintf function

jiridanek:
My attempt to rescue code from the following commit in the following PR.
The code has diverged since, and I had to edit manually.

https://github.com/nicob87/qpid-dispatch/commit/4fc99cf08d033817c02bc83a6d74ddb8f87515af

https://github.com/jiridanek/qpid-dispatch/compare/jd_nb_2020_09_24-mocking...nicob87:4fc99cf08d033817c02bc83a6d74ddb8f87515af

https://github.com/apache/qpid-dispatch/pull/684

### DIFF
--- a/include/qpid/dispatch/macros.h
+++ b/include/qpid/dispatch/macros.h
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __macros_h__
+#define __macros_h__ 1
+
+#ifdef TESTING
+#define STATIC
+  #define INLINE
+#else
+#define STATIC static
+#define INLINE inline
+#endif
+
+#endif //__macros_h__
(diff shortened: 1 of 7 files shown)


jiridanek:
There is multiple options how to do this. I've been writing a more complete list, but I only have an example for two.

## linking magic

See the PR

## elfspy

https://github.com/mollismerx/elfspy

The code below, plus need to add a library

```
#include "qdr_doctest.h"
#include "elfspy/SPY.h"
#include "elfspy/Fake.h"

TEST_CASE("safe_snprintf_vsnprintf_failed") {
    const int   OUTPUT_SIZE = 128;
    const char *TEST_MESSAGE = "something";
    const int   LEN = strlen(TEST_MESSAGE);
    size_t len;
    char output[OUTPUT_SIZE];

    // weird elfspy boilerplate
    char* argv[2] = {(char*)"c_unittests", nullptr};
    spy::initialise(0, argv);

    // setup the fake, it will be unset when the variables go out of scope
    // it changes dynamic linking in the running process, to replace call to vsnprintf in glibc with the fake
    auto vsnprintf_ = SPY(&vsnprintf);
    auto vsnprintf_fake = spy::fake(vsnprintf_, &fake_vsnprintf);
    // alternative way for previous line:
    // auto vsnprintf_fake = spy::fake(vsnprintf_, [](auto ...) { return -1; });  // I can haz lambdaz !!!

    // run the test, simulating a failed vsnprintf
    output[0] = 'a';
    len = safe_snprintf(output, LEN+10, TEST_MESSAGE);
    CHECK(0 == len);
    CHECK('\0' == output[0]);
    CHECK("" == output);
}
```

## more possibilities

* use the `__wrap` feature in gnu linker
* libraries similar to elfspy, such as mimick, https://github.com/Snaipe/Mimick
* linking order, just define vsnprintf in the same translation unit as the test and things work out just right
* LD_PRELOAD
* ...

## general consideration

In any case, static functions cannot be mocked in C, and also inlining, when compiler decides to optimize this way, destroys the ability to replace a function. That implies that this kind of tests can be either only run in a debug build, or it has to build a second, `-testing` version of the dispatch library, for this testing. Third option is not to link to the dispatch library and `#include` the needed .c files. That also works, if the surface area tested is small. It encourages creating many smaller ctest test targets, instead of one large one.

jiridanek:
## linking order solution (first found symbol wins)

https://github.com/apache/qpid-dispatch/blob/b5deb03579a7dedd81a56f32baa3e5f4b5b57136/tests/timer_test.c#L37-L44

(this is already being used in Dispatch in one test)

jiridanek:
The decision is to pursue dynamic PLT/GOT modification approach, such as elfspy. It is self contained in the test source code, and has the least number of requirements on the compilation process. The disadvantage is that it becomes harder to ma the tests work multiplatform.

There are various other libraries besides elfspy, I examined five so far, but each seems to have at least one significant limitation.

I'll close this PR, since we chose not to use preprocessor solution prototyped here. I will put further updates on the Jira.

PR #860, opened 2020-10-05
DISPATCH-1778: Extra data included in adaptor outbound streams

ChugR:
Body_data processing failed to contain output to the current section.
This patch prevents qd_message_body_data_buffers from sending too
much data if new sections are added before the current section is
disposed.


kgiusti (approved):

ChugR:
Fixed at commit 4fe73d8

PR #866, opened 2020-10-07
ENTMQIC-2560 undeliverable-here

pwright:


pwright:
@grs  can't see to add you as reviewer, but could you take a look?

docs/books/modules/user-guide/how-routers-route-messages.adoc
```diff
@@ -64,6 +64,12 @@ Flow control is "real" in that credits flow across the link route from the recei
 * Server-side selectors
 +
 With a link route, consumers can provide server-side selectors for broker subscriptions.
+
+* Modified delivery states
```
grs, line 68: I would use something like 'consumer specific negative acknowledgement' as the 'title' here. Modified states in general are not an issue it is the consumer scoped undeliverable-here flag that is the issue.

grs (commented):
One suggested change, otherwise looks ok to me.

k-wall:
Looks reasonable to me, thanks @pwright

gemmellr:
Seems like it should either have a DISPATCH Jira reference, or not have a ref.

pwright:
closed in favor of https://github.com/apache/qpid-dispatch/pull/867

gemmellr:
No need to close PRs for such things, you can just rename it and/or force push if updating commits.

PR #871, opened 2020-10-08
NO-JIRA: Skip building Proton examples and tests in GH Action

jiridanek:
These options are enabled in PROTON-2170 and PROTON-2171.
Proton versions that don't implement this yet will ignore them and print a CMake warning.


gemmellr:
Would there be a redundant use of "-DBUILD_TESTING=OFF" with the change? It already looks to be specified in the cmake command being run using the extra args variable.

jiridanek:
I'll reformat the two long arg lines, so that I can easily see all that is passed in, for the next time. Thanks.

PR #875, opened 2020-10-13
DISPATCH-1794: Set annotation OUT pad length equal to IN strip/test length

ChugR:


kgiusti (approved):

ChugR:
Fixed at commit f60330be

PR #879, opened 2020-10-14
DISPATCH-1803: prevent body_data sections from violating Q2 limit

kgiusti:
This may be the wrong way to solve this - fixing it at the message construction phase rather than at message decode.

This patch adds a new body-data message API call:
void qd_message_body_data_append(qd_message_t *msg, qd_buffer_list_t *data);

A single call to this would replace this logic in the adaptors:

-    //                                                                                                                                           
-    // Compose a DATA performative for this section of the stream                                                                                
-    //                                                                                                                                           
-    qd_composed_field_t *field = qd_compose(QD_PERFORMATIVE_BODY_DATA, 0);                                                                       
-    qd_compose_insert_binary_buffers(field, body);                                                                                               
-                                                                                                                                                 
-    //                                                                                                                                           
-    // Extend the streaming message and free the composed field                                                                                  
-    //                                                                                                                                           
-    qd_message_extend(msg, field);                                                                                                               
-    qd_compose_free(field);

### DIFF
--- a/include/qpid/dispatch/message.h
+++ b/include/qpid/dispatch/message.h
@@ -378,6 +378,20 @@ typedef enum {
 qd_message_body_data_result_t qd_message_next_body_data(qd_message_t *msg, qd_message_body_data_t **body_data);
 
 
+/**
+ * qd_message_body_data_append
+ *
+ * Append the buffers in data as a sequence of one or more BODY_DATA sections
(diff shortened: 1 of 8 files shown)


ChugR (approved):
All adaptors should be required to use this interface.

kgiusti:
Merged to dev-protocol-adaptors-2 branch

PR #885, opened 2020-10-18
DISPATCH-1750 Update all console deps except Patternfly and d3

jiridanek:
I've manually tried using the console and I did not see anything broken.

image

### DIFF
--- a/console/react/package-lock.json
+++ b/console/react/package-lock.json
@@ -14694,9 +14694,9 @@
       }
     },
     "node-forge": {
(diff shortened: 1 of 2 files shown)


ErnieAllen (approved):
As long as the console tests pass, and a manual viewing of the console looks good, I'm good with this. Thanks!

jiridanek:
Thanks. I should probably stop messing with the easy updates and do one of the two hard ones, at this point...

PR #890, opened 2020-10-22
DISPATCH-1751: Rework how AMQP session incoming-window is derived

ChugR:
Proton allows specification of a session 'capacity'. Initially the
incoming-window will be (capacity / max-frame-size), defining at most
how many max-size transfers will definitely be accepted on the session.

Dispatch listener config defines a maxFrameSize and a maxSessionFrames
the product of which is equal to the capacity to be configured in Proton.

Dispatch vhostUserGroup policy defines a maxFrameSize and a
maxSessionWindow. The maxSessionWindow is passed directly to Proton
as the capacity.

The interface to Proton defines capacity with type 'size_t'. It is possible
to have a (maxFrameSize * maxSessionFrames) product be larger than
that allowed by a size_t on a 32-bit system. Perform the capacity
calculations using 64-bit integers before applying the results and
sending them to Proton.

### DIFF
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -438,25 +438,35 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
         config->max_frame_size = QD_AMQP_MIN_MAX_FRAME_SIZE;
 
     //
-    // Given session frame count and max frame size compute session incoming_capacity
+    // Given session frame count and max frame size, compute session incoming_capacity
+    //   On 64-bit systems the capacity has no practial limit.
+    //   On 32-bit systems the largest default capacity is half the process address space.
     //
-    if (ssn_frames == 0)
-        config->incoming_capacity = (sizeof(size_t) < 8) ? 0x7FFFFFFFLL : 0x7FFFFFFFLL * config->max_frame_size;
-    else {
-        uint64_t mfs      = (uint64_t) config->max_frame_size;
-        uint64_t trial_ic = ssn_frames * mfs;
-        uint64_t limit    = (sizeof(size_t) < 8) ? (1ll << 31) - 1 : 0;
-        if (limit == 0 || trial_ic < limit) {
-            // Silently promote incoming capacity of zero to one
-            config->incoming_capacity = 
-                (trial_ic < QD_AMQP_MIN_MAX_FRAME_SIZE ? QD_AMQP_MIN_MAX_FRAME_SIZE : trial_ic);
+    bool is_64bit = sizeof(size_t) == 8;
+#define MAX_32BIT_CAPACITY ((size_t)(2147483647))
+    if (ssn_frames == 0) {
+        config->incoming_capacity = is_64bit ? MAX_32BIT_CAPACITY * (size_t)config->max_frame_size : MAX_32BIT_CAPACITY;
+    } else {
+        // Limited incoming frames.
+        if (is_64bit) {
+            // Specify this to proton by setting capacity to be
+            // the product (max_frame_size * ssn_frames).
+            config->incoming_capacity = (size_t)config->max_frame_size * (size_t)ssn_frames;
         } else {
-            config->incoming_capacity = limit;
-            uint64_t computed_ssn_frames = limit / mfs;
-            qd_log(qd->connection_manager->log_source, QD_LOG_WARNING,
-                   "Server configuation for I/O adapter entity name:'%s', host:'%s', port:'%s', "
-                   "requested maxSessionFrames truncated from %"PRId64" to %"PRId64,
-                   config->name, config->host, config->port, ssn_frames, computed_ssn_frames);
+            // 32-bit systems have an upper bound to the capacity
+            uint64_t max_32bit_capacity = (uint64_t)MAX_32BIT_CAPACITY;
+            uint64_t capacity     = (uint64_t)config->max_frame_size * (uint64_t)ssn_frames;
+            if (capacity <= max_32bit_capacity) {
+                config->incoming_capacity = (size_t)capacity;
+            } else {
+                config->incoming_capacity = MAX_32BIT_CAPACITY;
+                uint64_t actual_frames = max_32bit_capacity / (uint64_t)config->max_frame_size;
+
+                qd_log(qd->connection_manager->log_source, QD_LOG_WARNING,
+                    "Server configuation for I/O adapter entity name:'%s', host:'%s', port:'%s', "
+                    "requested maxSessionFrames truncated from %"PRId64" to %"PRId64,
+                    config->name, config->host, config->port, ssn_frames, actual_frames);
+            }
         }
     }
 
--- a/tests/system_tests_protocol_settings.py
+++ b/tests/system_tests_protocol_settings.py
@@ -26,6 +26,7 @@ from system_test import TestCase, Qdrouterd, main_module
 from system_test import unittest
 from proton.utils import BlockingConnection
 import subprocess
+import sys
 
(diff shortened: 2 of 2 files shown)


gemmellr:
>Dispatch listener config defines a maxFrameSize and a maxSessionFrames
the product of which is equal to the capacity to be configured in Proton.

> Dispatch vhostUserGroup policy defines a maxFrameSize and a
maxSessionWindow. The maxSessionWindow is passed directly to Proton
as the capacity.

Why do they differ? How do they interact? Does maxSessionWindow say default to 0, and overrides the value calculated by the other stuff if set?

Aside: what would it do if on a 32bit system the maxSessionWindow was configured above the '32bit cap' now calculated by the other bits, is it truncated before it is 'passed directly' (say giving similar effect to the overflow in calcs), or does the config cause a bang?

src/connection_manager.c
```diff
@@ -438,25 +438,35 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
...
-        uint64_t trial_ic = ssn_frames * mfs;
-        uint64_t limit    = (sizeof(size_t) < 8) ? (1ll << 31) - 1 : 0;
-        if (limit == 0 || trial_ic < limit) {
-            // Silently promote incoming capacity of zero to one
-            config->incoming_capacity = 
-                (trial_ic < QD_AMQP_MIN_MAX_FRAME_SIZE ? QD_AMQP_MIN_MAX_FRAME_SIZE : trial_ic);
+    bool is_64bit = sizeof(size_t) == 8;
+#define MAX_32BIT_CAPACITY ((size_t)(2147483647))
+    if (ssn_frames == 0) {
+        config->incoming_capacity = is_64bit ? MAX_32BIT_CAPACITY * (size_t)config->max_frame_size : MAX_32BIT_CAPACITY;
```
gemmellr, line 448: I dont really understand why this version goes back to always setting a capacity for the 64bit system, even in the default 'not really doing flow control' situation? Doing so causes the window to be precisely recomputed with every flow frame sent again, as it was to begin with. Wasnt that one of the things we wanted to stop with these changes?

Specifically calculating a ~32TB default capacity (when considering the default 16KB frames size) as this does for the 64bit system, just to have setting this session capacity value then cause proton to [repeatedly] calculate the the same peak value for the window as it it would simply used were nothing set, seems as odd now as it did to start with.

Capping the 32bit system at half its address space by default can certainly be argued given its the same limit applied by Dispatch in the calculations based on user-configuration. Plus its more likely to occur being that it is only 2GB. Though practicaly im not sure leaving it uncapped by default would be that relatively different from the 64bit case being given 32TB, neither is that likely to cope with such a window being filled (or there being other sessions each with their own limit..).

ChugR: > Why do they differ? How do they interact? Does maxSessionWindow say default to 0, and overrides the value calculated by the other stuff if set?
> 
> Aside: what would it do if on a 32bit system the maxSessionWindow was configured above the '32bit cap' now calculated by the other bits, is it truncated before it is 'passed directly' (say giving similar effect to the overflow in calcs), or does the config cause a bang?
> 
They differ and fixing that is not the goal of this jira. If that's a problem then please file a new jira addressing that defect.

Policy configuration overrides in configuration of listeners and connectors.

32-bit systems can't set int variables bigger than 2Gb.

ChugR: > I dont really understand why this version goes back to always setting a capacity for the 64bit system, even in the default 'not really doing flow control' situation? Doing so causes the window to be precisely recomputed with every flow frame sent again, as it was to begin with. Wasnt that one of the things we wanted to stop with these changes?
>
But the window is computed to precisely *different* values as the window is slowly consumed. Showing precisely the same value gives no indication of how many transfers are in flight. Having proton show the precise, *correct* value is not a big hardship: a subtraction and division per window calculation.

> Specifically calculating a ~32TB default capacity (when considering the default 16KB frames size) as this does for the 64bit system, just to have setting this session capacity value then cause proton to [repeatedly] calculate the the same peak value for the window as it it would simply used were nothing set, seems as odd now as it did to start with.
> 

Forcing Proton to use the values we expect it to use allows test code to prove that those values are actually being used.

> Capping the 32bit system at half its address space by default can certainly be argued given its the same limit applied by Dispatch in the calculations based on user-configuration. Plus its more likely to occur being that it is only 2GB. Though practicaly im not sure leaving it uncapped by default would be that relatively different from the 64bit case being given 32TB, neither is that likely to cope with such a window being filled (or there being other sessions each with their own limit..).
> 

AMQP protocol allows a max-frame-size in bytes and an incoming-window in frames. Proton allows a max-frame-size in bytes and a capacity in bytes. Furthermore, the Proton capacity is a size_t type which varies in its maximum values between 32-bit and 64-bit systems.

DIspatch listeners and connectors allowed maxFrameSize and maxSessionFrames (incoming-window) settings to directly match the AMQP protocol. If Proton offered those same two settings then this issue between 32-bit and 64-bit systems would never have happened. 

I propose that we commit this patch and move on.

gemmellr: > 
> 
> > I dont really understand why this version goes back to always setting a capacity for the 64bit system, even in the default 'not really doing flow control' situation? Doing so causes the window to be precisely recomputed with every flow frame sent again, as it was to begin with. Wasnt that one of the things we wanted to stop with these changes?
> 
> But the window is computed to precisely _different_ values as the window is slowly consumed. Showing precisely the same value gives no indication of how many transfers are in flight. Having proton show the precise, _correct_ value is not a big hardship: a subtraction and division per window calculation.
> 

No, it isnt a great hardship, I just wondered why this version went back to doing it, when the previous version expressly went the opposite way, is all.

Though note the value doesnt really demonstrate how many transfer frames are in flight at all in Protons case. It allows seeing how many max-size frames it would take to cover all the data recieved but not yet read out of all the receivers on the session combined at that point.  The number of frames receieved and their effect on the window size become largely decoupled from that (unlesss essentially all arriving frames are max-size too). You can reason about the size of combined data yet to be read, though without indication where it is to be read from / what frames its for (except the single receiver case to an extent).

> > Specifically calculating a ~32TB default capacity (when considering the default 16KB frames size) as this does for the 64bit system, just to have setting this session capacity value then cause proton to [repeatedly] calculate the the same peak value for the window as it it would simply used were nothing set, seems as odd now as it did to start with.
> 
> Forcing Proton to use the values we expect it to use allows test code to prove that those values are actually being used.
> 

Its provoking the same [initial] window as it would get from proton if nothing were set, so its not really proving that.

> > Capping the 32bit system at half its address space by default can certainly be argued given its the same limit applied by Dispatch in the calculations based on user-configuration. Plus its more likely to occur being that it is only 2GB. Though practicaly im not sure leaving it uncapped by default would be that relatively different from the 64bit case being given 32TB, neither is that likely to cope with such a window being filled (or there being other sessions each with their own limit..).
> 
> AMQP protocol allows a max-frame-size in bytes and an incoming-window in frames. Proton allows a max-frame-size in bytes and a capacity in bytes. Furthermore, the Proton capacity is a size_t type which varies in its maximum values between 32-bit and 64-bit systems.
> 
> DIspatch listeners and connectors allowed maxFrameSize and maxSessionFrames (incoming-window) settings to directly match the AMQP protocol. If Proton offered those same two settings then this issue between 32-bit and 64-bit systems would never have happened.
> 

I understand all that already, especially given I largely explained the original issue, it doesnt really relate to what I was aiming at with my comment, however at this point...

> I propose that we commit this patch and move on.

I can tell you are fed up with the matter, and I'm likely more fed up discussing it, so by all means.

PR #895, opened 2020-10-26
DISPATCH-1814 Autofix for flake8 E701 multiple statements on a single line

jiridanek:


jiridanek:
I read through the diff and it looks ok. In the future, it might be reasonable to let the tool go through all whitespace changes in one go; autopep8 should be trustworthy enough to simply do that.

There is one problem with merging these changes now; the feature branch for http2 support. Having these formatting changes on master would make it harder to merge that branch later.

jiridanek:
As per discussion on users@qpid, this is to contain all autofixes in single commit, and it is to be merged soon after a completed release.

PR #896, opened 2020-10-27
NO-JIRA fix usage of self.assertTrue in policy tests

jiridanek:
> Expression "(self.assertTrue(success), 'Expected router to generate close with condition: message size exceeded')" is assigned to nothing
>> https://www.deepcode.ai/app/gh/jiridanek/qpid-dispatch/5371dfbedc60fd6c429c237d7053c3e15f0d25fe/_/%2Ftests%2Fsystem_tests_policy_oversize_compound.py/python%2Fpylint%2FW0106/811/code


ChugR (approved):
How did the old code pass tox / compile / run without complaining? Hmmm.

jiridanek:
That's easy. Tox only runs flake8, which mostly checks formatting, it does not look into semantics too much; and most flake8 checks are currently marked as ignored in dispatch tox.ini anyways. The only situation when anything behaves out of place is if the assertion fails, and the exception is then missing the custom message; which is hard to spot at that time, anyways.

There's probably more code like that, also in the web console, but its going to be all low impact, most likely.

PR #897, opened 2020-10-27
DISPATCH-1815: Send PING frame every 4 seconds on an egress connection

ganeshmurthy:


ganeshmurthy:
@grs please take a look and comment and/or approve

src/adaptors/http2/http2_adaptor.h
```diff
@@ -138,6 +141,7 @@ struct qdr_http2_connection_t {
     bool                      ingress;
     bool                      timer_scheduled;
     bool                      client_magic_sent;
+    time_t                    prev_ping; // Time the previous PING frame was sent on egress connection.
```
grs, line 144: Strictly this could be the last time for any write to the socket, not just a ping. The time could then be updated on every write.

grs (approved):

src/adaptors/http2/http2_adaptor.h
ganeshmurthy: The check_send_ping_frame() only conditionally updates the conn->prev_ping field like this - 



    time_t current = time(NULL);
    time_t prev = conn->prev_ping;
    if (current - prev >= 4) {
        send_ping_frame(conn);
        qd_log(http2_adaptor->log_source, QD_LOG_TRACE, "[C%"PRIu64"] Sent PING frame", conn->conn_id);
        qd_timer_schedule(conn->ping_timer, 4000);
        conn->prev_ping = current;
    }

So the conn->prev_ping contains the time that the ping was sent out ?

grs: Right, and that is fine. All I'm saying is that if you send out any other frame within the second, there is no need to send a ping. However that is an optimisation. It is good as is.

ganeshmurthy: Ah, yes, agreed. Good point.

PR #905, opened 2020-10-29
DISPATCH-1807: Add self tests for tcp protocol adaptor

ChugR:
* echo_server and echo_client modified to have a test class
  that runs the test proper.

* tcp_adaptor test runs the client and server test classes in
  separate threads and not in separate processes.

The test is expected to try several clients with different sizes but there's a timeout issue with that.
Running this test triggers DISPATCH-1817, a structure leak, which makes the test fail. The TCP adaptor and the test clients succeed but the final ctest result is a disappointment.

Commit this PR after the leaks are fixed.

### DIFF
--- a/tests/TCP_echo_client.py
+++ b/tests/TCP_echo_client.py
@@ -19,25 +19,33 @@
 # under the License.
 #
 
+from __future__ import unicode_literals
+from __future__ import division
+from __future__ import absolute_import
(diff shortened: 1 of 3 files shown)


kgiusti (approved):

ChugR:
Fixed

PR #910, opened 2020-11-12
DISPATCH-1831: Tcp test - extend backbone to three routers

ChugR:
This patch is a straightforward addition of a third backbone interior router to the TCP self test.
On my system this patch reveals a curious side effect:

- Without this patch 'test_03' fails with a timeout.
- With this patch 'test_03' succeeds.

test_03 uses a TCP listener on INTA and a TCP connector on INTB. The test fails in the A-B backbone and the test passes in the A-B-C backbone.

In the success case: right after the client attaches to INTA TCP listener router INTA opens a link to INTB and sends a TCP INTB@mySite message to INTB. Then INTB opens the connection to the server. 

In the fail case the router-data link never gets opened.

Scraper creates a detailed narrative of the test. It combines logs from all the routers, six echo servers, and as many echo clients as the test cases run. View it in a web browser by running a script generated by the test.

- cd build
- ctest -VV -R tcp_adaptor
- cd ./tests/system_test.dir/system_tests_tcp_adaptor/TcpAdaptor/setUpClass
- bash -x TcpAdaptor-run-scraper.sh
- firefox TcpAdaptor.html

### DIFF
--- a/tests/system_tests_tcp_adaptor.py
+++ b/tests/system_tests_tcp_adaptor.py
@@ -107,36 +107,41 @@ class Logger():
 
 class TcpAdaptor(TestCase):
     """
-    4 edge routers connected via 2 interior routers.
-    6 echo servers are connected via tcpConnector, one to each router.
-    Each router has 7 listeners, one for each server and
+    6 edge routers connected via 3 interior routers.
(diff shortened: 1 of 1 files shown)


ChugR:
Fixed at commit 1e2bf

PR #915, opened 2020-11-17
DISPATCH-1842: Handle managament deletes of httpConnector by deleting…

ganeshmurthy:
… connection objects associated with the connector. This prevents the router crash from happening


kgiusti (commented):
LGTM

PR #917, opened 2020-11-18
DISPATCH-1816: avoid race between conn disconnect and activation

kgiusti:


kgiusti:
dev-protocol-adaptors-2 commit 7529eb2d01de6dcf3c228a8dbd3927df342e1807

PR #928, opened 2020-11-20
DISPATCH-1851: TCP adaptor - add more connection and link IDs to log output

ChugR:
Also identify callbacks that are ignored as NOOPs.

The connection and link IDs in the log output are really helpful.

Adding asserts reveals that function qdr_tcp_activate is called routinely with no context. Is that correct?


ChugR:
Fixed at Commit 648ff4c

PR #931, opened 2020-11-25
DISPATCH-1846: Debugging the A-B-C hang

ChugR:
This patch is not meant to be merged. DEBUG ONLY!

Refresh: A TCP listener on INTA can't connect to a TCP server
on INTC. INTA-INTA works, and INTA-INTB works.

The TCP self test is trimmed to just the three interior routers.
Then bunches of log statements are sprinkled around to track what
the routers are doing.

Note that log statements in router-proper code are from log
source 'TCP_TEST' at critical level. They appear in Scraper logs
as coming from container 'TCP_TEST'.  This is all just for debug.

### DIFF
--- a/src/router_core/transfer.c
+++ b/src/router_core/transfer.c
@@ -477,6 +477,13 @@ static void qdr_link_forward_CT(qdr_core_t *core, qdr_link_t *link, qdr_delivery
     if (!dlv_link)
         return;
 
+    if (addr) {
+        const char *key = (const char*) qd_hash_key_by_handle(addr->hash_handle);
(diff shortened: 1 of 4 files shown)


ChugR:
The test(s) work if the echo client stalls a bit more before sending any traffic to be echoed.

-           time.sleep(0.1)  # delete

+          time.sleep(0.5)  # add

The adaptor appears to be allowing the echo client to send to the router network before the replyTo address has stabilized.

ChugR:
Fixed under PR-934

PR #934, opened 2020-12-02
DISPATCH-1846: TCP adaptor: fix test stall

ChugR:
* Wait until all echo server addresses are known to interior routers.
* Add 'ES_' prefix to server addresses. Having the mobile address
  the same as the router name worked but was confusing when viewed
  by qdstat.
* Turn off echo server logging.


ChugR:
Fixed at commit b182b2

PR #936, opened 2020-12-08
DISPATCH-1874: Free the unused buffers on a stream that has been forc…

ganeshmurthy:
…e closed. Also free unused compose objects

### DIFF
--- a/src/adaptors/http2/http2_adaptor.c
+++ b/src/adaptors/http2/http2_adaptor.c
@@ -1634,6 +1634,11 @@ uint64_t handle_outgoing_http(qdr_http2_stream_data_t *stream_data)
                                                             count,
                                                             stream_data);
 
+            for (uint32_t idx = 0; idx < count; idx++) {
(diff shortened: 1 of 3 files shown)


src/adaptors/http2/http2_adaptor.c
```diff
@@ -393,6 +395,10 @@ static qdr_http2_stream_data_t *create_http2_stream_data(qdr_http2_session_data_
     qd_message_set_stream_annotation(stream_data->message, true);
     stream_data->session_data = session_data;
     stream_data->app_properties = qd_compose(QD_PERFORMATIVE_APPLICATION_PROPERTIES, 0);
+    if (!stream_data->app_properties) {
+        qd_compose_free(stream_data->app_properties);
```
kgiusti, line 399: wait, what? check for null, free null?

ganeshmurthy: Bad code removed and pushed up

PR #937, opened 2020-12-08
DISPATCH-1867: Added a lock to fix the race around http adaptor creat…

ganeshmurthy:
…ion/initialization

### DIFF
--- a/src/adaptors/http2/http2_adaptor.c
+++ b/src/adaptors/http2/http2_adaptor.c
@@ -65,9 +65,10 @@ typedef struct qdr_http2_adaptor_t {
 } qdr_http2_adaptor_t;
 
 
-static qdr_http2_adaptor_t *http2_adaptor;
+static qdr_http2_adaptor_t *http2_adaptor = 0;
 
 static void handle_connection_event(pn_event_t *e, qd_server_t *qd_server, void *context);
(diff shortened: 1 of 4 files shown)


src/adaptors/http2/http2_adaptor.c
```diff
@@ -2482,49 +2505,56 @@ static void qdr_http2_adaptor_final(void *adaptor_context)
...
-                                            qdr_http_conn_close,
-                                            qdr_http_conn_trace);
-    adaptor->log_source = qd_log_source(QD_HTTP_LOG_SOURCE);
-    adaptor->protocol_log_source = qd_log_source("PROTOCOL");
-    adaptor->lock = sys_mutex();
-    *adaptor_context = adaptor;
-    DEQ_INIT(adaptor->listeners);
-    DEQ_INIT(adaptor->connectors);
-    DEQ_INIT(adaptor->connections);
+    sys_mutex_lock(core->adaptor_startup_lock);
```
fgiorgetti, line 2508: @ganeshmurthy this is going to cause a deadlock right?
This PR is not causing that crash, but it is hanging during router initialization.

ganeshmurthy: Put the lock in the wrong place. Moved it up. Please test now. Thanks

fgiorgetti: @ganeshmurthy worked great this time!

PR #941, opened 2020-12-09
DISPATCH-1744: bugfix - treat 1xx responses as incomplete

kgiusti:
bugfix: force error disposition on request when cancelled


kgiusti:
Abandoned

PR #943, opened 2020-12-11
DISPATCH-1879: Fixed use after free issue with logger trying to acces…

ganeshmurthy:
…s freed memory. Also fixed nghttp2 session memory leak


src/adaptors/http2/http2_adaptor.c
```diff
@@ -2242,7 +2243,8 @@ static void handle_connection_event(pn_event_t *e, qd_server_t *qd_server, void
             send_settings_frame(conn);
             qd_log(log, QD_LOG_INFO, "[C%"PRIu64"] Accepted Ingress ((PN_RAW_CONNECTION_CONNECTED)) from %s", conn->conn_id, conn->remote_address);
         } else {
-            nghttp2_session_client_new(&conn->session_data->session, (nghttp2_session_callbacks*)http2_adaptor->callbacks, (void *)conn);
+            if (!conn->session_data->session)
+                nghttp2_session_client_new(&conn->session_data->session, (nghttp2_session_callbacks*)http2_adaptor->callbacks, (void *)conn);
```
jiridanek, line 2247: style: either there should be space before * in both casts, or there shouldn't be

PR #945, opened 2020-12-13
DISPATCH-1881 Fix build dependencies for http2 protocol adaptor

jiridanek:


jiridanek:
CI looks ok; the expected failures are present, and there is nothing new.

ganeshmurthy (approved):

jiridanek:
All right, let's do it like this for now. I am still thinking that http2 should be an optional feature, but that could be a matter for a subsequent discussion and eventual PR.

Requiring at least Proton 0.33.0 probably has to remain mandatory, because trying to make it optional would complicate things too much.

PR #946, opened 2020-12-13
DISPATCH-1881 install deps for previously skipped tests in GitHub Action

jiridanek:


jiridanek:
I've checked GitHub Action logs and saw no conditionally skipped tests. Looks like its working. Will merge.

PR #949, opened 2020-12-14
DISPATCH-1883: Install selectors to run tcp echo server/client tools

ChugR:


ganeshmurthy (approved):

.travis.yml
```diff
@@ -69,6 +69,8 @@ jobs:
     - python -m pip install --user --upgrade tox virtualenv==20.0.23
     # Install quart to run the http2 tests.
     - python -m pip install --user quart
+    # DISPATCH-1883: Install selectors to run tcp echo server/client tools
+    - python -m pip install --user selectors
```
jiridanek, line 73: is this something that the GitHub Action also needs?

ganeshmurthy:
@ChugR  can you also please add that details about installing selectors to https://github.com/apache/qpid-dispatch/blob/master/README#L81 ?

.travis.yml
ChugR: yes. well spotted.

PR #960, opened 2020-12-16
DISPATCH-1886 - Close race window on connection activate/close.

ted-ross:


ganeshmurthy:
This looks good to me. Every call to qd_server_activate(conn); must be protected by the server->conn_activation_lock. 
1. The call to qd_server_activate in CORE_connection_activate() (in router_node.c) is already protected by this lock.
2. This PR removes qd_link_activate()
3. This PR introduces the lock around qd_server_activate call in qd_connection_invoke_deferred (in server.c)

ganeshmurthy (approved):

ganeshmurthy:
Just a quick note that the server->conn_activation_lock was introduced as a fix to https://issues.apache.org/jira/browse/DISPATCH-1417

PR #961, opened 2020-12-16
DISPATCH-1780: initial support for aggregated multicast

grs:


src/adaptors/http1/http1_server.c
```diff
@@ -70,6 +70,7 @@ typedef struct _server_request_t {
     bool            request_settled; // set by adaptor
     bool            request_acked;   // true if dispo sent to core
     bool            headers_encoded; // True when header encode done
+    bool            response_settled;
```
kgiusti, line 73: Sorry - late night but for the life of me I don't see where this flag is set!  I'd assume it would be set in qdr_http1_server_core_delivery_update when the outcome for the response arrives...

src/adaptors/http1/http1_client.c
```diff
@@ -1238,6 +1505,24 @@ uint64_t qdr_http1_client_core_link_deliver(qdr_http1_adaptor_t    *adaptor,
     _client_response_msg_t *rmsg = DEQ_TAIL(hreq->responses);
     assert(rmsg && rmsg->dlv == delivery);
 
+    // when aggregating responses, they are saved on the list until
+    // the request has been settled, then encoded in the configured
+    // aggregation format
+    if (hconn->cfg.aggregation != QD_AGGREGATION_NONE) {
+        qd_log(qdr_http1_adaptor->log, QD_LOG_DEBUG, "[C%"PRIu64"][L%"PRIu64"] Received response (%i responses received), settling", hconn->conn_id, link->identity, DEQ_SIZE(hreq->responses));
+        rmsg->dispo = PN_ACCEPTED;
+        qd_message_set_send_complete(msg);
```
kgiusti, line 1514: Should this be predicated on qd_message_receive_complete(msg)?  I assume responses may be fragmented and we want to avoid updating the outcome until it is done arriving.

src/adaptors/http1/http1_client.c
```diff
@@ -1238,6 +1505,24 @@ uint64_t qdr_http1_client_core_link_deliver(qdr_http1_adaptor_t    *adaptor,
...
+        qdr_link_flow(qdr_http1_adaptor->core, link, 1, false);
+        qdr_delivery_remote_state_updated(qdr_http1_adaptor->core,
+                                          rmsg->dlv,
+                                          rmsg->dispo,
+                                          true,   // settled,
+                                          0,      // error
+                                          0,      // dispo data
+                                          false);
+        return PN_ACCEPTED;
```
kgiusti, line 1523: Since we are holding the response message buffers, I believe it may be necessary to disable Q2 on the arriving response messages by calling qd_message_Q2_holdoff_disable(). and possibly "restart" the receiving thread via qd_link_restart_rx(qd_message_get_receiving_link(msg)).  

See https://github.com/apache/qpid-dispatch/blob/master/src/message.c#L2578  - something like that. 

This could be tested by having the multicast responses contain > 512K bodies.

Oh what a tangled web we weave...

src/adaptors/http1/http1_codec.c
```diff
@@ -181,6 +184,8 @@ struct h1_codec_connection_t {
         bool is_request;
         bool is_chunked;
 
+        char *boundary_marker;//used for multipart content
```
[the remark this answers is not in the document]
grs: It is freed in h1_codec_tx_end_multipart, which I think covers the normal path. However I agree logically it seems like it should be also freed in reset if needed and that may indeed be required for some error paths.

src/adaptors/http1/http1_server.c
grs: Agreed I have removed that.

src/adaptors/http1/http1_client.c
grs: Probably. I have added that check.

src/adaptors/http1/http1_client.c
grs: I did test with large responses (1M) and it seems to work so while you may be right that something is needed here, I've left that for now.

grs:
Running locally the tests all passed for me.

src/adaptors/http1/http1_client.c
kgiusti: Yes let's not block this patch for that - we'll add a test case for it in a follow up.
We'll need a multi-hop router test case since Q2 is a proton layer thing.  I'd expect Q2 to not be a factor in router standalone configurations.

kgiusti (approved):
Let's merge this now - the errors posted in the CI are known issues I'm working on.

PR #965, opened 2020-12-21
DISPATCH-1900 calling calloc with zero nmemb or zero size is permitted

jiridanek:
Let's see what Travis thinks about this.

### DIFF
--- a/include/qpid/dispatch/ctools.h
+++ b/include/qpid/dispatch/ctools.h
@@ -237,7 +237,7 @@ static inline void *qd_malloc(size_t size)
 static inline void *qd_calloc(size_t nmemb, size_t size)
 {
     void *ptr = calloc(nmemb, size);
-    if (!ptr) {
+    if (!ptr && nmemb && size) {
         perror("qd_calloc");
         abort();
     }
--- a/include/qpid/dispatch/ctools.h
+++ b/include/qpid/dispatch/ctools.h
@@ -236,7 +236,6 @@ static inline void *qd_malloc(size_t size)
 
 static inline void *qd_calloc(size_t nmemb, size_t size)
 {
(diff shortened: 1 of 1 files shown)


include/qpid/dispatch/ctools.h
```diff
@@ -236,7 +236,6 @@ static inline void *qd_malloc(size_t size)
 
 static inline void *qd_calloc(size_t nmemb, size_t size)
 {
-    assert(nmemb && size);
     void *ptr = calloc(nmemb, size);
     if (!ptr) {
```
ted-ross, line 240: The null-pointer check is not valid in this case.  The man page for calloc states that the returned value may or may not be 0 when one of the input arguments is 0.  I strongly suspect that it is non-zero in this case for Fedora.  I would be inclined to leave the assertion in place, since we shouldn't be calling qd_calloc with either a zero nmemb or a zero size.

ted-ross (changes requested):

include/qpid/dispatch/ctools.h
jiridanek: Right. I forgot that `free(NULL)` (which is going to happen later, if `calloc` returned NULL) is perfectly valid. Anyways, in this case the return is non-zero on Ubuntu as well, so Travis is passing with this.

Personally, I think that allocating zero-sized arrays and then looping over them is kind of neat, and I would leave that in place (and fix the check here).

I fix the PR; if you want to resolve the issue differently, then feel free to reject the PR.

jiridanek: (I'd maybe disallow `size==0`, because that sounds unlikely to be intentional)

ted-ross (approved):
This covers all the cases.  I will merge it.

PR #966, opened 2020-12-22
DISPATCH-1901: Allow any proton:io reason failing connection to bad host name

ChugR:
The system_tests_bad_configuration test sets up a connector to a
host that is never reachable. The test expects the connection to fail.

On Linux proton normally returns 'Name or service not known'. On macOS
there is some other specific reason.

Another class of errors shows up when DNS is unavailable. On Linux the
reason is 'Temporary failure in name resolution'.

This patch stops looking for specific text from the proton:io error
messages. The test declares success when a qpid-dispatch log message
indicates a connection failure to the host in question.

### DIFF
--- a/tests/system_tests_bad_configuration.py
+++ b/tests/system_tests_bad_configuration.py
@@ -112,12 +112,8 @@ class RouterTestBadConfiguration(TestCase):
         """
         with open('../setUpClass/test-router.log', 'r') as router_log:
             log_lines = router_log.read().split("\n")
-            expected_errors = [
-                "Connection to %s" % self.unresolvable_host_name
-            ]
-            errors_caught = [line for line in log_lines
-                             if any([expected_error in line for expected_error in expected_errors])
-                             and "failed" in line]
+            expected_log_snip = "Connection to %s" % self.unresolvable_host_name
+            errors_caught = [line for line in log_lines if expected_log_snip in line and "failed" in line]
 
             self.error_caught = any(errors_caught)
 
--- a/tests/system_tests_bad_configuration.py
+++ b/tests/system_tests_bad_configuration.py
@@ -51,12 +51,13 @@ class RouterTestBadConfiguration(TestCase):
         """
         super(RouterTestBadConfiguration, cls).setUpClass()
         cls.name = "test-router"
+        cls.unresolvable_host_name = 'unresolvable.host.name'
         cls.config = Qdrouterd.Config([
             ('router', {'mode': 'standalone', 'id': 'QDR.A'}),
             # Define a connector that uses an unresolvable hostname
             ('connector',
              {'name': 'UnresolvableConn',
-              'host': 'unresolvable.host.name',
+              'host': cls.unresolvable_host_name,
               'port': 'amqp'}),
             ('listener',
              {'port': cls.tester.get_port()}),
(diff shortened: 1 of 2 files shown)


tests/system_tests_bad_configuration.py
```diff
@@ -112,11 +113,11 @@ def wait_for_unresolvable_host(self):
...
             log_lines = router_log.read().split("\n")
             expected_errors = [
-                "proton:io Name or service not known",  # Linux
-                "proton:io unknown node or service",  # macOS
+                "Connection to %s" % self.unresolvable_host_name
             ]
             errors_caught = [line for line in log_lines
-                             if any([expected_error in line for expected_error in expected_errors])]
+                             if any([expected_error in line for expected_error in expected_errors])
+                             and "failed" in line]
```
jiridanek, line 120: If you turn `expected_errors` list above into single `expected_error` value, you can then simplify this into

```suggestion
                             if expected_error in line and "failed" in line]
```

tests/system_tests_bad_configuration.py
```diff
@@ -112,11 +113,11 @@ def wait_for_unresolvable_host(self):
         with open('../setUpClass/test-router.log', 'r') as router_log:
             log_lines = router_log.read().split("\n")
             expected_errors = [
-                "proton:io Name or service not known",  # Linux
-                "proton:io unknown node or service",  # macOS
+                "Connection to %s" % self.unresolvable_host_name
```
jiridanek, line 116: Such error message does not necessarily has to be caused by unresolvable host.

When I was looking into this due to the bubblewrap problem, I was thinking that maybe I should skip the test if I get a failure when contacting a DNS server (because without working DNS I can't properly test the unresolvable host scenario). In the end I was able to get the test pass without modifying it.

I guess this change is OK, I believe the changed test is able to catch pretty much the same bugs it could previously.

jiridanek (commented):
Maybe remove the `--bind /dev/zero /etc/nsswitch.conf` from https://github.com/apache/qpid-dispatch/blob/99eee5a337393878ca6348f0438e7a46b2d7a8bc/tests/CMakeLists.txt#L66-L73 since that workaround should not be necessary any more. GitHub Actions CI is using bubblewrap.

tests/system_tests_bad_configuration.py
```diff
@@ -112,11 +113,11 @@ def wait_for_unresolvable_host(self):
...
             ]
             errors_caught = [line for line in log_lines
-                             if any([expected_error in line for expected_error in expected_errors])]
+                             if any([expected_error in line for expected_error in expected_errors])
+                             and "failed" in line]
 
             self.error_caught = any(errors_caught)
```
jiridanek, line 122: Just a thought. The test does not check that Dispatch tries to connect repeatedly; it is enough to see the log line once.

tests/system_tests_bad_configuration.py
ChugR: cleaned up

tests/system_tests_bad_configuration.py
ChugR: The connection retry is on a five second cadence today. The test polls more often than that and should see the first failed connection. The code accepts one or more logs indicating failure.

jiridanek: Yeah, thats what I meant. If dispatch (due to a bug) stopped retrying the connection, the test would still pass.

jiridanek (approved):
Thanks for the bwrap change. Passing GitHub Action is a sufficient test that it works, because GHA job uses bwrap (it runs ctest -j2 to make tests run faster, and bwrap is needed to avoid port clashes when more than one test runs at the same time).

PR #974, opened 2021-01-08
DISPATCH-1909: Added qdmanage retries to give some time for the core …

ganeshmurthy:
…thread to calculate stats


kgiusti (approved):
These are always fun.

PR #976, opened 2021-01-11
DISPATCH-1850: Added a subscriber count parameter to MobileAddressMul…

ganeshmurthy:
…ticastTest in order to make sure that senders are created only after the interior routers know about all its subscribers

### DIFF
--- a/tests/system_tests_edge_router.py
+++ b/tests/system_tests_edge_router.py
@@ -624,7 +624,9 @@ class RouterTest(TestCase):
                                           self.routers[2].addresses[0],
                                           self.routers[2].addresses[0],
                                           self.routers[2].addresses[0],
-                                          "multicast.24")
+                                          "multicast.24",
+                                          self.routers[2].addresses[0],
+                                          subscriber_count=3)
         test.run()
         self.assertEqual(None, test.error)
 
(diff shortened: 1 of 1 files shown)


tests/system_tests_edge_router.py
```diff
@@ -2521,14 +2568,18 @@ def check_address(self):
         self.num_attempts += 1
         for result in outs.results:
             if self.address in result[0]:
-                found = True
-                self.create_sndr()
-                local_node.close()
-                self.addr_timer.cancel()
-                break
+                if self.subscriber_count == 0 or (result[6] + result[7] == self.subscriber_count):
```
kgiusti, line 2571: Might it be safer to use the actual attribute names rather than hardcoded offsets to access the results?
Example: https://github.com/apache/qpid-dispatch/blob/master/tests/system_tests_one_router.py#L616

kgiusti (approved):
Overall the fix looks good - I have one inline optional suggestion...

tests/system_tests_edge_router.py
ganeshmurthy: What is going on in https://github.com/apache/qpid-dispatch/blob/master/tests/system_tests_one_router.py#L616 is not quite what is going on here but I added a commit which will make you happy. Please take a look.

PR #979, opened 2021-01-13
Tross dispatch 1903 1

ted-ross:
See DISPATCH-1903 for details.

### DIFF
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -22,6 +22,7 @@
 #include <qpid/dispatch/threading.h>
 #include <qpid/dispatch/atomic.h>
 #include <qpid/dispatch/failoverlist.h>
+#include <qpid/dispatch/tempstore.h>
 #include <proton/listener.h>
 #include "dispatch_private.h"
 #include "connection_manager_private.h"
@@ -155,6 +156,44 @@ static qd_config_sasl_plugin_t *qd_find_sasl_plugin(qd_connection_manager_t *cm,
 
     return 0;
 }
+
+
+/**
+ * qdcm_normalize_filename
+ *
+ * If the infile (file name) is absolute or null, simply pass it through.  If it is 
+ * relative (contains no slashes or ".."), check to see if the file is in the temporary
+ * store.  If it is, return the full path to the location of the file in the temporary
+ * store.
+ *
+ * If a new string is allocated to handle the return value, free infile.
+ *
+ * Set the qd_error if there is an error condition.
+ */
+static char* qdcm_normalize_filename(char* infile)
+{
+    if (!infile || *infile == '/' || *infile == '\\')
+        return infile;
+
+    if (!!strchr(infile, '/') || !!strchr(infile, '\\') || !!strstr(infile, "..")) {
+        qd_error(QD_ERROR_CONFIG, "Relative filename '%s' contains '/', '\\', or '..'", infile);
+        return infile;
+    }
+
+    if (!qd_temp_is_store_created()) {
+        qd_error(QD_ERROR_CONFIG, "Relative filename '%s' provided, but the temp store has not been created", infile);
+        return infile;
+    }
+
+    char *new_path = qd_temp_get_path(infile);
+    if (!new_path) {
+        qd_error(QD_ERROR_CONFIG, "Relative filename '%s' not found in the temp store", infile);
+    }
+
+    free(infile);
+    return new_path;
+}
+
 
 void qd_server_config_free(qd_server_config_t *cf)
 {
@@ -411,6 +450,7 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
         qd_config_process_password(&actual_pass, config->sasl_password, &is_file_path, false, qd->connection_manager->log_source);
         if (actual_pass) {
             if (is_file_path) {
+                actual_pass = qdcm_normalize_filename(actual_pass);
                 qd_set_password_from_file(actual_pass, &config->sasl_password, qd->connection_manager->log_source);
                 free(actual_pass);
             }
@@ -420,6 +460,7 @@ static qd_error_t load_server_config(qd_dispatch_t *qd, qd_server_config_t *conf
             }
         }
     }
+    CHECK();
 
     //
     // Handle the defaults for various settings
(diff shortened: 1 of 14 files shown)


tests/system_tests_ssl.py
```diff
@@ -452,6 +458,76 @@ def get_expected_tls_result(self, expected_results):
...
+    def inject_file(self, source_file, dest_file):
+        source_path = self.ssl_file(source_file)
+        url         = Url("amqp://0.0.0.0:%d" % self.PORT_CLEAR)
+        try:
+            connection = BlockingConnection(url, sasl_enabled=False, timeout=self.TIMEOUT)
+            sender     = connection.create_sender('_$qd.store_file')
+            fd         = open(source_path, "r")
+            body       = fd.read()
+            fd.close()
+            sender.send(Message(subject=dest_file, body=body))
```
ganeshmurthy, line 470: Instead of calling this inject_file three times, why not send a message with a body as a dict.

ted-ross: That would definitely be the right way to go for a production tool.  This was an easy way to upload those files for the test.  I didn't feel the need to overly optimize the test code.

tests/system_tests_ssl.py
```diff
@@ -452,6 +458,76 @@ def get_expected_tls_result(self, expected_results):
...
+            return False
+        return True
+
+    def create_ssl_profile(self, name, args):
+        url = Url("amqp://0.0.0.0:%d" % self.PORT_CLEAR)
+        try:
+            connection = BlockingConnection(url, sasl_enabled=False, timeout=self.TIMEOUT)
+            receiver   = connection.create_receiver(None, dynamic=True)
+            sender     = connection.create_sender('$management')
+            sender.send(Message(properties={'operation': 'CREATE',
```
ganeshmurthy, line 487: Instead of doing this, why not use the QdManager class in system_test.py (https://github.com/apache/qpid-dispatch/blob/master/tests/system_test.py#L1147) to create the sslProfile entity.

tests/system_tests_ssl.py
```diff
@@ -452,6 +458,76 @@ def get_expected_tls_result(self, expected_results):
...
+            return False
+        return True
+
+    def create_listener(self, name, args):
+        url = Url("amqp://0.0.0.0:%d" % self.PORT_CLEAR)
+        try:
+            connection = BlockingConnection(url, sasl_enabled=False, timeout=self.TIMEOUT)
+            receiver   = connection.create_receiver(None, dynamic=True)
+            sender     = connection.create_sender('$management')
+            sender.send(Message(properties={'operation': 'CREATE',
```
ganeshmurthy, line 512: This too can use the QdManager class in system_test.py

tests/system_tests_ssl.py
ted-ross: +1 I'll look at that.

src/connection_manager.c
```diff
@@ -156,6 +157,44 @@ static qd_config_sasl_plugin_t *qd_find_sasl_plugin(qd_connection_manager_t *cm,
...
+    if (!infile || *infile == '/' || *infile == '\\')
+        return infile;
+
+    if (!!strchr(infile, '/') || !!strchr(infile, '\\') || !!strstr(infile, "..")) {
+        qd_error(QD_ERROR_CONFIG, "Relative filename '%s' contains '/', '\\', or '..'", infile);
+        return infile;
+    }
+
+    if (!qd_temp_is_store_created()) {
+        qd_error(QD_ERROR_CONFIG, "Relative filename '%s' provided, but the temp store has not been created", infile);
```
ganeshmurthy, line 184: What about the case when a user accidentally puts in a relative filename and knows nothing about the temp store, this log message is potentially confusing?

src/connection_manager.c
```diff
@@ -156,6 +157,44 @@ static qd_config_sasl_plugin_t *qd_find_sasl_plugin(qd_connection_manager_t *cm,
...
+ * If the infile (file name) is absolute or null, simply pass it through.  If it is 
+ * relative (contains no slashes or ".."), check to see if the file is in the temporary
+ * store.  If it is, return the full path to the location of the file in the temporary
+ * store.
+ *
+ * If a new string is allocated to handle the return value, free infile.
+ *
+ * Set the qd_error if there is an error condition.
+ */
+static char* qdcm_normalize_filename(char* infile)
```
ganeshmurthy, line 173: Why is this static function called qdcm_normalize_filename? Why not just qd_normalize_filename. (I am assuming "cm" stands for connection manager? Should we rename other static functions in this file as qdcm_ ?

src/connection_manager.c
ted-ross: Agreed.  There's a pending change where relative filenames will work the same as before.  Temp store files will be denoted using something like "tmp:<filename>".

src/connection_manager.c
ted-ross: We should probably establish a convention for naming static functions.

tests/system_tests_ssl.py
```diff
@@ -516,6 +592,28 @@ def test_tls_all(self):
         self.assertEqual(self.get_expected_tls_result([True, True, True, True]),
                           self.get_allowed_protocols(self.PORT_TLS_ALL))
 
+    @SkipIfNeeded(RouterTestSslBase.DISABLE_SSL_TESTING, RouterTestSslBase.DISABLE_REASON)
+    def test_tls_live(self):
+        """
+        Expects all supported versions: TLSv1, TLSv1.1, TLSv1.2 and TLSv1.3 to be allowed
+        Uses a live-created listener and profile with injected certificate files.
+        """
+        self.inject_file("ca-certificate.pem",     "live-ca.pem")
```
ganeshmurthy, line 601: We probably need one more test where the environment variable TMPDIR is set before starting the router to see if the server code that handles TMPDIR works well and a very long TMPDIR that exceeds MAX_TEMPLATE

tests/system_tests_ssl.py
ted-ross: I didn't realize that QdManager spawns a subprocess to run the qdmanage cli.  That seems heavy-handed for a simple one-off request reply exchange.  I'm going to leave this code as-is.

PR #981, opened 2021-01-15
[trivial] Fix path in doc examples

4383:
qpid-proton subdirectories have been inversed in the doc.

Please take a look to:
https://github.com/apache/qpid-proton/blob/master/python/examples/simple_recv.py

### DIFF
--- a/docs/books/modules/user-guide/sending-test-messages.adoc
+++ b/docs/books/modules/user-guide/sending-test-messages.adoc
@@ -42,7 +42,7 @@ A broker is not used in this procedure, so there is no _"store and forward"_ mec
 --
 [source,bash,options="nowrap",subs="+quotes"]
 ----
-$ cd __<install-dir>__/examples/python/
+$ cd __<install-dir>__/python/examples/
 ----
 
 <install-dir>:: The directory where you installed {ClientAmqpPythonName}.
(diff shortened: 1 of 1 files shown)


ted-ross (commented):
I believe the original text is correct.  Are you looking at the source directory structure or the installed directory structure?  The documentation is referring to the installed artifacts.

4383:
> I believe the original text is correct. Are you looking at the source directory structure or the installed directory structure? The documentation is referring to the installed artifacts.

As this is an example I was looking at the source directory.

I didn't try to build qpid-proton maybe I should have started by that :)

In this case it could be worth to indicate the build step in the documentation.

Thoughts?

jiridanek:
Well, the paths in the doc start with `__<install-dir>__`.

Ultimately, I'd prefer to have the same path for both sources and install. The `examples` part of the path is more important than the fact it is written in python, so in both cases I'd put that first. Grouping things according to technology is sometimes reasonable and necessary; in this case I think it is not necessary.

4383:
I think you're right, the documentation is right, `__<install-dir>__` is explicit.

I'm going to close this PR.
 
Thanks for your time.

jiridanek:
(I forgot to consider, before, that we are talking about qpid-proton examples here; in the qpid-proton repo.)

4383:
> (I forgot to consider, before, that we are talking about qpid-proton examples here; in the qpid-proton repo.)

@jiridanek I'm not sure what you meant by saying that. Do you want to propose something?

jiridanek:
I was thinking about changing the path in the sources to `examples/proton`. Doing things this way would IMO make sense in the Dispatch repo, but not in the Proton repo. (Actually, the Proton repo used to be organized this way; it was later reorganized so that each language binding has its own top level directory.)

4383:
Thanks for details. 

I haven't preference... 

As long as the documentation matches the reality, that's fine with me.

PR #983, opened 2021-01-16
DISPATCH-1750 Update all console deps except Patternfly and d3 

jiridanek:
I found one visual regression in this PR: the console page does not fill the entire browser viewport; this is most visible when the side panel is collapsed and white bars of empty space show on the left and the bottom of the page. In the second screenshot, notice two scrollbars on the right of the page.

Other than the issue above, I did not see anything else new. I will now try to upgrade Patternfly and maybe I will resolve the issue while doing that. If not, I will revisit it here.

dashboard

entities


jiridanek:
I was able to fix the viewport issue by upgrading Patternfly in #984, so I guess we can ignore this PR and move onto that one.

PR #984, opened 2021-01-16
DISPATCH-1750 Update Patternfly packages in console except patternfly-charts

jiridanek:
Followup to #983 

## TODO

* [x] Fix cross-hair cursor over hamburger icon
* [x] Fix missing profile picture in top right corner
* [x] Fix layout breakage for the minute/hour switch in rate graph at Dashboard page
* [x] Fix missing padding between panes at Dashboard page
* [ ] Fix changed behavior of the left-hand side menu; it may be just a desired Patternfly behavior update, not sure yet, it is different, not strictly wrong. Edit: let's call it updated behavior.


jiridanek:
I have the commits split in this way so it can be more easily reviewed. I realize that some of my "fixes" are ugly workarounds, but tbh I don't know the web stuff as well as I would've liked, so this ended up the way I was working. There is not that many hacky workarounds and they are quite short, so I think it is ok...

ErnieAllen (approved):
Thanks Jiri.

PR #985, opened 2021-01-18
NO-JIRA Fix capitalization of "Qpid Dispatch Router"

jiridanek:
When talking about "router network", I was not sure if router should be capitalized. I decided to make it so, but that can be changed...

CC @pwright

### DIFF
--- a/console/react/src/connect/connectPage.js
+++ b/console/react/src/connect/connectPage.js
@@ -82,7 +82,7 @@ class ConnectPage extends React.Component {
           </TextContent>
           <TextContent>
             <Text component="h2" className="connect-description">
-              The console displays information about a qpid dispatch router network. It
+              The console displays information about a Qpid Dispatch Router network. It
               allows monitoring and management control of the router's entities.
             </Text>
             <Text component="h2" className="connect-description">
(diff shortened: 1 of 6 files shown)


pwright (approved):
capitalisation is good, the "Dispatch Router", but a "router" seems to be standard through out https://qpid.apache.org/releases/qpid-dispatch-1.12.0/user-guide/index.html

console/react/src/connect/connectPage.js
```diff
@@ -82,7 +82,7 @@ class ConnectPage extends React.Component {
           </TextContent>
           <TextContent>
             <Text component="h2" className="connect-description">
-              The console displays information about a qpid dispatch router network. It
+              The console displays information about a Qpid Dispatch Router network. It
```
jiridanek, line 85: @pwright This is where I wasn't sure about "router". Should the sentence be

* displays information about a Qpid Dispatch *R*outer network

or

* displays information about a Qpid Dispatch *r*outer network?

I am currently thinking that lowercase r is better choice here.

pwright: to be consistent with the docs you'd use a capital R here.
An alternative would be to reword to:

* displays information about a router network

which, to my eye, looks better. No need to explicitly state the 'long name' everywhere. wdyt?

jiridanek: Thanks, I'll go with capital R then.

PR #991, opened 2021-01-20
DISPATCH-1920: enable TSAN check in travis CI

kgiusti:


ganeshmurthy (approved):

.travis.yml
```diff
@@ -80,6 +81,12 @@ jobs:
     os: linux
     env:
     - PATH="/usr/bin:$PATH" PROTON_VERSION=0.33.0
+  - name: "qdrouterd:TSAN"
+    os: linux
+    dist: bionic
```
jiridanek, line 86: I know it's a bit too late; still, why not use the latest available Ubuntu for this (since newer GCC may have some fixes/features for sanitizers).

PR #992, opened 2021-01-20
DISPATCH-1750 Set nodejs version to latest LTS on Travis

jiridanek:
There is a known issue with a transitive dependency `node-fs-extra` which, starting in version 9.0.0, needs at least nodejs 10 or maybe 10.3.

https://github.com/google/docsy/issues/265 and https://github.com/jprichardson/node-fs-extra/issues/856

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@
 
 language: c
 cache: ccache
+node_js: lts/*
 jobs:
   fast_finish: true
   allow_failures:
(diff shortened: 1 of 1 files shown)


jiridanek:
@ganeshmurthy Travis is extremely slow to run, so I will leave this here until it runs and I see if it helped or not. (It's been already enqueued for 2 hours; I guess it will take until my tomorrow to see results.)

PR #993, opened 2021-01-21
DISPATCH-1920: bump TSAN distro to latest (focal)

kgiusti:


jiridanek (approved):
lgtm; I am thinking about GCC version used; if it's 9, then maybe gcc-10 can be set... Travis is going to take a long time to run this, so I won't know which is the case...

PR #995, opened 2021-01-22
DISPATCH-1926: Added code to selectively allow deletion of http:yes l…

ganeshmurthy:
…isteners based on available libwebsockets version


ted-ross (approved):
I tested this on a LWS v4+ system.  It does the expected thing.

PR #1006, opened 2021-01-29
DISPATCH-1870 Do not test gRPC if HTTP/2 support was not built

jiridanek:
CC @fgiorgetti

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,17 @@ jobs:
   - name: "qdrouterd:TSAN"
     os: linux
     dist: focal
+    before_install:
+      # https://github.com/pypa/virtualenv/issues/1740
+      # https://github.com/pypa/virtualenv/issues/1873
+      - python -m pip install --user --upgrade pip
+      - python -m pip install --user --upgrade tox virtualenv==20.0.23
+      # Install quart to run the http2 tests.
+      - python -m pip install --user quart
+      # DISPATCH-1883: Install selectors to run tcp echo server/client tools
+      - python -m pip install --user selectors
+      # Install grpcio and protobuf to run the grpc tests.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -65,6 +65,8 @@ message(STATUS "Found Proton: ${Proton_LIBRARIES} (found version \"${Proton_VERS
 
 # http2 support is optional
 find_package(libnghttp2 1.4.0)
+CMAKE_DEPENDENT_OPTION(USE_LIBNGHTTP2 "Use libnghttp2 for HTTP/2 support" ON
+                         "libnghttp2_FOUND" OFF)
```
jiridanek, line 69: @ganeshmurthy I think this is worthwhile change. It is the same pattern that is used for libwebsockets defined below this.

The name of the variable could be better, maybe? Something like `QD_USE_LIBNGHTTP2` or maybe `QD_ENABLE_HTTP2`? But then i'd have to rename the other one too, and would have to create alias for backwards compatibility....

ganeshmurthy: Can you please elaborate on why this change is needed ? Thanks.

jiridanek: It is the same pattern that is used for libwebsockets defined below this. It is not needed, but I think that Alan did nice work about CMake definitions for libwebsockets.

The reason for not using `libnghttp2_FOUND` everywhere in the CMakeLists.txt is that "the decision to use the library in compiling dispatch does not only depend on whether the library is present, but the user should also have the option to choose whether the library should be used".

jiridanek: OK, new PR, new Jira.

ganeshmurthy (approved):

PR #1009, opened 2021-01-31
DISPATCH-1937: Free copied http2 header strings

ganeshmurthy:


jiridanek (approved):
Looks like it solved the leak!

PR #1013, opened 2021-02-02
DISPATCH-1946: Ensure that router schema is presented to json as a string

ChugR:
On some platforms the result of get_data is 'bytes' not 'str'.

### DIFF
--- a/python/qpid_dispatch_internal/management/qdrouter.py
+++ b/python/qpid_dispatch_internal/management/qdrouter.py
@@ -40,7 +40,7 @@ class QdSchema(schema.Schema):
 
     def __init__(self):
         """Load schema."""
-        qd_schema = get_data('qpid_dispatch.management', 'qdrouter.json')
+        qd_schema = get_data('qpid_dispatch.management', 'qdrouter.json').decode('utf8')
         try:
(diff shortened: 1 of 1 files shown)


python/qpid_dispatch_internal/management/qdrouter.py
```diff
@@ -40,7 +40,7 @@ class QdSchema(schema.Schema):
 
     def __init__(self):
         """Load schema."""
-        qd_schema = get_data('qpid_dispatch.management', 'qdrouter.json')
+        qd_schema = get_data('qpid_dispatch.management', 'qdrouter.json').decode('utf8')
```
jiridanek, line 43: This is always supposed to produce `bytes`. Weird that it ever worked on Python 3. https://docs.python.org/3/library/pkgutil.html#pkgutil.get_data.

jiridanek (approved):

python/qpid_dispatch_internal/management/qdrouter.py
jiridanek:
```python
def get_data(package: str, resource: str) -> Optional[bytes]: ...
```
https://github.com/python/typeshed/blob/0afc2f9f437e3f20a8109884baf43eac9b848aca/stdlib/pkgutil.pyi#L38

PR #1019, opened 2021-02-05
DISPATCH-1952: fix qd_bitmask_t message streaming leak

kgiusti:
This patch moves the point where the message annotations are updated
to after the check for Q2/streaming.  Not only does this fix the leak,
but it avoids potentially doing the annotation computation multiple
times for the same message.


ganeshmurthy (approved):

kgiusti:
Merged 1.15.0

PR #1024, opened 2021-02-10
DISPATCH-1909: Requests are now recorded just before freeing streams.…

ganeshmurthy:
… Also modified test account for no stats being present


grs (approved):
LGTM!

PR #1025, opened 2021-02-10
DISPATCH-1903 Added logic for storage of temporary files.  Added use …

ted-ross:
…of temp files for SSL profiles.

Added policy and policy enforcement for the temporary file store.
Return appropriate disposition for messages handled.
Added a test for file transfer and the all the error/policy-reject cases.  This tests the disposition return path introduced in DISPATCH-1911.
Use "tmp:" prefix to refer to temporary files in sslProfile.

### DIFF
--- a/src/tempstore-posix.c
+++ b/src/tempstore-posix.c
@@ -1,0 +1,309 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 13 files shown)


src/tempstore-posix.c
```diff
@@ -0,0 +1,309 @@
...
+        return PN_REJECTED;
+    }
+
+    char *filename = (char*) qd_iterator_copy(subject_iter);
+    qd_iterator_free(subject_iter);
+
+    if (!!strchr(filename, '/') || !!strchr(filename, '\\') || !!strstr(filename, "..")) {
+        qd_log(state->log, QD_LOG_ERROR, "Invalid message received - File name with path traversal characters");
+        free(filename);
+        *error = qdr_error(QD_AMQP_COND_INVALID_FIELD, "File name contains path traversal characters");
```
ganeshmurthy, line 215: Can we please mention in the qdrouter.json that file name containing path traversal characters are invalid?

PR #1045, opened 2021-02-18
WIP do not merge: DISPATCH-1968: tcp self test echo server exits when data comes in

ChugR:
This patch is for testing DISPATCH-1968 only.

In this patch the tcp echo server used in system_tests_tcp_adaptor is modified to close the connection to the tcp connector as soon as any data is received.

The tcp tests normally close the tcpListener's connection from the echo client and the closure propagates through dispatch so that the connection through the tcpConnector is initiated from the dispatch side. Normal self tests never close the tcpConnector's connection from the network side.

This self test patch provides an easy test bed for investigating what goes wrong when a server disconnects.

### DIFF
--- a/tests/TCP_echo_server.py
+++ b/tests/TCP_echo_server.py
@@ -88,7 +88,7 @@ def split_chunk_for_display(raw_bytes):
 
 class TcpEchoServer:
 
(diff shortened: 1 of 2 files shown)


ChugR:
New pr coming right up

PR #1047, opened 2021-02-19
DISPATCH-1968: Avoid proton calls on closed raw connections

ChugR:
* Do not write new buffers if connection is CLOSED_WRITE
* Do not call connection_wake if CLOSED_READ or CLOSED_WRITE
This fixes crashes but there is still work left with leaking messages and buffers when server connections close before client connections.

### DIFF
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -56,6 +56,8 @@ struct qdr_tcp_connection_t {
     bool                  egress_dispatcher;
     bool                  connector_closed;//only used if egress_dispatcher=true
     bool                  in_list;         // This connection is in the adaptor's connections list
+    bool                  raw_closed_read;
+    bool                  raw_closed_write;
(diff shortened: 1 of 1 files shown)


src/adaptors/tcp_adaptor.c
```diff
@@ -231,6 +236,31 @@ static int handle_incoming(qdr_tcp_connection_t *conn)
...
+        for (size_t i = conn->outgoing_buff_idx;
+            i < conn->outgoing_buff_idx + conn->outgoing_buff_count;
+            ++i) {
+            if (conn->outgoing_buffs[i].context) {
+                qd_message_stream_data_release(
+                    (qd_message_stream_data_t*)conn->outgoing_buffs[i].context);
+            }
+        }
+    }
+    conn->outgoing_buff_count = 0;
```
ganeshmurthy, line 254: does conn->outgoing_buff_idx need to be reset here as well ?

grs (approved):
:shipit: lgtm

PR #1048, opened 2021-02-20
DISPATCH-1962 Make LSan suppressions more targeted and specific

jiridanek:
This introduces build option -DSANITIZE_3RD_PARTY, defaulted to OFF. The reason is that targeted suppressions only work when the .so libraries are installed with debug symbols, which is not always the case.

Leaks are suppressed more granularly. I have already ideas about fixing some of them, but I wanted to make their suppressions explicit in this way first.

### DIFF
--- a/cmake/RuntimeChecks.cmake
+++ b/cmake/RuntimeChecks.cmake
@@ -106,9 +106,17 @@ elseif(RUNTIME_CHECK STREQUAL "asan")
     message(FATAL_ERROR "libubsan not installed - address sanitizer not available")
   endif(UBSAN_LIBRARY-NOTFOUND)
   message(STATUS "Runtime memory checker: gcc/clang address sanitizers")
+  option(SANITIZE_3RD_PARTY "Detect leaks in 3rd party libraries used by Dispatch while running tests" OFF)
+  file (COPY "${CMAKE_SOURCE_DIR}/tests/lsan.supp" DESTINATION "${CMAKE_BINARY_DIR}/tests")
+  if (NOT SANITIZE_3RD_PARTY)
+    # Append wholesale library suppressions
+    #  this is necessary if target system does not have debug symbols for these libraries installed
+    #  and therefore the more specific suppressions do not match
+    file(APPEND "${CMAKE_BINARY_DIR}/tests/lsan.supp" "\nleak:/libpython2.*.so\nleak:/libpython3.*.so\n")
+  endif ()
   set(SANITIZE_FLAGS "-g -fno-omit-frame-pointer -fsanitize=address,undefined")
   set(RUNTIME_ASAN_ENV_OPTIONS "detect_leaks=true suppressions=${CMAKE_SOURCE_DIR}/tests/asan.supp")
-  set(RUNTIME_LSAN_ENV_OPTIONS "suppressions=${CMAKE_SOURCE_DIR}/tests/lsan.supp")
+  set(RUNTIME_LSAN_ENV_OPTIONS "suppressions=${CMAKE_BINARY_DIR}/tests/lsan.supp")
 
 elseif(RUNTIME_CHECK STREQUAL "tsan")
   assert_has_sanitizers()
--- a/cmake/RuntimeChecks.cmake
+++ b/cmake/RuntimeChecks.cmake
@@ -107,13 +107,23 @@ elseif(RUNTIME_CHECK STREQUAL "asan")
   endif(UBSAN_LIBRARY-NOTFOUND)
   message(STATUS "Runtime memory checker: gcc/clang address sanitizers")
   option(SANITIZE_3RD_PARTY "Detect leaks in 3rd party libraries used by Dispatch while running tests" OFF)
-  file (COPY "${CMAKE_SOURCE_DIR}/tests/lsan.supp" DESTINATION "${CMAKE_BINARY_DIR}/tests")
-  if (NOT SANITIZE_3RD_PARTY)
+  if (SANITIZE_3RD_PARTY)
+    add_custom_command(
+        OUTPUT ${CMAKE_BINARY_DIR}/tests/lsan.supp
+        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_SOURCE_DIR}/tests/lsan.supp ${CMAKE_BINARY_DIR}/tests/lsan.supp
+        DEPENDS ${CMAKE_SOURCE_DIR}/tests/lsan.supp
+        VERBATIM)
+  else (SANITIZE_3RD_PARTY)
     # Append wholesale library suppressions
     #  this is necessary if target system does not have debug symbols for these libraries installed
     #  and therefore the more specific suppressions do not match
-    file(APPEND "${CMAKE_BINARY_DIR}/tests/lsan.supp" "\nleak:/libpython2.*.so\nleak:/libpython3.*.so\n")
+    add_custom_command(
+        OUTPUT ${CMAKE_BINARY_DIR}/tests/lsan.supp
+        COMMAND bash -c 'cat ${CMAKE_SOURCE_DIR}/tests/lsan.supp ${CMAKE_SOURCE_DIR}/tests/lsan_3rdparty.supp > ${CMAKE_BINARY_DIR}/tests/lsan.supp'
+        DEPENDS ${CMAKE_SOURCE_DIR}/tests/lsan.supp ${CMAKE_SOURCE_DIR}/tests/lsan_3rdparty.supp)
   endif ()
+  add_custom_target(generate_lsan.supp ALL
+        DEPENDS ${CMAKE_BINARY_DIR}/tests/lsan.supp)
   set(SANITIZE_FLAGS "-g -fno-omit-frame-pointer -fsanitize=address,undefined")
   set(RUNTIME_ASAN_ENV_OPTIONS "detect_leaks=true suppressions=${CMAKE_SOURCE_DIR}/tests/asan.supp")
   set(RUNTIME_LSAN_ENV_OPTIONS "suppressions=${CMAKE_BINARY_DIR}/tests/lsan.supp")
--- a/tests/lsan.supp
+++ b/tests/lsan.supp
@@ -10,6 +10,27 @@ leak:qd_policy_c_counts_alloc
 leak:qd_policy_open_fetch_settings
 leak:qdr_error_description
 
+# to be triaged; unit_tests
+leak:http-libwebsockets.c
+
+# to be triaged; pretty much all tests
+leak:^IoAdapter_init$
+
+# to be triaged; system_tests_http
+leak:^callback_healthz$
+leak:^callback_metrics$
+
+# to be triaged; system_tests_http1_adaptor
+leak:^pn_condition$
+leak:^pn_raw_connection$
+leak:^pgetaddrinfo$
+
+# to be triaged; system_tests_link_routes
+leak:^pni_init_default_logger$
+
+# Ignore test code
+leak:run_unit_tests.c
+
 # DISPATCH-1844 - shutdown leak
 leak:sys_mutex
 
@@ -23,15 +44,56 @@ leak:_ctypes_alloc_format_string
 leak:__strdup
 
 ####
-#### Miscellaneous 3rd party libraries, test code, etc:
+#### Miscellaneous 3rd party libraries:
 ####
 
-leak:*libpython*
-leak:*libwebsockets*
-leak:*python2*
+### Python
 
-# We should be able to uncomment these once all known dispatch leaks have been fixed
-leak:*libqpid-proton*
+# these Python leaks happen even after simple Py_Initialize(); Py_Finalize();
+#  https://bugs.python.org/issue1635741
+leak:^_PyObject_Realloc
+leak:^PyObject_Malloc$
+leak:^PyThread_allocate_lock$
 
-# Ignore test code
-leak:run_unit_tests.c
+# the PyMalloc mechanism is incompatible with Valgrind, it must be disabled or reported "leaks" must be suppressed
+#  https://pythonextensionpatterns.readthedocs.io/en/latest/debugging/debug_python.html#debug-version-of-python-memory-alloc-label
+leak:^PyMem_Malloc$
+leak:^PyMem_Calloc$
+leak:^_PyObject_GC_Resize$
+# Python uses these alloc functions if you define PYTHONDEVMODE=1
+leak:^_PyMem_DebugRawAlloc$
+leak:^_PyMem_DebugRawRealloc$
+# All the rest
+leak:^list_append$
+leak:^list_resize$
+leak:^_PyBytes_Resize$
+leak:^resize_compact$
+leak:^unicode_resize$
+# Python 2.7
+leak:^PyString_FromStringAndSize$
+leak:^PyString_FromString$
+leak:^PyObject_Realloc$
+leak:^_PyObject_GC_Malloc$
+leak:^_PyString_Resize$
+leak:^PyUnicodeUCS4_FromUnicode$
+leak:^PyList_Append$
+leak:^PyList_New$
+
+### Qpid Proton
+
+# Proton suppressions taken from Proton's lsan.supp
+#  this appears in system_tests_open_properties:
+leak:^pni_data_grow$
+leak:^pn_buffer$
+leak:^pn_buffer_ensure$
+#  this appears in system_tests_http1_adaptor:
+leak:^pn_string_grow$
+leak:^pn_object_new$
+leak:^pn_list$
+leak:^pni_record_create$
+
+### libwebsockets
+
+leak:/libwebsockets.so
+
+### CMake will append .so 3rd party suppressions here, unless disabled:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,6 +53,7 @@ jobs:
         -DCONSOLE_INSTALL=OFF
         -DUSE_BWRAP=ON
         -DRUNTIME_CHECK=${{matrix.runtimeCheck}}
+        -DSANITIZE_3RD_PARTY=ON
 
       CCACHE_BASEDIR: ${{github.workspace}}
       CCACHE_DIR: ${{github.workspace}}/.ccache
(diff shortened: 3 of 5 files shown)


jiridanek:
Needed to suppress more:

```
67: Indirect leak of 32 byte(s) in 1 object(s) allocated from:
67:     #0 0x7f538ada2602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
67:     #1 0x7f5388512802 in pni_mem_allocate /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/memory.c:270
67:     #2 0x7f5388512802 in pn_buffer /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/buffer.c:47
```

```
67: Indirect leak of 32 byte(s) in 1 object(s) allocated from:
67:     #0 0x4964dd in malloc (/home/travis/build/apache/qpid-dispatch/build/router/qdrouterd+0x4964dd)
67:     #1 0x7f997fa1362d in pni_mem_allocate /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/memory.c:270:71
67:     #2 0x7f997fa1362d in pn_buffer /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/buffer.c:47:38
67:     #3 0x7f997fa1362d in pni_data_intern_node.llvm.9093988883496252145 /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/codec.c:475:17
```

```
63: Indirect leak of 16 byte(s) in 1 object(s) allocated from:
63:     #0 0x4976ed in malloc (/home/travis/build/apache/qpid-dispatch/build/tests/test-sender+0x4976ed)
63:     #1 0x7f3b9dbf942a in pni_mem_suballocate /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/memory.c:274:88
63:     #2 0x7f3b9dbf942a in pn_stringn /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/object/string.c:114:28
63:     #3 0x7f3b9dbf942a in pn_string /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/object/string.c:103:10
63:     #4 0x7f3b9dbf942a in pni_init_default_logger /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/core/logger.c:119:32
```

Lets see if its clean result now.

jiridanek:
One more leak in system_tests_http1_adaptor, which appears only sometimes, that needs to be suppressed

```
68: ==15712==ERROR: LeakSanitizer: detected memory leaks
68: 
68: Indirect leak of 64 byte(s) in 1 object(s) allocated from:
68:     #0 0x7fc7780f9602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
68:     #1 0x7fc776378396  (/lib/x86_64-linux-gnu/libc.so.6+0xeb396)
68:     #2 0x7fc77637bd5d in getaddrinfo (/lib/x86_64-linux-gnu/libc.so.6+0xeed5d)
68:     #3 0x7fc7780d268f  (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x7168f)
68:     #4 0x7fc77572568e in pgetaddrinfo /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/proactor/epoll.c:1368
68:     #5 0x7fc77572568e in pn_proactor_raw_connect /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/proactor/epoll_raw_connection.c:196
68:     #6 0x7fc7778ce0ad in _do_reconnect /home/travis/build/apache/qpid-dispatch/src/adaptors/http1/http1_server.c:449
68:     #7 0x7fc777a86632 in qd_timer_visit /home/travis/build/apache/qpid-dispatch/src/timer.c:201
68:     #8 0x7fc777a773b6 in handle /home/travis/build/apache/qpid-dispatch/src/server.c:1008
68:     #9 0x7fc777a7d9b7 in thread_run /home/travis/build/apache/qpid-dispatch/src/server.c:1122
68:     #10 0x7fc777a7f7f6 in qd_server_run /home/travis/build/apache/qpid-dispatch/src/server.c:1484
68:     #11 0x402111 in main_process /home/travis/build/apache/qpid-dispatch/router/src/main.c:113
68:     #12 0x401d68 in main /home/travis/build/apache/qpid-dispatch/router/src/main.c:367
68:     #13 0x7fc7762ad82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
68: 
68: -----------------------------------------------------
68: Suppressions used:
68:   count      bytes template
68:       1         24 ^pn_condition$
68:       1       1536 ^pn_raw_connection$
68:       1         56 qdr_core_subscribe
68:      37        120 _ctypes_alloc_format_string
68:       3        144 ^pn_object_new$
68:       1        128 ^pn_list$
68:       1         24 ^pni_record_create$
68:    5002    3527472 /libpython2.*.so
68: -----------------------------------------------------
68: 
68: SUMMARY: AddressSanitizer: 64 byte(s) leaked in 1 allocation(s).
```

```
68: Indirect leak of 64 byte(s) in 1 object(s) allocated from:
68:     #0 0x7fa6cfbf7602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
68:     #1 0x7fa6cdda0396  (/lib/x86_64-linux-gnu/libc.so.6+0xeb396)
68:     #2 0x7fa6cdda3d5d in getaddrinfo (/lib/x86_64-linux-gnu/libc.so.6+0xeed5d)
68:     #3 0x7fa6cfbd068f  (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x7168f)
68:     #4 0x7fa6cd15198e in pgetaddrinfo /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/proactor/epoll.c:1369
68:     #5 0x7fa6cd15198e in pn_proactor_raw_connect /home/travis/build/apache/qpid-dispatch/qpid-proton/c/src/proactor/epoll_raw_connection.c:193
68:     #6 0x7fa6cf30ccd5 in _do_reconnect /home/travis/build/apache/qpid-dispatch/src/adaptors/http1/http1_server.c:449
68:     #7 0x7fa6cf4f35b3 in qd_timer_visit /home/travis/build/apache/qpid-dispatch/src/timer.c:201
68:     #8 0x7fa6cf4e621e in handle /home/travis/build/apache/qpid-dispatch/src/server.c:1008
68:     #9 0x7fa6cf4e771a in thread_run /home/travis/build/apache/qpid-dispatch/src/server.c:1122
68:     #10 0x7fa6cf3c6d91 in _thread_init /home/travis/build/apache/qpid-dispatch/src/posix/threading.c:172
68:     #11 0x7fa6ced936b9 in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76b9)
68: 
68: -----------------------------------------------------
68: Suppressions used:
68:   count      bytes template
68:       4         96 ^pn_condition$
68:       4       6144 ^pn_raw_connection$
68:       1         56 qdr_core_subscribe
68:      37        120 _ctypes_alloc_format_string
68:       6        336 ^pn_string_grow$
68:      24       1128 ^pn_object_new$
68:       4        512 ^pn_list$
68:       7        168 ^pni_record_create$
68:    4999    3525904 /libpython2.*.so
68: -----------------------------------------------------
68: 
68: SUMMARY: AddressSanitizer: 64 byte(s) leaked in 1 allocation(s).
```

cmake/RuntimeChecks.cmake
```diff
@@ -106,9 +106,17 @@ elseif(RUNTIME_CHECK STREQUAL "asan")
     message(FATAL_ERROR "libubsan not installed - address sanitizer not available")
   endif(UBSAN_LIBRARY-NOTFOUND)
   message(STATUS "Runtime memory checker: gcc/clang address sanitizers")
+  option(SANITIZE_3RD_PARTY "Detect leaks in 3rd party libraries used by Dispatch while running tests" OFF)
+  file (COPY "${CMAKE_SOURCE_DIR}/tests/lsan.supp" DESTINATION "${CMAKE_BINARY_DIR}/tests")
+  if (NOT SANITIZE_3RD_PARTY)
```
ganeshmurthy, line 111: The flag is called SANITIZE_3RD_PARTY but it seems to be specific to libpython2.*.so  and libpython3.*.so
Is Python really third party ?

cmake/RuntimeChecks.cmake
```diff
@@ -106,9 +106,17 @@ elseif(RUNTIME_CHECK STREQUAL "asan")
     message(FATAL_ERROR "libubsan not installed - address sanitizer not available")
   endif(UBSAN_LIBRARY-NOTFOUND)
   message(STATUS "Runtime memory checker: gcc/clang address sanitizers")
+  option(SANITIZE_3RD_PARTY "Detect leaks in 3rd party libraries used by Dispatch while running tests" OFF)
+  file (COPY "${CMAKE_SOURCE_DIR}/tests/lsan.supp" DESTINATION "${CMAKE_BINARY_DIR}/tests")
```
ganeshmurthy, line 110: Instead of copying to the build folder can we keep the lsan.supp in the source folder but add it to the .gitignore ? Then add a file called lsan.supp.in file to the tests folder and add that to git.

tests/lsan.supp
```diff
@@ -23,15 +44,56 @@ leak:_ctypes_alloc_format_string
...
-leak:*libqpid-proton*
+# these Python leaks happen even after simple Py_Initialize(); Py_Finalize();
+#  https://bugs.python.org/issue1635741
+leak:^_PyObject_Realloc
+leak:^PyObject_Malloc$
+leak:^PyThread_allocate_lock$
 
-# Ignore test code
-leak:run_unit_tests.c
+# the PyMalloc mechanism is incompatible with Valgrind, it must be disabled or reported "leaks" must be suppressed
```
ganeshmurthy, line 58: If SANITIZE_3RD_PARTY is OFF, these following Python leak lines are overridden by leak:/libpython2.*.so\nleak:/libpython3.*.so\n ??

tests/lsan.supp
```diff
@@ -10,6 +10,27 @@ leak:qd_policy_c_counts_alloc
 leak:qd_policy_open_fetch_settings
 leak:qdr_error_description
 
+# to be triaged; unit_tests
+leak:http-libwebsockets.c
```
ganeshmurthy, line 14: Just trying to understand why leak:http-libwebsockets.c was added. Can you please elaborate ?

cmake/RuntimeChecks.cmake
jiridanek: There used to be python, libwebsockets and proton in the .so ignores. Proton is I'd say 2nd party and it is not supposed to leak, so the so should not be suppressed wholesale. The libwebsockets ignore will be unnecessary, so I pulled it out into the lsan.supp directly and I have a plan to fix the leak there. Now only Python remains.

I think I prefer having an option with a generic name. imagine if there were any leaks in cyrus sasl or openssl, those would go under this option. It is possible that by fixing other suppressions, something like this will crop up.

cmake/RuntimeChecks.cmake
jiridanek: I saw that pattern used elsewhere in dispatch. I thought that having generated files outside of `CMAKE_BINARY_DIR` is discouraged.

One problem with that is when you have multiple `build` folders, e.g. debug and release build. Then running cmake for both at approx the same time (my IDE does that) can cause weird issues.

I guess I can do as you ask. Also, I should probably try to use `configure_file` in cmake.

tests/lsan.supp
jiridanek: Yes, exactly. These following lines require Python debug symbols to be installed, so the function names can be matched.

tests/lsan.supp
jiridanek: `leak:http-libwebsockets.c` is a narrower version of `leak:*libwebsockets*` which was there previously. (The suppression patterns are matched a substrings, as I mentioned on the Jira) I made the suppression more targeted here.

ganeshmurthy: Ok, what you are saying makes sense to me. What I would like to happen is for any leaks in our code in the http-libwebsockets.c to be not ignored. Like so, there is a bunch of code we wrote relating to stats and metrics and if there are any leaks in our code, I would like those to be not suppressed. I do understand that we previously had leak:*libwebsockets*

cmake/RuntimeChecks.cmake
ganeshmurthy: Good point about not having generated files outside of CMAKE_BINARY_DIR. I remember following that rule when working in Java projects using Maven. We don't seem to follow that rule here which I have not questioned so far. 
And, yes, I was talking about using the configure_file in cmake.
But you should wait on this to see what @kgiusti 's take on this is. Thx

tests/lsan.supp
ganeshmurthy: So, if I understand correctly, no matter what you have in this list below, having the leak:/libpython2..so\nleak:/libpython3..so\n at the end of the file totally ignores this list. So, does it make sense to not even have this list if we are going to have the above two lines at the end of the file ?

jiridanek: We do not append the `leak:/libpython3..so` always, only when the SANITIZE_3RD_PARTY option is OFF. In GitHub Actions, the Python debug info is installed and works ok, so there SANITIZE_3RD_PARTY=ON and the more granular suppressions are used. (And therefore e.g. the leak in IoAdapter is picked up). Same thing on my laptop.

On Travis, and on other people's computers, the debug symbols might not be present. In that case, the whole-library suppressions are needed, otherwise they would see test failures.

tests/lsan.supp
jiridanek: I'd be happy to fix this in a subsequent step, or at least make an attempt at it, but for now, currently there _are_ leaks coming from `http-libwebsockets.c` so the suppression needs to stay here:

https://github.com/jiridanek/qpid-dispatch/runs/1964063810?check_suite_focus=true#step:9:593

```
9: =================================================================
9: ==2292==ERROR: LeakSanitizer: detected memory leaks
9: 
9: Indirect leak of 216 byte(s) in 1 object(s) allocated from:
9:     #0 0x7fa3b6a8bdc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
9:     #1 0x7fa3b63ce928 in qd_http_server ../src/http-libwebsockets.c:963
9:     #2 0x7fa3b63b5ec7 in qd_server ../src/server.c:1364
9:     #3 0x7fa3b6207fff in qd_dispatch_prepare ../src/dispatch.c:330
9:     #4 0x7fa3b1208ff4  (/lib/x86_64-linux-gnu/libffi.so.7+0x6ff4)
9:     #5 0x7ffce6d1b37f  ([stack]+0x1f37f)
9: 
9: Indirect leak of 64 byte(s) in 1 object(s) allocated from:
9:     #0 0x7fa3b6a8caa5 in posix_memalign (/lib/x86_64-linux-gnu/libasan.so.5+0x10eaa5)
9:     #1 0x7fa3b6270a94 in sys_cond ../src/posix/threading.c:77
9:     #2 0x7fa3b63ce96d in work_queue_init ../src/http-libwebsockets.c:231
9:     #3 0x7fa3b63ce96d in qd_http_server ../src/http-libwebsockets.c:965
9:     #4 0x7fa3b63b5ec7 in qd_server ../src/server.c:1364
9:     #5 0x7fa3b6207fff in qd_dispatch_prepare ../src/dispatch.c:330
9:     #6 0x7fa3b1208ff4  (/lib/x86_64-linux-gnu/libffi.so.7+0x6ff4)
9:     #7 0x7ffce6d1b37f  ([stack]+0x1f37f)
9: 
9: -----------------------------------------------------
9: Suppressions used:
9:   count      bytes template
9:       1       1160 run_unit_tests.c
9:       1         64 sys_mutex
9:       1         56 qdr_core_subscribe
9:       9       5080 ^_PyObject_Realloc
9:     540     862547 ^PyObject_Malloc$
9:       1         32 ^PyThread_allocate_lock$
9:       4       9897 ^PyMem_Malloc$
9:       1        856 ^_PyObject_GC_Resize$
9:       1        295 ^_PyBytes_Resize$
9:       5    1058064 /libwebsockets.so
9: -----------------------------------------------------
9: 
9: SUMMARY: AddressSanitizer: 280 byte(s) leaked in 2 allocation(s).
 7/37 Test  #9: unit_tests .....................................***Failed    2.03 sec
```

jiridanek: (IIRC that leak ^^^ is essentially a test issue. There are subsequent suppressions `leak:^callback_healthz$` and `leak:^callback_metrics$` which looked to me more real, but it's hard to tell. And there is little point in investigating/fixing things when I'm no longer dealing with upstream master branch, but with my own patched version, so any PR I might produce would end up rebased few times down the road...)

ganeshmurthy: ok, let's fix this in a subsequent step. Thx

tests/lsan.supp
ganeshmurthy: It would be great if the following list *and* leak:/libpython2..so\nleak:/libpython3..so\n don't appear at the same time in the file.

jiridanek: Why?

The way it is now (in the PR), the `lsan.supp` (or `lsan.supp.in` if I make that change) is a valid and usable LSan suppression file by itself, without any preprocessing. If I to do what you suggest, this will no longer be the case: (it will either contain CMake variable and won't be valid because of it, or it will not contain the individual Python function suppressions and will be incomplete).

ganeshmurthy: In case you end up doing the configure_file from the first comment, you could add this as well? It is deceiving to see a  suppression list being overridden by two lines at the bottom.

jiridanek: It isn't really overridden. The two .so suppressions simply cover all the Python suppressions that went before. I am thinking I could remove the conditional CMake logic and put the .so suppressions for python there always.

The only leak I flushed out by doing this in the more complicated way is th `IoAdapter` one, and that is fairly insignificant. There is about 4 or 5 instances of the class, only. In the whole qdr run. So that leak is hardly the most significant one.

cmake/RuntimeChecks.cmake
jiridanek: `configure_file` does not work all that well for my purposes; I want to essentially call a `cat` to concatenate suppression lists together, so I used a custom command

tests/lsan.supp
jiridanek: https://github.com/apache/qpid-dispatch/pull/1080

PR #1053, opened 2021-02-21
DISPATCH-1814 use autopep8 to reformat Python code for PEP 8

jiridanek:


jiridanek:
I read through the diff twice and it looks OK to me. Besides, it was created by automated tool, so there shouldn't be careless mistakes in it. I guess I won't be able to persuade anyone else to read it too :( It is probably best to get this over with.

PR #1059, opened 2021-02-27
DISPATCH-1974 Update deprecated assertEquals, remove suppression H234

jiridanek:


jiridanek:
This is a trivial fix, unlikely to cause any flake8 problems with PRs down the road, so let's just merge it.

PR #1064, opened 2021-03-05
DISPATCH-1989 Test w/ libuv in GitHub Actions CI

jiridanek:


jiridanek:
There does not seem to be much interest in libuv beyond that one issue. And it is probably better to keep GHA jobs lean and mean; I prefer to put in a Windows job eventually in the future, rather than libuv one.

PR #1074, opened 2021-03-15
DISPATCH-2001: set streaming link priority to 4

kgiusti:
Use the default priority when creating streaming links.

### DIFF
--- a/src/router_core/connections.c
+++ b/src/router_core/connections.c
@@ -1443,6 +1443,7 @@ qdr_link_t *qdr_connection_new_streaming_link_CT(qdr_core_t *core, qdr_connectio
 
     if (out_link) {
         out_link->streaming = true;
+        out_link->priority = 4;
         if (!conn->has_streaming_links) {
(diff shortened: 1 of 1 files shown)


ChugR (approved):

src/router_core/connections.c
```diff
@@ -1443,6 +1443,7 @@ qdr_link_t *qdr_connection_new_streaming_link_CT(qdr_core_t *core, qdr_connectio
 
     if (out_link) {
         out_link->streaming = true;
+        out_link->priority = 4;
```
ganeshmurthy, line 1446: Should we instead default the priority field in the qdr_link to 4  ?

kgiusti: We could - this was the least-risk minimal fix in case we needed to backport to 1.15.0.

Going forward we probably need to consider making all router control traffic a higher priority than any user-generated traffic.

ganeshmurthy (approved):

PR #1077, opened 2021-03-19
DISPATCH-2011 Add aarch64 .travis.yml job; add HWASAN option

jiridanek:


jiridanek:
I noticed that Aarch64 tended to get scheduled on Travis a bit sooner than the amd64 builds! Also, I noticed that Travis has the s390x arch available, https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z. That could be a good thing to take advantage of.

PR #1090, opened 2021-03-27
DISPATCH-2025 Compile using C11 language version

jiridanek:
I thought about this more and I see no downside in doing this change now. C11 is well supported, with the exceptions of those atomics in GCC 4.8.5, which I already took care of, and C99 -> C11 is not a huge change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ if (DEFINED CMAKE_C_COMPILE_FEATURES)
 else()
     set(C_STANDARD_GNU "-std=gnu11")
     set(C_STANDARD_Clang "-std=gnu11")
-    set(C_STANDARD_SunPro "-xc11")
+    set(C_STANDARD_SunPro "-std=c11")
 
     set(C_STANDARD_FLAGS ${C_STANDARD_${CMAKE_C_COMPILER_ID}})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -20,11 +20,46 @@
...
+    set(CMAKE_C_STANDARD_REQUIRED ON)
+    set(CMAKE_C_EXTENSIONS ON) # gnu11
+
+    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+else()
+    set(C_STANDARD_GNU "-std=gnu11")
+    set(C_STANDARD_Clang "-std=gnu11")
+    set(C_STANDARD_SunPro "-xc11")
```
jiridanek, line 35: or `-std=c11`? https://docs.oracle.com/cd/E37069_01/html/E37074/appxc.html

jiridanek: My best guess is `-std=c11` for SunPro. I'll put it there and we will see when somebody who actually has SunOS tries to compile.

jiridanek, line 35:
```suggestion
    set(C_STANDARD_SunPro "-std=c11")
```

PR #1097, opened 2021-04-01
DISPATCH-2029: Added a base class and moved some of the http1 adaptor…

ganeshmurthy:
… tests to this base class. Also introduced a new test that executes http1 tests over tcp

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -162,6 +162,7 @@ foreach(py_test_module
     system_tests_open_properties
     ${SYSTEM_TESTS_HTTP2}
     system_tests_http1_adaptor
+    system_tests_http1_over_tcp
     system_tests_tcp_adaptor
     )
 
--- a/tests/system_tests_http1_base.py
+++ b/tests/system_tests_http1_base.py
@@ -1,0 +1,1187 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+import socket
+import uuid
+from threading import Thread
+
+from time import sleep
+try:
+    from http.server import HTTPServer, BaseHTTPRequestHandler
+    from http.client import HTTPConnection
+    from http.client import HTTPException
+except ImportError:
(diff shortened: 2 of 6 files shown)


tests/system_tests_http1_base.py
```diff
@@ -0,0 +1,1187 @@
...
+                                % (key, value, rsp.getheader(key)))
+
+        body = rsp.read()
+        if (self.expect_body and self.expect_body != body):
+            raise Exception("Bad response body expected %s got %s"
+                            % (self.expect_body, body))
+        return body
+
+
+class CommonHttp1AdaptorEdge2EdgeTest():
```
kgiusti, line 370: add object as parent class

tests/system_tests_http1_base.py
```diff
@@ -0,0 +1,1187 @@
...
+                                      repeat=2)
+        client10.wait()
+        self.assertIsNone(client10.error)
+        self.assertEqual(4, client10.count)
+
+        server11.wait()
+        server10.wait()
+
+
+class CommonHttp1AdaptorOneRouterTest():
```
kgiusti, line 732: also here

tests/system_tests_http1_base.py
```diff
@@ -0,0 +1,1187 @@
+#
```
kgiusti, line 1: Hmmm... I think this file shouldn't be a "system_tests_*" file since it does not define any TestCases.  It defines a set of HTTP/1.x tests that are shared between the http1_adaptor TestCases and http1_over_tcp TestCases.

Name this fine something like http1_tests.py instead otherwise I think people will be tempted to add HTTP/1.x TestCases here instead of in the proper system_tests_*.py files.

tests/system_tests_http1_base.py
```diff
@@ -0,0 +1,1187 @@
...
+        client.close()
+
+    def test_008_put_10(self):
+        client = HTTPConnection("127.0.0.1:%s" % self.http_listener10_port,
+                                timeout=TIMEOUT)
+        self._do_request(client, self.TESTS_10["PUT"])
+        client.close()
+
+
+class Http1OneRouterTestBase(TestCase):
```
kgiusti, line 1125: Why is this broken up into two classes: Http1OneRouterTestBase and CommonHttp1AdaptorOneRouterTest?   Since they must be inherited from together why not just combine them into a single class?

ganeshmurthy: The CommonHttp1AdaptorOneRouterTest contains only the tests, it does not have any setup information. In the future, this separation enables the the battery of tests in CommonHttp1AdaptorOneRouterTest to be run against a totally different setup

tests/CMakeLists.txt
```diff
@@ -162,6 +162,7 @@ foreach(py_test_module
     system_tests_open_properties
     ${SYSTEM_TESTS_HTTP2}
     system_tests_http1_adaptor
+    system_tests_http1_over_tcp
```
jiridanek, line 165: This test needs to be disabled on macOS, because it depends on raw-connections functionality in Proton. That's not available in the libuv proactor. I filled a Jira for it before I realized the test is simply incompatible with macOS https://issues.apache.org/jira/browse/DISPATCH-2042.

Please add test name to

https://github.com/apache/qpid-dispatch/blob/db30cb043bece740b5273b137057aa67b161ed50/.travis.yml#L139-L140

kgiusti (commented):
I'm good with this once Jiri's request has been added.

PR #1101, opened 2021-04-03
DISPATCH-2030 Add additional sanitizer checks and surface router output better in tests

jiridanek:
1. UBSAN is not killing the process the moment undefined behavior is detected. This is easily fixed by doing `set(RUNTIME_UBSAN_ENV_OPTIONS "halt_on_error=1 abort_on_error=1")`, but then the tests timeout all the time. Not useful.
2. UBSAN is not printing diagnostics. That is fixed by doing `set(RUNTIME_UBSAN_ENV_OPTIONS "print_stacktrace=1 print_summary=1")`
3. ASAN has extra checks that can be enabled. Enabling them does not find any new issues, so far. But it sounds like a sensible thing to turn on, for the future. `set(RUNTIME_ASAN_ENV_OPTIONS "strict_string_checks=1 detect_stack_use_after_return=1 check_initialization_order=1 strict_init_order=1")`
4. System tests don't print `.out` files produced by running routers, even if there are messages from sanitizers in them. Let's fix that. This means that tests become _quite_ chatty. Maybe not something that should be turned on forever, but now it's good to see the UBSAN messages.

### DIFF
--- a/cmake/RuntimeChecks.cmake
+++ b/cmake/RuntimeChecks.cmake
@@ -141,7 +141,7 @@ elseif(RUNTIME_CHECK STREQUAL "asan" OR RUNTIME_CHECK STREQUAL "hwasan")
     set(DETECT_LEAKS true)
   endif()
   # https://github.com/openSUSE/systemd/blob/1270e56526cd5a3f485ae2aba975345c38860d37/docs/TESTING_WITH_SANITIZERS.md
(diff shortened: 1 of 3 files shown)


jiridanek:
This change does not appear to cause any extra test failures as I was afraid (esp. regarding the strict_string_checking asan flag), so after offline review, I will push this based on silent consensus principle

PR #1103, opened 2021-04-05
DISPATCH-2023: Set self.reactor in on_start

ganeshmurthy:


jiridanek (approved):
lgtm, this should take care of the exception, hopefully!

PR #1107, opened 2021-04-06
DISPATCH-1969 Fix pdf rendering issue in  docs

pwright:
pdf output producing warning due to use of &ndash;


jiridanek:
This is weird, I kind of like all those m-dashes and n-dashes in a text, I am sorry to see this one go. :P

PR #1108, opened 2021-04-07
NO-JIRA-YET: Set the buffer size to 4k. Also adjusts Q2 and Q3 accord…

ganeshmurthy:
…ingly.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -24,7 +24,6 @@
 #include <stdint.h>
 #include <string.h>
 
-// BUFFER_SIZE is set to 4k
 size_t BUFFER_SIZE     = 4096;
 
 ALLOC_DECLARE(qd_buffer_t);
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -24,6 +24,7 @@
 #include <stdint.h>
 #include <string.h>
 
+// IMPORTANT: When changing BUFFER_SIZE, please change the corresponding constant declared in tests/system_test.py
 size_t BUFFER_SIZE     = 4096;
 
 ALLOC_DECLARE(qd_buffer_t);
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -24,8 +24,8 @@
 #include <stdint.h>
 #include <string.h>
 
-
-size_t BUFFER_SIZE     = 512;
+// BUFFER_SIZE is set to 4k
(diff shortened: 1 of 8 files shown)


src/buffer.c
```diff
@@ -24,8 +24,8 @@
 #include <stdint.h>
 #include <string.h>
 
-
-size_t BUFFER_SIZE     = 512;
+// BUFFER_SIZE is set to 4k
```
ChugR, line 27: "size_t BUFFER_SIZE = 4096;" 

stands pretty well on its own and does not need a comment.

ChugR (approved):
This patch looks OK. It would be interesting to verify that 512 vs 4096 router versions interoperate well.

src/buffer.c
ganeshmurthy: Removed comment, thanks

tests/run_unit_tests_size.c
```diff
@@ -45,7 +45,7 @@ int main(int argc, char** argv)
     qd_buffer_set_size(buffer_size);
 
     int result = 0;
-    result += message_tests();
+    result += message_tests(buffer_size);
```
kgiusti, line 48: should the default buffer size be updated on line 34 above?

kgiusti:
There's some test failures that might be relying on the 512 buffer size.

https://github.com/apache/qpid-dispatch/blob/main/tests/system_tests_one_router.py#L3577

Also we need to add a 4096 test case here:
https://github.com/apache/qpid-dispatch/blob/main/tests/CMakeLists.txt#L76

kgiusti (commented):
I would also define python constants in tests/system_test.py for BUFFER_SIZE and all the Q2 and Q3 values in message.h.   These can be used by the tests for testing buffer/Q2/Q3 limits.

I'd also add a big comment around the C definitions that says if these values change the corresponding values in tests/system_test.py must be updated.

kgiusti:
Oooh boy - HTTP1 tests don't see to be too happy with these changes...

ganeshmurthy:
> Oooh boy - HTTP1 tests don't see to be too happy with these changes...

I saw several "OSError: [Errno 98] Address already in use" errors and assumed those had to do with https://issues.apache.org/jira/browse/DISPATCH-1986

PR #1112, opened 2021-04-12
NotLinkingJiraYet fail test if subprocess aborts or segfaults

jiridanek:


jiridanek:
just a note to myself, two segfaults reported by ctest without any details or backtraces

https://travis-ci.com/github/apache/qpid-dispatch/jobs/497981423#L4673
https://travis-ci.com/github/apache/qpid-dispatch/jobs/497981421#L10288

jiridanek:
The problem with this is that the router not always segfaults or aborts, sometimes it just exits. Some tests do expect a !=0 ecode exit. The code has to look for expected ecode in the event handler, compare those two, and somehow print output (so it does not get lost) and either kill the test, or continue from the next testcase with restored routers.

jiridanek:
I was thinking about few more things. First, reporting/dealing with leaked threads, https://issues.apache.org/jira/browse/DISPATCH-2323?focusedCommentId=17484381&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17484381

and with adding rr --chaos, https://groups.google.com/g/mozilla.dev.platform/c/wJJUkf0GESE, which requires dealing with rr's inability (or unwillingness) to forward signals it receives to the inferior process under debugging. Solving this does not sound hard and should be interesting. (Btw, I first saw https://issues.apache.org/jira/browse/DISPATCH-2318 under rr chaos)

PR #1115, opened 2021-04-13
DISPATCH-2040 - For link routes, even if there is no change in dispos…

ganeshmurthy:
…ition state, still copy the disposition data

### DIFF
--- a/src/router_node.c
+++ b/src/router_node.c
@@ -2050,9 +2050,12 @@ static void CORE_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t di
         return;
 
     //
-    // If the disposition has changed and the proton delivery has not already been settled, update the proton delivery.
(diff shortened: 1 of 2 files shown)


kgiusti (changes requested):
There's a bunch of test failures due to an unrelated leak bug:
https://travis-ci.com/github/apache/qpid-dispatch/jobs/498886172#L3365

However travis show failures in the link route tests as well:
https://travis-ci.com/github/apache/qpid-dispatch/jobs/498886172#L2728

This may be due to the following ASAN failure:
https://travis-ci.com/github/apache/qpid-dispatch/jobs/498886172#L2759

While the patch itself looks safe I'm like to see it re-run with the previous errors fixed if possible.

kgiusti (approved):

PR #1116, opened 2021-04-13
DISPATCH-2030: Use memcpy to copy the tag into the delivery

ganeshmurthy:


jiridanek:
lgtm, the warning has disappeared and there is probably no point spending more time on this. I was originally thinking about aligning the field in the struct to access it as uint64, but that is probably not worth it, since this PR solves it.

kgiusti (approved):

PR #1125, opened 2021-04-15
POC: Tcp backpressure - DO NOT MERGE!

kgiusti:


src/router_core/delivery.c
```diff
@@ -712,14 +712,15 @@ static void qdr_delivery_anycast_update_CT(qdr_core_t *core, qdr_delivery_t *dlv
...
+    // If the disposition or extended delivery-state has changed and there is a
+    // peer link, set the disposition/delivery-state of the peer
     // If remote settled, the delivery must be unlinked and freed.
     // If remote settled and there is a peer, the peer shall be settled and
     // unlinked.  It shall not be freed until the connection-side thread
     // settles the PN delivery.
     //
-    if (new_disp != dlv->remote_disposition) {
+    // kag fix me!
+    if (new_disp != dlv->remote_disposition || dlv->remote_state) {
```
grs, line 723: An alternative here would be to test the disposition. E.g. TCP_SEQ_ACK (or PN_RECEIVED) would imply an update required.

src/router_core/delivery.c
```diff
@@ -1246,6 +1247,8 @@ qd_delivery_state_t *qdr_delivery_take_local_delivery_state(qdr_delivery_t *dlv,
 //
 void qdr_delivery_move_delivery_state_CT(qdr_delivery_t *dlv, qdr_delivery_t *peer)
 {
+    /// kag: crap, this may be an issue: possible drop of seq updates!
+
```
grs, line 1251: For TCP_SEQ_ACK (or indeed PN_RECEIVED) it is just a sequence number (or two) to be updated. Perhaps we could update that in place?

src/adaptors/tcp_adaptor.c
```diff
@@ -34,6 +34,11 @@
 #include <inttypes.h>
 #include <stdio.h>
 
+const uint32_t TCP_MAX_CAPACITY = 131072;
+const uint64_t TCP_SEQ_ACK = 999;  // TBD
```
grs, line 38: PN_RECEIVED is very similar to this new type, but has two sequences (we could ignore one or the other in the case of TCP should we wish).

One minor upside of using that is that the protocol trace logging in proton already includes the state.

src/router_core/delivery.c
kgiusti: Good point - I'm going to need to refactor this a bit more.  This POC is exposing a few assumptions we've made regarding disposition propagation that need to be revisited.

kgiusti:
Discarded.  Replaced by fix for DISPATCH-2073

PR #1127, opened 2021-04-15
DISPATCH-2048: Added code to clean out the qdr_delivery_cleanup_t on …

ganeshmurthy:
…router shutdown

### DIFF
--- a/src/router_core/router_core.c
+++ b/src/router_core/router_core.c
@@ -43,6 +43,8 @@ ALLOC_DEFINE(qdr_subscription_ref_t);
 const uint64_t QD_DELIVERY_MOVED_TO_NEW_LINK = 999999999;
 
 static void qdr_general_handler(void *context);
+static void qdr_general_handler_cleanup(void *context);
+
 
 static void qdr_core_setup_init(qdr_core_t *core)
 {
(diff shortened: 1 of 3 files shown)


src/router_core/router_core.c
```diff
@@ -258,6 +260,22 @@ void qdr_core_free(qdr_core_t *core)
         link = DEQ_HEAD(core->open_links);
     }
 
+    //
+    // Clean up any qdr_delivery_cleanup_t's that are still left in the core->delivery_cleanup_list
+    //
+    if (DEQ_SIZE(core->delivery_cleanup_list) > 0) {
```
kgiusti, line 266: I'm confused - this check shouldn't be necessary since if length is zero the loop below will not execute since DEQ_HEAD/cleanup will == 0?

kgiusti: Also why not simply call qdr_do_message_to_addr_free() here instead of duplicating the code?

src/router_core/router_core.c
```diff
@@ -896,6 +914,27 @@ void qdr_del_subscription_ref_CT(qdr_subscription_ref_list_t *list, qdr_subscrip
...
+
+    sys_mutex_lock(core->work_lock);
+    DEQ_MOVE(core->work_list, work_list);
+    sys_mutex_unlock(core->work_lock);
+
+    work = DEQ_HEAD(work_list);
+    while (work) {
+        DEQ_REMOVE_HEAD(work_list);
+        work->handler(core, work);
+        qdr_do_message_to_addr_free(core, work);
```
kgiusti, line 929: Isn't it possible that this will double free if work->handler is qdr_do_message_to_addr_free()?

PR #1129, opened 2021-04-15
DISPATCH-1878: Handle half-closed TCP connections - DO NOT MERGE

ChugR:
DISPATCH-2043: Various TCP Adaptor logging improvements

* monitor and control sequence of half closed events
* raw read caches data if read_close before stream message created
* raw write defers EOS write_close until stream message created

This version leaks buffers at the rate of about 15 per connection pair.
The suspect is that proton never emits a PN_RAW_CONNECTION_READ after
emitting CLOSED_READ and this code never gets to reclaim the buffers.
That said, this code could have a bug in that area as well.

### DIFF
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -388,15 +388,6 @@ static int handle_incoming(qdr_tcp_connection_t *conn, const char *msg)
             conn->conn_id, conn->incoming_id);
         qd_message_set_receive_complete(qdr_delivery_message(conn->instream));
         qdr_delivery_continue(tcp_adaptor->core, conn->instream, true);
-        qdr_delivery_decref(tcp_adaptor->core, conn->instream, "tcp-adaptor.pn_raw_connection_closed_read - instream");
-        conn->instream = 0;
-        if (conn->incoming) {
-            qd_log(log, QD_LOG_DEBUG,
-                    "[C%"PRIu64"][L%"PRIu64"] pn_raw_connection_closed_read - detach incoming link",
-                    conn->conn_id, conn->incoming_id);
-            qdr_link_detach(conn->incoming, QD_LOST, 0);
-            conn->incoming = 0;
-        }
         conn->raw_read_shutdown = true;
     }
 
@@ -1447,20 +1438,15 @@ static void qdr_tcp_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t
                DLV_ARGS(dlv), disp, settled ? "true" : "false");
 
         if (settled) {
-            if (dlv == tc->instream) {
-                qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG,
-                       DLV_FMT" qdr_tcp_delivery_update: instream is settled: call pn_raw_connection_read_close()",
-                       DLV_ARGS(dlv));
-                pn_raw_connection_read_close(tc->pn_raw_conn);
-            } else if (dlv == tc->outstream) {
-                qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG,
-                       DLV_FMT" qdr_tcp_delivery_update: outstream is settled: call pn_raw_connection_write_close()",
-                       DLV_ARGS(dlv));
-                pn_raw_connection_write_close(tc->pn_raw_conn);
-            } else {
-                qd_log(tcp_adaptor->log_source, QD_LOG_ERROR, "Unexpected delivery settled");
-                assert(false);
-            }
+            // the only settlement occurs when the initial delivery is
+            // settled, which occurs when the connector is unable to
+            // connect to the configured tcp endpoint, so in this case
+            // we can just close the connection
+            // (The end of the message is used to convey half closed status)
+            qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG,
+                   DLV_FMT" qdr_tcp_delivery_update: call pn_raw_connection_close()",
+                   DLV_ARGS(dlv));
+            pn_raw_connection_close(tc->pn_raw_conn);
         }
     } else {
         qd_log(tcp_adaptor->log_source, QD_LOG_ERROR, "qdr_tcp_delivery_update: no link context");
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -810,13 +810,6 @@ static void handle_connection_event(pn_event_t *e, qd_server_t *qd_server, void 
         sys_mutex_lock(conn->activation_lock);
         conn->raw_closed_write = true;
         sys_mutex_unlock(conn->activation_lock);
-        if (conn->ingress) {
-            // connection from client is no longer writable
-            qd_log(log, QD_LOG_DEBUG,
-                   "[C%"PRIu64"] PN_RAW_CONNECTION_CLOSED_WRITE call pn_raw_connection_close(). client no longer writable",
-                   conn->conn_id);
-            pn_raw_connection_close(conn->pn_raw_conn);
-        }
         break;
     }
     case PN_RAW_CONNECTION_DISCONNECTED: {
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -329,16 +329,6 @@ static int handle_incoming(qdr_tcp_connection_t *conn, const char *msg)
         //qd_compose_insert_null(props);                      // reply-to-group-id
         qd_compose_end_list(props);
 
-        unsigned int length = qd_buffer_list_length(&conn->early_raw_read_bufs);
-        if (length > 0) {
-            // TODO: This won't overun peer's Q2 will it?
-            // At some point the early raw read buffs should be chunked
-            // into several BODY_DATA segments.
-            props = qd_compose(QD_PERFORMATIVE_BODY_DATA, props);
-            qd_compose_insert_binary_buffers(props, &conn->early_raw_read_bufs);
-            DEQ_INIT(conn->early_raw_read_bufs);
-        }
-
         qd_message_compose_2(msg, props, false);
         qd_compose_free(props);
 
@@ -347,9 +337,26 @@ static int handle_incoming(qdr_tcp_connection_t *conn, const char *msg)
         qd_message_set_q2_unblocked_handler(msg, qdr_tcp_q2_unblocked_handler, conn_sp);
 
         conn->instream = qdr_link_deliver(conn->incoming, msg, 0, false, 0, 0, 0, 0);
+
         qd_log(log, QD_LOG_DEBUG,
-               "[C%"PRIu64"][L%"PRIu64"][D%"PRIu64"] Initiating ingress stream message with %u bytes",
-               conn->conn_id, conn->incoming_id, conn->instream->delivery_id, length);
+               "[C%"PRIu64"][L%"PRIu64"][D%"PRIu64"] Initiating ingress stream message with 0 bytes",
+               conn->conn_id, conn->incoming_id, conn->instream->delivery_id);
+
+        unsigned int length = qd_buffer_list_length(&conn->early_raw_read_bufs);
+        if (length > 0) {
+            qd_log(log, QD_LOG_TRACE,
+                    "[C%"PRIu64"][L%"PRIu64"] client link sending %d cached bytes",
+                    conn->conn_id, conn->incoming_id, length);
+            qd_message_stream_data_append(qdr_delivery_message(conn->instream), &conn->early_raw_read_bufs, &conn->q2_blocked);
+            if (conn->q2_blocked) {
+                // note: unit tests grep for this log!
+                qd_log(log, QD_LOG_TRACE,
+                        "[C%"PRIu64"][L%"PRIu64"] client link blocked on Q2 limit",
+                        conn->conn_id, conn->incoming_id);
+            }
+            DEQ_INIT(conn->early_raw_read_bufs);
+        }
+
         conn->incoming_started = true;
 
         // Handle deferment of write side close.
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -129,15 +129,15 @@ static inline const char * qdr_link_direction_name(const qdr_link_t *link)
 static inline const char * qdr_tcp_connection_role_name(const qdr_tcp_connection_t *tc)
 {
     assert(tc);
-    return tc->ingress ? "client" : "server";
+    return tc->ingress ? "listener" : "connector";
 }
 
 static const char * qdr_tcp_quadrant_id(const qdr_tcp_connection_t *tc, const qdr_link_t *link)
 {
     if (tc->ingress)
-        return link->link_direction == QD_INCOMING ? "(client raw-to-rtr)" : "(client rtr-to-raw)";
+        return link->link_direction == QD_INCOMING ? "(listener incoming)" : "(listener outgoing)";
     else
-        return link->link_direction == QD_INCOMING ? "(server raw-to-rtr)" : "(server rtr-to-raw)";
+        return link->link_direction == QD_INCOMING ? "(connector outgoing)" : "(connector outgoing)";
 }
 
 static void on_activate(void *context)
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -342,21 +342,6 @@ static int handle_incoming(qdr_tcp_connection_t *conn, const char *msg)
                conn->conn_id, conn->incoming_id, conn->instream->delivery_id);
 
         conn->incoming_started = true;
-
-        // Handle deferment of write side close.
-        sys_mutex_lock(conn->activation_lock);
-        if (conn->read_eos_seen && !conn->raw_closed_write) {
-            // to-raw-conn EOS was seen before the from-raw-conn instream delivery existed.
-            // Now that the delivery has been created the write clousre can be sent.
-            conn->raw_closed_write = true;
-            sys_mutex_unlock(conn->activation_lock);
-            qd_log(log, QD_LOG_DEBUG,
-                "[C%"PRIu64"][L%"PRIu64"][D%"PRIu64"] EOS already seen. call pn_raw_connection_write_close",
-                conn->conn_id, conn->incoming_id, conn->instream->delivery_id);
-            pn_raw_connection_write_close(conn->pn_raw_conn);
-        } else {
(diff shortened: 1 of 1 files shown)


src/adaptors/tcp_adaptor.c
```diff
@@ -782,6 +965,14 @@ static void qdr_tcp_open_server_side_connection(qdr_tcp_connection_t* tc)
 
     // This attach passes the ownership of the delivery from the core-side connection and link
     // to the adaptor-side outgoing connection and link.
+    uint64_t i_conn_id = 0;
+    uint64_t i_link_id = 0;
+    uint64_t i_dlv_id = 0;
+    if (!!tc->initial_delivery) {
+        i_conn_id = tc->initial_delivery->conn_id;
```
grs, line 972: Minor point, but why are these initialised in a separate conditional block up here, when they seem only to be used in the subsequent block for the identical condition and are not used in the qdr_link_first_attach() call that is the only thing separating these blocks? Makes it easier to read if we just declared these just before use I think.

src/adaptors/tcp_adaptor.c
```diff
@@ -57,11 +57,15 @@ struct qdr_tcp_connection_t {
...
     bool                  egress_dispatcher;
     bool                  connector_closed;//only used if egress_dispatcher=true
     bool                  in_list;         // This connection is in the adaptor's connections list
-    bool                  raw_closed_read;
-    bool                  raw_closed_write;
+    bool                  raw_closed_read;   // proton event seen
+    bool                  raw_closed_write;  // proton event seen or write_close called
+    bool                  raw_read_shutdown; // stream closed
+    bool                  read_eos_seen;
+    qd_buffer_list_t      early_raw_read_bufs; // read from raw conn before ingress stream ready
```
grs, line 68: I'm not too keen on granting buffers before we are ready for them. It doesn't seem necessary to me and it complicates reasoning about the code.

src/adaptors/tcp_adaptor.c
```diff
@@ -630,64 +780,92 @@ static void handle_connection_event(pn_event_t *e, qd_server_t *qd_server, void
...
     case PN_RAW_CONNECTION_CLOSED_WRITE: {
-        qd_log(log, QD_LOG_DEBUG, "[C%"PRIu64"] PN_RAW_CONNECTION_CLOSED_WRITE", conn->conn_id);
+        qd_log(log, QD_LOG_DEBUG,
+               "[C%"PRIu64"] PN_RAW_CONNECTION_CLOSED_WRITE",
+               conn->conn_id);
         sys_mutex_lock(conn->activation_lock);
         conn->raw_closed_write = true;
         sys_mutex_unlock(conn->activation_lock);
-        pn_raw_connection_close(conn->pn_raw_conn);
+        if (conn->ingress) {
```
grs, line 799: Why only on ingress?

I would expect that on CLOSED_WRITE we settle the incoming qdr delivery, i.e. outstream (since we can no longer forward any data received). That signals the peer that they should close their socket for read.

src/adaptors/tcp_adaptor.c
```diff
@@ -1257,13 +1489,17 @@ static void qdr_tcp_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t
     void* link_context = qdr_link_get_context(qdr_delivery_link(dlv));
     if (link_context) {
         qdr_tcp_connection_t* tc = (qdr_tcp_connection_t*) link_context;
-        qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG, DLV_FMT" qdr_tcp_delivery_update: disp: %"PRIu64", settled: %s",
+        qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG,
+               DLV_FMT" qdr_tcp_delivery_update: disp: %"PRIu64", settled: %s",
                DLV_ARGS(dlv), disp, settled ? "true" : "false");
 
         //
         // If one of the streaming deliveries is ever settled, the connection must be torn down.
```
grs, line 1497: I think now that we have separate close calls for read and write sides, we should close for read if the instream is settled and close for write when the outstream is settled.

If we then settle and complete the instream on CLOSED_READ and settle the outstream on CLOSED_WRITE, we are then conveying more accurately what each tcp endpoint does to its peer. If the instream has not been started when we get CLOSED_READ on ingress, we can close the connection entirely (on egress we can just send an empty reply).

src/adaptors/tcp_adaptor.c
```diff
@@ -116,6 +120,26 @@ static inline uint64_t qdr_tcp_conn_linkid(const qdr_tcp_connection_t *conn)
...
+static inline const char * qdr_link_direction_name(const qdr_link_t *link)
+{
+    assert(link);
+    return qdr_link_direction(link) == QD_OUTGOING ? "outgoing" : "incoming";
+}
+
+static inline const char * qdr_tcp_connection_role_name(const qdr_tcp_connection_t *tc)
+{
+    assert(tc);
+    return tc->ingress ? "client" : "server";
```
grs, line 132: I find the use of client and server confusing here. What you call the client is actually the server side in tcp terms. I think ingress/egress is less ambiguous (perhaps qualified by bridge, e.g. 'ingress bridge') and is already used elsewhere even with this patch.

src/adaptors/tcp_adaptor.c
```diff
@@ -288,18 +338,87 @@ static int handle_incoming_impl(qdr_tcp_connection_t *conn, bool close_pending)
...
-        qd_log(tcp_adaptor->log_source, QD_LOG_DEBUG, "[C%"PRIu64"][L%"PRIu64"] Initiating message with %i bytes", conn->conn_id, conn->incoming_id, count);
+        qd_log(log, QD_LOG_DEBUG,
+               "[C%"PRIu64"][L%"PRIu64"][D%"PRIu64"] Initiating ingress stream message with %u bytes",
+               conn->conn_id, conn->incoming_id, conn->instream->delivery_id, length);
+        conn->incoming_started = true;
+
+        // Handle deferment of write side close.
+        sys_mutex_lock(conn->activation_lock);
+        if (conn->read_eos_seen && !conn->raw_closed_write) {
+            // to-raw-conn EOS was seen before the from-raw-conn instream delivery existed.
```
grs, line 349: Why do we need to wait for the instream delivery before calling write_close()?

grs (commented):
I confirm this works very well with ncat now, even in two router setup.

src/adaptors/tcp_adaptor.c
ChugR: This was on recommendation from proton developers: giving buffers to the read side is pretty harmless.

Reading from raw connection before the ingress stream is ready is the first strategy of several that has worked. Over the wire the client has executed open-write-close before the TCP adaptor has even been notified that the connection has been accepted. Then follows a READ event which TCP can't handle since it has no stream or credit. Then follows a READ_CLOSED which still can't be handled since there is no stream or credit. Eventually the stream comes up and credit arrives but these events don't arrive in the proper context for reading the raw connection. One could schedule a wake but complicated state is required to signal that on a random wake it's time to read from a closed connection.

src/adaptors/tcp_adaptor.c
ChugR: Ambiguity abounds no matter how it is described. I figure that a client originates a connection to a listener and that a connector originates a connection to a server. Clients never accept connections and servers never originate connections. 
Also, there was some bad code logic in these descriptions that caused the logs to say wrong things. An update is coming shortly.

src/adaptors/tcp_adaptor.c
ChugR: Because if the instream creation is too slow then close the raw connection becomes WRITE_CLOSED, then the server sends the response and closes it's connection, the raw connection becomes READ_CLOSED, and DISCONNECTED and everything is discarded.
All this happens while waiting for the instream to come up and get credit.
By deferring the write-close until the instream comes up then any server response has a place to go before the raw connection gets closed.
On a single router test this condition would show up after maybe 500 or 3000 successful tries. It was hard to track down, easy to fix.

src/adaptors/tcp_adaptor.c
ChugR: I will try removing that conditiona.

src/adaptors/tcp_adaptor.c
ChugR: qdr_link_first_attach conditionally "adjusts" the delivery's identity. This code is conditionally memorizing what the initial delivery had before the adjustment.
The log statement giving identical [Cx]{Lx][Dx] values before and after was pretty lame.

src/adaptors/tcp_adaptor.c
grs: > Clients never accept connections and servers never originate connections.

But that is not what the line above implies, is it? If ingress=true it is a connection managed by a TcpListener, i,e, it was an accepted connection.

src/adaptors/tcp_adaptor.c
ChugR: I disagree. This code patch is trying to fix DISPATCH-1878 where a client sends data and closes the connection. The expectation is that the data gets sent over the wire to where it's going and the propagates the close after that. Any data will sent by the server will come back to the client. Then when the server closes that close will be propagated to the client.
A tcpListener READ_CLOSED event must be propagated to the peer tcpConnector as a write_close() control. Then the connector READ_CLOSE event becomes the listener write_close() control.
Short circuiting that control flow will lose data.

src/adaptors/tcp_adaptor.c
grs: Sorry, I still don't get it. The i_dlv_id is not referenced in the qdr_link_first_attach call.

src/adaptors/tcp_adaptor.c
grs: I don't believe we would need a wake as we would be creating the instream on the io thread anyway. All that is needed is to drain the data available and append it to the instream.

We can't get a READ event if we don't grant any buffers anyway. Presumably the READ_CLOSED event must come after we have read all the data, else how would we know there was no more left. @astitcher ?

ChugR:
Adding a self test for this now

src/adaptors/tcp_adaptor.c
grs: That is exactly my point. The same is true for the other direction, a tcpListener CLOSED_WRITE should be propagated to the peer tcpConnector where it will call read_close(). The CLOSED_WRITE at the tcpConnector will be propagated to the tcpListener for it to call read_close(). Read and write streams are chained in both directions.

I argue that the comment above is actually short circuiting this (as it predates the API to handle read_close()/write_close() separately). Though it may work for the ncat case, that short circuiting seems less clean than a proper symmetric chaining.

src/adaptors/tcp_adaptor.c
grs: That sounds like the problem is that we handle CLOSED_WRITE by closing the whole connection, rather than simply relaying what the two endpoints themselves do.

ChugR: From a trace the other day.

```
TCP Adaptor half-closed hang after 1500 connections

 on the server side connection [C3110] is tcpConnector connection:
 * connection opens
 * client data 5 bytes is written
 * client connection stream to server is closed (EOS)
 * server connection does WRITE_CLOSE (same as original client does open-write-close to tcpListener)
 * server replies with 21 bytes
 * server caches the 21 bytes since "Waiting for credit before initiating ingress stream message"
   there's no message to put the bytes into yet
 * server application closes after writing the 21 byte reply
 
 * with server connection WRITE_CLOSED and READ_CLOSED the CONNECTION_DISCONNECTED comes along
 * handle_disconnected tears everything down and the cached bytes never get written into stream message
 
 * There's a bug since the server connection should stay up long enough to write the cached bytes to the stream message
 
TCP_ADAPTOR (debug) [C3110][L6219] (server outgoing) Created link to amqp:/_topo/0/D-1878/temp.RHkTIA2kerAComs
TCP_ADAPTOR (trace) [C3110] handle_incoming qdr_tcp_deliver for server connection. read_closed:F, flow_enabled:F
TCP_ADAPTOR (debug) [C3110][L6218] Waiting for credit before initiating ingress stream message
TCP_ADAPTOR (debug) [C3110] pn_raw_connection_write_buffers wrote 5 bytes
TCP_ADAPTOR (info) [C3110] EOS
    here the adaptor calls write_close() to server connection when EOS is seen
TCP_ADAPTOR (debug) [C3110] handle_outgoing calling pn_raw_connection_write_close(). rcv_complete:T, send_complete:T
TCP_ADAPTOR (debug) [C3110][L6218] qdr_tcp_offer: NOOP
TCP_ADAPTOR (debug) [C3110][L6218] qdr_tcp_get_credit: NOOP
TCP_ADAPTOR (debug) [C3110][L6218] (server outgoing) qdr_tcp_second_attach
TCP_ADAPTOR (debug) [C3110][L6218] qdr_tcp_get_credit: NOOP
TCP_ADAPTOR (debug) [C3110] PN_RAW_CONNECTION_CLOSED_WRITE
    adaptor connector to server is now write closed
TCP_ADAPTOR (debug) [C3110] PN_RAW_CONNECTION_WRITTEN pn_raw_connection_take_written_buffers wrote 5 bytes. Total written 5 bytes
TCP_ADAPTOR (debug) [C3110] pn_raw_connection_take_read_buffers() took 1, freed 0
TCP_ADAPTOR (debug) [C3110] PN_RAW_CONNECTION_READ Read 21 bytes. Total read 21 bytes
TCP_ADAPTOR (debug) [C3110] PN_RAW_CONNECTION_READ Read 0 bytes. Total read 21 bytes
    server echos the 5 bytes with a prefix to make it 21 bytes. Then server closes connection.
TCP_ADAPTOR (debug) [C3110][L6219] PN_RAW_CONNECTION_CLOSED_READ
TCP_ADAPTOR (trace) [C3110] handle_incoming PNRC_CLOSED_READ for server connection. read_closed:T, flow_enabled:F
    adaptor still has no place to put the 21 bytes
TCP_ADAPTOR (debug) [C3110][L6218] Waiting for credit before initiating ingress stream message
    raw connection is write- and read-closed. it shuts down and the 21 bytes never went to tcpListener client
TCP_ADAPTOR (info) [C3110] PN_RAW_CONNECTION_DISCONNECTED
TCP_ADAPTOR (debug) [C3110][L6218] handle_disconnected - close outstream
TCP_ADAPTOR (debug) [C3110][L6219] handle_disconnected - detach incoming
TCP_ADAPTOR (debug) [C3110][L6218] handle_disconnected - detach outgoing
TCP_ADAPTOR (debug) qdr_tcp_activate: no connection context
TCP_ADAPTOR (debug) [C3110] qdr_add_tcp_connection_CT 127.0.0.1:9090 (2)
TCP_ADAPTOR (debug) [C3110] qdr_del_tcp_connection_CT 127.0.0.1:9090 deleted. bytes_in=21, bytes_out=5, opened_time=28, last_in_time=28, last_out_time=28. Connections remaining 1
```

grs: If C3110  is on the tcpConnector, then it does not need to wait for reply-to in order to create the reply message as that was in the message it already received and read to EOS. If it does not have credit it should not have granted read buffers and so would not get the READ events until it did have credit.

src/adaptors/tcp_adaptor.c
ChugR: Maybe a solution would be not to accept an incoming client raw connection until after all of the TCP adaptor mechanisms have been set up. This would stall the initial connection setup but it would simplify propagating closed read/write in both directions when they happen.

grs: Readiness to read will always be a per-connection setup.

src/adaptors/tcp_adaptor.c
ChugR: Removing the conditional causes data loss on the connector side.
On a connector/server connection the sequence goes:

- EOS seen. Call pnrc_write_close() for connector connection
- PNRC_CLOSED_WRITE event
- Call pnrc_connection_close()  <-- called because conditional was removed
- PNRC_CLOSED_READ event
- PNRC_READ and PNRC_WRITTEN to recover buffers
- PNRC_CONNECTION_DISCONNECTED

If connection_close is not called on the connector then the write_close() propagates to the server, the server closes the connection, the connector-side of the adaptor receives the server's last data, and then CLOSED_READ comes in and the connection is shut.

ChugR: Getting rid of the conditional AND the pn_connection_close() call seems to work correctly. Just remove the code block entirely.

grs:
Looking good! I posted a patch with a few simplifications: https://github.com/ChugR/qpid-dispatch/pull/76. That passes all local tests and has got through more than 500,000 ncat tests without issue (still running fine).

ChugR:
Current code branch DISPATCH-1878-04 at commit 77765 has an issue when there is no server.
* On main branch the command _echo abcd | nc localhost 9999_ exits without printing anything
* With this PR the command hangs without printing anything
This issue should be corrected.

src/adaptors/tcp_adaptor.c
grs: Update on this. I was wrong. We don't actually explicitly settle deliveries except in the case where the connector cannot connect (i.e. disconnects before connecting), in which case the 'initial-delivery' is settled. Deliveries are implicitly settled if there is any failure. We convey the half clsoed state solely through completing the message (i.e. the EOS is the only trigger).

I've pushed a fix that handles the case where there is no server. Also tested router failures, including middle router in a three-way chain with tcp-listener and tcp-connector at opposite ends.

grs: https://github.com/ChugR/qpid-dispatch/pull/81

ChugR:
Problem: This branch hangs if there is no server
There is no useful comparison between main branch, which does not hang, and this branch. Main branch closes the client connection when the connection is closed for reading whether there is a server or not.
On the connector side server connection there are two events:

- pn_proactor_raw_connect() 
- DISCONNECTED event

The tcp connector side must go through enough initialization so that the listener-connector links are established. Then the connector side can tear down the links and the client connection will close.

grs:
> Problem: This branch hangs if there is no server

I believe I have a fix for this in https://github.com/ChugR/qpid-dispatch/pull/81

> On the connector side server connection there are two events:
> 
> * pn_proactor_raw_connect()
> * DISCONNECTED event
> 
> The tcp connector side must go through enough initialization so that the listener-connector links are established. Then the connector side can tear down the links and the client connection will close.

If the DISCONNECTED happens before CONNECTED, then the initial_delivery is released. The fix above ensures this is handled correctly and the connection is closed.

PR #1132, opened 2021-04-16
DISPATCH-2048: Fix shutdown leak of qd_deferred_call_t

ganeshmurthy:


kgiusti (approved):
Confirmed: no qd_deferred_call_t leaks in any of the CI test runs

PR #1118, opened 2021-04-14
DISPATCH-2039 Poison the memory pool so that ASAN works better with it

jiridanek:
TODO:
* does this actually do the right thing? Edit: looks right to me.
* is it necessary to instrument `unordered_move_stack` function? Edit: me thinks no, it only changes linked list metadata, it does not touch the (empty) data.

When there is a read or write to the poisoned area, the error message looks like

```
==15792==ERROR: AddressSanitizer: use-after-poison on address 0x611000034dd8 at pc 0x7fdaa75fc713 bp 0x7fff2d0c8d80 sp 0x7fff2d0c8d78
14: WRITE of size 8 at 0x611000034dd8 thread T0
14:     #0 0x7fdaa75fc712 in qd_hash_internal_remove_item /home/jdanek/repos/qpid/qpid-dispatch/cmake-build-debug-asan/../src/hash.c:131:30
14:     #1 0x7fdaa75fb51d in qd_hash_free /home/jdanek/repos/qpid/qpid-dispatch/cmake-build-debug-asan/../src/hash.c:146:13
[...]
```

PR #1135, opened 2021-04-18
DISPATCH-2039 WIP Instead of poison, set zero capacity of the pool and let it degenerate into immediate malloc/free

jiridanek:
This is an alternative approach compared to #1118. From looking at the (many) errors it produces, it is just as well able to reveal issues

* https://issues.apache.org/jira/browse/DISPATCH-2045 qd_hash_internal_remove_item writes to freed (pooled) memory on router shutdown
* https://issues.apache.org/jira/browse/DISPATCH-2056 AddressSanitizer: use-after-poison in qdr_connection_set_context during system_tests_http2

It is able to also find the following, that's why I had to put the no_sanitize attribute there

* https://issues.apache.org/jira/browse/DISPATCH-2060 use-after free in qd_alloc_deref_safe_ptr if a pool item has been freed due to global_free_list size limit

The stacktraces produced from malloc/free are much nicer than the ones from poisoning. The poisoning report is bare-bones, whereas with malloc/free
1. there is a stacktrace for the free call
2. the allocation stack trace is related to this particular use of the pool item, not simply the trace from when the item was first allocated if now it is being reused

### DIFF
--- a/src/alloc_pool.c
+++ b/src/alloc_pool.c
@@ -262,8 +262,9 @@ struct qd_alloc_pool_t {
     qd_alloc_linked_stack_t free_list;
 };
 
-qd_alloc_config_t qd_alloc_default_config_big   = {16,  32, 0};
-qd_alloc_config_t qd_alloc_default_config_small = {64, 128, 0};
+qd_alloc_config_t qd_alloc_default_config_big   = {16,  32, -1};
+qd_alloc_config_t qd_alloc_default_config_small = {64, 128, -1};
+qd_alloc_config_t qd_alloc_default_config_asan  = { 1,   0,  0};
 #define BIG_THRESHOLD 2000
 
 static sys_mutex_t          *init_lock = 0;
@@ -279,11 +280,12 @@ static void qd_alloc_init(qd_alloc_type_desc_t *desc)
         if (desc->additional_size)
             desc->total_size += *desc->additional_size;
 
-        if (desc->config == 0)
-            desc->config = desc->total_size > BIG_THRESHOLD ?
-                &qd_alloc_default_config_big : &qd_alloc_default_config_small;
-
-        assert (desc->config->local_free_list_max >= desc->config->transfer_batch_size);
+        desc->config = &qd_alloc_default_config_asan;
+//        if (desc->config == 0)
+//            desc->config = desc->total_size > BIG_THRESHOLD ?
+//                &qd_alloc_default_config_big : &qd_alloc_default_config_small;
+//
+//        assert (desc->config->local_free_list_max >= desc->config->transfer_batch_size);
 
         desc->global_pool = NEW(qd_alloc_pool_t);
         DEQ_ITEM_INIT(desc->global_pool);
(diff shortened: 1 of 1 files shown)


src/alloc_pool.c
```diff
@@ -280,11 +282,12 @@ static void qd_alloc_init(qd_alloc_type_desc_t *desc)
...
-            desc->config = desc->total_size > BIG_THRESHOLD ?
-                &qd_alloc_default_config_big : &qd_alloc_default_config_small;
-
-        assert (desc->config->local_free_list_max >= desc->config->transfer_batch_size);
+        desc->config = &qd_alloc_default_config_asan;
+//        if (desc->config == 0)
+//            desc->config = desc->total_size > BIG_THRESHOLD ?
+//                &qd_alloc_default_config_big : &qd_alloc_default_config_small;
+//
+//        assert (desc->config->local_free_list_max >= desc->config->transfer_batch_size);
```
jiridanek, line 290: I think the assert is wrong; `local_free_list_max == 0` and `transfer_batch_size == 1` works just fine.

src/alloc_pool.c
```diff
@@ -280,11 +282,12 @@ static void qd_alloc_init(qd_alloc_type_desc_t *desc)
...
 
-        if (desc->config == 0)
-            desc->config = desc->total_size > BIG_THRESHOLD ?
-                &qd_alloc_default_config_big : &qd_alloc_default_config_small;
-
-        assert (desc->config->local_free_list_max >= desc->config->transfer_batch_size);
+        desc->config = &qd_alloc_default_config_asan;
+//        if (desc->config == 0)
+//            desc->config = desc->total_size > BIG_THRESHOLD ?
+//                &qd_alloc_default_config_big : &qd_alloc_default_config_small;
```
jiridanek, line 288: There will have to be a way to enable this from CMake; for now I am doing the change this way.

src/alloc_pool.c
```diff
@@ -495,7 +498,7 @@ void qd_dealloc(qd_alloc_type_desc_t *desc, qd_alloc_pool_t **tpool, char *p)
     // If there's a global_free_list size limit, remove items until the limit is
     // not exceeded.
     //
-    if (desc->config->global_free_list_max != 0) {
+    if (desc->config->global_free_list_max != -1) {
```
jiridanek, line 501: 0 used to mean "unlimited". I need 0 to mean 0.

jiridanek:
This happens all the time on this Xenial build, https://travis-ci.com/github/apache/qpid-dispatch/jobs/499988187

```
14: Process 12186 error: exit code -6, expected -1
14: qdrouterd -c A.conf -I /home/travis/build/apache/qpid-dispatch/python
14: /home/travis/build/apache/qpid-dispatch/build/tests/system_test.dir/system_tests_link_routes/EmptyTransferTest/setUpClass/A-6.cmd
14: >>>>
14: qdrouterd: /home/travis/build/apache/qpid-dispatch/src/alloc_pool.c:525: qd_alloc_sequence: Assertion `item->header == PATTERN_FRONT' failed.
14: <<<<
14: 
14: ----------------------------------------------------------------------
14: Ran 40 tests in 127.945s
14: 
14: FAILED (errors=1)
14/72 Test #14: system_tests_link_routes ..........................***Failed  128.03 sec
```

jiridanek:
I thought if I can carry this forward somewhat, but so far I don't see a way. I'll close this the next time I revisit.

jiridanek:
I don't feel ready do give up on this. There are some use-after-poison jiras that might benefit from free-call stracktrace obtained by this approach.

jiridanek:
https://issues.apache.org/jira/browse/DISPATCH-2095

PR #1137, opened 2021-04-19
DISPATCH-2047: Moved setting the deliveries_ingress to qdr_link_deliv…

ganeshmurthy:
…er_CT. Also fixed test to try the custom timer a few more times

### DIFF
--- a/src/router_core/transfer.c
+++ b/src/router_core/transfer.c
@@ -742,7 +742,7 @@ static void qdr_link_deliver_CT(qdr_core_t *core, qdr_action_t *action, bool dis
 
     link->total_deliveries++;
 
-    if (link->link_type == QD_LINK_ENDPOINT)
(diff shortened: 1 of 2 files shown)


jiridanek:
Similar issue as before fixed in https://issues.apache.org/jira/browse/DISPATCH-1796 , PR #873

src/router_core/transfer.c
```diff
@@ -745,6 +742,9 @@ static void qdr_link_deliver_CT(qdr_core_t *core, qdr_action_t *action, bool dis
 
     link->total_deliveries++;
 
+    if (link->link_type == QD_LINK_ENDPOINT)
```
kgiusti, line 745: Should we also move the check for fallback here as well:

... ENDPOINT && !dlv_link->fallback)  ???

kgiusti (approved):

PR #1138, opened 2021-04-20
DISPATCH-2052: Partial test fix. Make sure the address has propagated…

ganeshmurthy:
… to Router A before starting to send messages


ganeshmurthy:
Only a partial fix, DO NOT MERGE

PR #1139, opened 2021-04-20
DISPATCH-2032: Clean out peer->link_work on shutdown

ganeshmurthy:


kgiusti (commented):
link_work items are reference counted by the deliveries that point to them via the ->link_work pointer.

Simply zero'ing out the pointer will leave the link_work item dangling.

We need to fix this by ensuring all the code behaves properly when it comes to link_work lifecycle.

kgiusti:
See https://github.com/apache/qpid-dispatch/commit/d68ceff4a767ddec84fa044ecc8bb3d4a609c5d5 for a reference count based fix.

PR #1146, opened 2021-04-21
DISPATCH-2055: Eliminated log_lock and used the log_source_lock instead

ganeshmurthy:


src/log.c
```diff
@@ -222,27 +222,36 @@ static level_t levels[] = {
...
+        if (log_error)
+            qd_error(QD_ERROR_CONFIG, "'%d' is not a valid log level bit.", bit);
+        return &invalid_level;
     }
     return &levels[i];
 }
 
+static bool is_log_level_invalid(const level_t *level)
+{
+    return ! strcmp(level->name, "invalid");
```
ChugR, line 243: A better (faster) comparison might be 
    level->bit != -2 
Then inline it.

src/log.c
```diff
@@ -222,27 +222,36 @@ static level_t levels[] = {
     LEVEL("critical", QD_LOG_CRITICAL, LOG_CRIT)
 };
 
+static const level_t invalid_level = {"invalid", -2, -2, 0};
+
 static char level_names[TEXT_MAX] = {0}; /* Set up in qd_log_initialize */
 
 /// Return NULL and set qd_error if not a valid bit.
-static const level_t* level_for_bit(int bit) {
+static const level_t* level_for_bit(int bit, bool log_error) {
```
kgiusti, line 230: IIUC this level_for_bit is only called by write_log(), which passes log_error=false.
So why add the bool at all?  Simply remove the qd_error since it will never be called.

src/log.c
```diff
@@ -425,7 +434,6 @@ void qd_vlog_impl(qd_log_source_t *source, qd_log_level_t level, bool check_leve
     }
 
     // Bounded buffer of log entries, keep most recent.
-    sys_mutex_lock(log_lock);
     qd_log_entry_t *entry = new_qd_log_entry_t();
     DEQ_ITEM_INIT(entry);
     entry->module = source->module;
```
ChugR, line 439: source->module is a char*. Copying the pointer to entry leaves it vulnerable to deletion as described in the comment at line 446. This string should be copied into entry the same way that entry->file is strdup'd on line 441. Then the entry is not depending on source still existing the whole time entry is in the entries list (line 542).

src/log.c
```diff
@@ -271,7 +280,7 @@ static const char* level_name(int level) {
 static const char *SEPARATORS=", ;:";
 
 /// Calculate the bit mask for a log enable string. Return -1 and set qd_error on error.
-static int enable_mask(const char *enable_) {
+static int enable_mask(const char *enable_, bool log_error) {
```
kgiusti, line 283: This also appears to only be called with log_error=false, which makes the log_error parameter & logic not necessary.

src/log.c
```diff
@@ -525,71 +538,143 @@ void qd_log_finalize(void) {
...
+        // A module attribute MUST be specified for a log entity.
+        // Every other attribute is optional.
+        //
         module = qd_entity_get_string(entity, "module");
+
+        //
+        // If the module is not specified, there is nothing to do, just log an
+        // error and break out of this do loop.
+        //
         QD_ERROR_BREAK();
```
kgiusti, line 571: IIRC these QD_ERROR_BREAK's force a jump to the end of the while (0) loop.
At the end of the while (0) loop there is a mutex unlock call, which will be unlocking a mutex which is not locked.

src/log.c
```diff
@@ -222,27 +222,36 @@ static level_t levels[] = {
...
 }
 
+static bool is_log_level_invalid(const level_t *level)
+{
+    return ! strcmp(level->name, "invalid");
+}
+
 /// Return NULL and set qd_error if not a valid level.
-static const level_t* level_for_name(const char *name, int len) {
+static const level_t* level_for_name(const char *name, int len, bool log_error) {
```
kgiusti, line 247: same here, log_error is essentially unused.

kgiusti (changes requested):
I think those BREAK macros in qd_log_entity() may cause a call to sys_mutex_unlock on a mutex that was never taken...

src/log.c
ganeshmurthy: I am proposing that we just move the lock back to where it originally was, like just before this line - qd_log_entry_t *entry = new_qd_log_entry_t();

kgiusti (approved):

PR #1147, opened 2021-04-21
DISPATCH-2033: Wait for both routers to connect to each other before …

ganeshmurthy:
…running qdstat

### DIFF
--- a/tests/system_tests_qdstat.py
+++ b/tests/system_tests_qdstat.py
@@ -565,6 +565,7 @@ class QdstatLinkPriorityTest(system_test.TestCase):
         cls.router_2 = cls.tester.qdrouterd('test_router_2', config_2, wait=True)
         cls.router_1 = cls.tester.qdrouterd('test_router_1', config_1, wait=True)
         cls.router_1.wait_router_connected('R2')
+        cls.router_2.wait_router_connected('R1')
 
     def address(self):
         return self.router_1.addresses[0]


tests/system_tests_qdstat.py
```diff
@@ -565,6 +565,7 @@ def setUpClass(cls):
         cls.router_2 = cls.tester.qdrouterd('test_router_2', config_2, wait=True)
         cls.router_1 = cls.tester.qdrouterd('test_router_1', config_1, wait=True)
         cls.router_1.wait_router_connected('R2')
+        cls.router_2.wait_router_connected('R1')
```
ChugR, line 568: That's a surprising fix. If 1 is connected to 2 then how can 2 not be connected to 1?
Not to say that this is not a fix but maybe it's just the time spent executing the wait allows for some other thing to stabilize and not that 2 was not connected to 1.

ChugR (approved):

tests/system_tests_qdstat.py
ganeshmurthy: If you look here - https://github.com/apache/qpid-dispatch/blob/main/tools/qdstat.in#L1052
The has_nodes() function returns False if the number of nodes is less than 2. And the string "UTC" is being displayed if has_nodes() is True, hence this PR. Also take a look at this output when we print the output of qdstat, one of the routers is missing. That being said, I have been unable to reproduce this problem locally and based on the evidence, I am hoping that this PR would fix the problem.

ganeshmurthy: I saw this new failure - https://travis-ci.com/github/apache/qpid-dispatch/jobs/500203449#L3640
that looks not related to this fix. Closing this PR

PR #1153, opened 2021-04-22
DO NOT MERGE - extend travis timeouts hugely

kgiusti:


ganeshmurthy (approved):

.travis.yml
```diff
@@ -204,7 +212,7 @@ script:
 - popd
 # Workaround on macOS for PROTON-808 Binaries have their library locations stripped
 - if [[ "${OSTYPE}" == "darwin"* ]]; then install_name_tool -add_rpath $PREFIX/lib/. $PREFIX/lib/proton/bindings/python/_cproton.so; fi
-- ctest -V ${DISPATCH_CTEST_EXTRA} && if [ "$BUILD_TYPE" = "Coverage" ]; then cmake --build . --target coverage; fi
+- ctest --timeout 1200 -V ${DISPATCH_CTEST_EXTRA} && if [ "$BUILD_TYPE" = "Coverage" ]; then cmake --build . --target coverage; fi
```
jiridanek, line 215: Let's not forget that there is a Travis timeout which kills a build if it has not produced any output for 10 minutes. With a big timeout like this, what can happen is that if a test gets stuck, the first thing that triggers is the Travis timeout and that kills the entire CI job.

Solution would be to play with timeouts more, or run a background process that is printing dots every 5 minutes or so... Lets see if we actually hit this problem.

kgiusti: Good point.  The 300 second (5 minute) per-test timeout would trigger first I believe, since each test prints something at the start and end.

PR #1158, opened 2021-04-24
DISPATCH-1959 Update PatternFly packages for the 1.16.0 release

jiridanek:


jiridanek:
This fixes an issue that `npm audit` does not catch, which happens to be a CVE in a transitive dev-only dependency package is-svg. https://nvd.nist.gov/vuln/detail/CVE-2021-28092

PR #1159, opened 2021-04-25
DISPATCH-2078 Fix GitHub Actions SASL issue by recompiling Cyrus SASL and using fqdn

jiridanek:


jiridanek:
Doesn't work on Travis. Of course, I should've expected a gotcha...

jiridanek:
This works on GHA but not on Travis (the fqdn part). And rebuilding Sasl as part of CI is a bad idea. Closing.

PR #1160, opened 2021-04-25
DISPATCH-2078 Fix GitHub Actions SASL issue by using fqdn in sasldb

jiridanek:


jiridanek:
Funny thing. This sometimes works to fix the problem, and sometimes doesn't. See e.g. https://github.com/jiridanek/qpid-dispatch/runs/2432658131 for a passing run. The PR #1159 seems to work reliably, in contrast.

gemmellr:
GitHub Actions jobs have had some hostname/dns issues in the alst couple weeks, though they are meant to be fixed now: https://github.com/actions/virtual-environments/issues/3185 (though could the fix be related to the issue?)

gemmellr:
well, perhaps not if you find recompiling Cyrus helps? Did the version of Cyrus change between older builds that work, and new ones that dont?

jiridanek:
Cyrus did not change.It's the same version since 2018 or so. There are patches in distros that do change, but Ubuntu did not update anything recently, as I was able to tell.

I would be a bit mad if this turns out to be env issue resolving hostname. I almost learned to use gdb investigating this...

Lets see how #1167 runs.

gemmellr:
Yeah, I dont know if it will help, I just thought it seemed potentially related. I had just saw there was a thread on builds@a.o about GitHub Actions host resolution, and then almost immediately after that noticed your JIRA/PR updates here mentioning FQQN settings relating to test failures, and thought....hmmm....

jiridanek:
I did not find the GHA issue before, so thanks!

Ok, now I probably should try #1167 and this #1160 together. Except, I can't add `'-u', fqdn` to saslpasswd2, because that breaks Travis. And I absolutely don't want to also recompile Cyrus on Travis (assuming that fixes the problem), because then it's no longer testing for functionality on stock Ubuntu. So my best bet really is disabling test on GHA and hoping that something somewhere in the depths of universe will shift again and the test goes back to working...

jiridanek:
The next principled step in resolving this is probably for me to figure out why `log_level: 7` does nothing for me. Better logs would help. Maybe strace (that works on GHA, rr doesn't). Or possibly gdb with watchpoints (that print info, don't stop the program), Or bpf probe, that can log function arguments.

PR #1167, opened 2021-04-26
DISPATCH-2078 Fix GitHub Actions SASL issue by allowing reverse-hostname resolution through /etc/hosts

jiridanek:


jiridanek:
Doesn't work. I really thought this will be it :(

jiridanek:
Does not work, closing.

PR #1169, opened 2021-04-27
NO-JIRA: Skip the GHA Docs job until Proton 0.34.0 Ubuntu package is published

jiridanek:
Ubuntu packages with required Proton 0.34.0 are not yet built. So the job is now failing.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -239,7 +239,7 @@ jobs:
   docs:
     name: 'Docs (${{ matrix.os }})'
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


gemmellr:
Why does this bit need to use a package when the test etc bits have already built proton?

jiridanek:
I came to this _wanting_ to use the Ubuntu deb package. It avoids dependency between the jobs, they can run concurrently. And I got discouraged from pursuing a different way of doing things (I am no longer sure what exactly it was) after hitting https://issues.apache.org/jira/browse/DISPATCH-1995.

PR #1170, opened 2021-04-27
DISPATCH-1679: Make qd_timer_cancel and qd_timer_free synchronous

kgiusti:
This patch makes the qd_timer_cancel() and qd_timer_free() routines
block if the timer callback is currently running on another thread.
This ensures that the callback is not executing when these calls
return.

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -25,7 +25,7 @@
 
 #include "qpid/dispatch/alloc.h"
 #include "qpid/dispatch/ctools.h"
-#include "qpid/dispatch/threading.h"
+#include "qpid/dispatch/atomic.h"
 
 #include <assert.h>
 #include <stdio.h>
--- a/src/timer.c
+++ b/src/timer.c
@@ -31,9 +31,53 @@
 #include <stdio.h>
 #include <time.h>
 
+
+// timer state machine
+//
+// IDLE: initial state or state immediately after the callback finishes
+// running.
+//
+// SCHEDULED: timer has been scheduled to run.  It is on the scheduled_timer
+// list.  Valid next states are IDLE (if timer canceled), RUNNING, or DELETED.
+//
+// RUNNING: the timer callback is executing.  Valid next states are IDLE,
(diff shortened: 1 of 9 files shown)


src/timer.c
```diff
@@ -129,8 +214,9 @@ qd_timestamp_t qd_timer_now() {
 void qd_timer_schedule(qd_timer_t *timer, qd_duration_t duration)
 {
     sys_mutex_lock(lock);
-    timer_cancel_LH(timer);  // Timer is now on the idle list
-    DEQ_REMOVE(idle_timers, timer);
+
```
ganeshmurthy, line 217: If timer state is indeed QD_TIMER_STATE_DELETED should we return immediately ?

ganeshmurthy (approved):

src/timer.c
kgiusti: If state is DELETED that indicates a bug where a codepath is attempting to schedule a timer that has been deleted.  Returning immediately would mask that bug - better to assert().

src/timer.c
```diff
@@ -110,16 +174,37 @@ qd_timer_t *qd_timer(qd_dispatch_t *qd, qd_timer_cb_t cb, void* context)
...
+    assert(timer->state != QD_TIMER_STATE_DELETED);  // double free!!!
+
+    if (timer->state == QD_TIMER_STATE_RUNNING) {
+        if (sys_thread_self() != callback_thread) {
+            // Another thread is running the callback (see qd_timer_visit())
+            // Wait until the callback finishes
+            timer->state = QD_TIMER_STATE_BLOCKED;
+            sys_cond_wait(timer->condition, lock);
```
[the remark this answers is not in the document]
kgiusti: Timers are executed sequentially, not in parallel.  This is by design to avoid races between timer callbacks.  A 10 second timer handler would cause all timers scheduled to expire after it to be blocked for 10 seconds and miss their deadlines regardless of this patch.

Timer callbacks are meant to run quickly and should not block.  Long running events must be scheduled as background tasks.

PR #1174, opened 2021-04-28
DISPATCH-2088: Free message_stream_data objects on connection thread

ChugR:
and not on the core thread. Move flush_outgoing_bufs call to
handle_disconnected function.


mgoulish:
DEADBUG

Before this patch I had failure rate of 11.8%
Now I have zero failures in 100 tries.   Odd of happening randomly are 1 in 300,000.
Ship it.

kgiusti (approved):
Oh yeah, probably not a good idea to flush buffers from the core thread  :+1:

PR #1176, opened 2021-04-29
DISPATCH-2046: prevent crash when accessing connector pointer

kgiusti:
DISPATCH-1679: fix qd_connector_t leak
DISPATCH-1917: fix race when accessing connector->conn_msg buffer

### DIFF
--- a/src/alloc_pool.c
+++ b/src/alloc_pool.c
@@ -90,7 +90,7 @@ DEQ_DECLARE(qd_alloc_type_t, qd_alloc_type_list_t);
 #if QD_MEMORY_STATS
 static const char *leaking_types[] = {
     // DISPATCH-1679:
-    "qd_timer_t", "qd_connector_t",
+    //"qd_timer_t", "qd_connector_t",
     "qd_hash_handle_t",       // DISPATCH-1696
 
     // system_tests_edge_router (centos7)
--- a/src/alloc_pool.c
+++ b/src/alloc_pool.c
@@ -89,8 +89,6 @@ DEQ_DECLARE(qd_alloc_type_t, qd_alloc_type_list_t);
 //
 #if QD_MEMORY_STATS
 static const char *leaking_types[] = {
-    // DISPATCH-1679:
-    //"qd_timer_t", "qd_connector_t",
     "qd_hash_handle_t",       // DISPATCH-1696
 
     // system_tests_edge_router (centos7)
--- a/src/connection_manager.c
+++ b/src/connection_manager.c
@@ -783,16 +783,16 @@ static int get_failover_info_length(qd_failover_item_list_t   conn_info_list)
  */
 qd_error_t qd_entity_refresh_connector(qd_entity_t* entity, void *impl)
 {
-    qd_connector_t *ct = (qd_connector_t*) impl;
+    qd_connector_t *connector = (qd_connector_t*) impl;
 
-    int conn_index = ct->conn_index;
+    int conn_index = connector->conn_index;
 
     int i = 1;
     int num_items = 0;
 
-    sys_mutex_lock(ct->lock);
+    sys_mutex_lock(connector->lock);
 
-    qd_failover_item_list_t   conn_info_list = ct->conn_info_list;
+    qd_failover_item_list_t   conn_info_list = connector->conn_info_list;
 
     int conn_info_len = DEQ_SIZE(conn_info_list);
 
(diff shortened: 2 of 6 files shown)


src/server.c
```diff
@@ -1644,25 +1657,31 @@ void qd_listener_decref(qd_listener_t* li)
...
-        return 0;
-    }
-    return ct;
+    qd_connector_t *connector = new_qd_connector_t();
+    if (!connector) return 0;
+    ZERO(connector);
+    sys_atomic_init(&connector->ref_count, 1);
+    DEQ_INIT(connector->conn_info_list);
+
+    connector->server  = server;
```
ganeshmurthy, line 1666: How about initializing the timer, conn_msg, lock *first* and if those are successful, then set server, conn_index, etc.

src/connection_manager.c
```diff
@@ -859,16 +861,23 @@ qd_error_t qd_entity_refresh_connector(qd_entity_t* entity, void *impl)
...
       break;
     }
 
     if (qd_entity_set_string(entity, "failoverUrls", failover_info) == 0
-            && qd_entity_set_string(entity, "connectionStatus", state_info) == 0
-            && qd_entity_set_string(entity, "connectionMsg", ct->conn_msg) == 0)
+        && qd_entity_set_string(entity, "connectionStatus", state_info) == 0
+        && qd_entity_set_string(entity, "connectionMsg", ct->conn_msg) == 0) {
+
+        sys_mutex_unlock(ct->lock);
```
ganeshmurthy, line 876: can we change this to connector->lock ? In some places we have ct->lock and in some places we have connector->lock. The connector->lock is easier to understand IMO.

src/server.c
```diff
@@ -1644,25 +1657,31 @@ void qd_listener_decref(qd_listener_t* li)
...
+    connector->timer = qd_timer(connector->server->qd, try_open_cb, connector);
+    if (!connector->timer)
+        goto error;
+    connector->conn_msg = (char*) malloc(300);
+    if (!connector->conn_msg)
+        goto error;
+    memset(connector->conn_msg, 0, 300);
+    return connector;
+
+error:
```
ganeshmurthy, line 1681: Can we please remove "goto" if you don't feel strongly about it?

src/server.c
```diff
@@ -1155,56 +1156,55 @@ static qd_failover_item_t *qd_connector_get_conn_info(qd_connector_t *ct) {
...
-        /* No longer referenced by pn_connection or timer */
-        qd_connector_decref(ct);
+    // Allocate a new connection. No other thread will touch this
+    // connection until pn_proactor_connect is called below
+    qd_connection_t *qd_conn = qd_server_connection(connector->server, &connector->config);
+    if (!qd_conn) {                 /* Try again later */
+        qd_log(connector->server->log_source, QD_LOG_CRITICAL, "Allocation failure connecting to %s",
+               connector->config.host_port);
+        connector->delay = 10000;
+        connector->state = CXTR_STATE_CONNECTING;
```
ganeshmurthy, line 1168: Logically speaking, should the state here be CXTR_STATE_FAILED ? After all the qd_server_connection() failed to allocate a new qd_connection_t object. When the next timer fires after 10 seconds, it will set the state to CXTR_STATE_CONNECTING, right?

src/alloc_pool.c
```diff
@@ -90,7 +90,7 @@ DEQ_DECLARE(qd_alloc_type_t, qd_alloc_type_list_t);
 #if QD_MEMORY_STATS
 static const char *leaking_types[] = {
     // DISPATCH-1679:
-    "qd_timer_t", "qd_connector_t",
+    //"qd_timer_t", "qd_connector_t",
```
ganeshmurthy, line 93: Just delete the two lines. We could add a link in the JIRA DISPATCH-2046 about DISPATCH-1679

src/connection_manager.c
```diff
@@ -790,6 +790,8 @@ qd_error_t qd_entity_refresh_connector(qd_entity_t* entity, void *impl)
     int i = 1;
     int num_items = 0;
 
+    sys_mutex_lock(ct->lock);
+
```
ganeshmurthy, line 794: If the connector state is CXTR_STATE_DELETED, should we just return immediately? Why loop thru the conn_info_list and so on? When the user runs a qdmanage QUERY --type=connector, I think the deleted connector (CXTR_STATE_DELETED) should not even show up in the list of connectors.

PR #1178, opened 2021-04-29
DISPATCH-2091: disable TCP backpressure when delivery is settled

kgiusti:


src/adaptors/tcp_adaptor.c
```diff
@@ -231,7 +232,7 @@ static int handle_incoming_raw_read(qdr_tcp_connection_t *conn, qd_buffer_list_t
     int free_count = 0;
     const bool was_open = conn->bytes_unacked < TCP_MAX_CAPACITY;
 
-    while ((count + conn->bytes_unacked < TCP_MAX_CAPACITY)
+    while ((conn->backpressure_disabled || (count + conn->bytes_unacked < TCP_MAX_CAPACITY))
```
ChugR, line 235: This is probably the only line of code that needs to change to eliminate back pressure problem. See https://github.com/apache/qpid-dispatch/pull/1177

The theory on PR1177 is that when the sending end of a TCP streaming delivery settles then that settlement is reflected in the far end of the stream as an EOS on it's stream. The TCP on the far end of the stream issues a raw connection close_write to half-close the raw connection. Further, the half-closed condition automatically disables the TCP_MAX_CAPACITY back pressure and drains that raw connection's incoming stream.

ChugR (changes requested):

src/adaptors/tcp_adaptor.c
kgiusti: Cool.  I'll "unpropose" this patch and we'll got with #1177 instead.

PR #1182, opened 2021-05-04
DISPATCH-2098 - Stop HTTP server before the AMQP server

jiridanek:
The HTTP server is weird, I can't find an ideal place where to stop it. It is somewhat like the adapter modules, but it is implemented differently. This here survives the Ctrl+C without a crash. I'll try to add a test. Worst case, the commit that broke this can be reverted; that would solve the problem too.

### DIFF
--- a/src/dispatch.c
+++ b/src/dispatch.c
@@ -363,6 +363,10 @@ static void qd_dispatch_set_router_area(qd_dispatch_t *qd, char *_area) {
 void qd_dispatch_free(qd_dispatch_t *qd)
 {
     if (!qd) return;
+
+    /* Stop HTTP threads immediately */
+    qd_http_server_free(qd_server_http(qd->server));
+
     free(qd->sasl_config_path);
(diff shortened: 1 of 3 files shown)


ganeshmurthy:
You could also call the qd_http_server_free() function directly from qd_dispatch_free(qd_dispatch_t *qd) by adding this function in server_private.h
qd_http_server_t *qd_server_http(qd_server_t *server);

And implement that in server.c 
qd_http_server_t *qd_server_http(qd_server_t *server)
{
       return server->http;
}

Call it directly from like this qd_dispatch_free like this - 
qd_http_server_free(qd_server_http(qd->server));

jiridanek:
@ganeshmurthy should I change this to the way you suggest?

ganeshmurthy:
> @ganeshmurthy should I change this to the way you suggest?
Sure, thanks.

ganeshmurthy (approved):

PR #1185, opened 2021-05-05
DISPATCH-2033: Wait for routers to talk to each other before starting…

ganeshmurthy:
… the tests


ganeshmurthy:
This is not a fix for DISPATCH-2033 (I have been unable to reproduce it so far), but since it runs qdstat commands for all routers, it is important that the routers know about each other before the test starts.

kgiusti (approved):

PR #1188, opened 2021-05-05
NO-JIRA: Comment out system_tests_http1_adaptor, system_tests_http2 a…

ganeshmurthy:
…nd system_tests_grpc


jiridanek (approved):
I'm thinking about adding a new CI job that runs only the now-disabled tests with ctest parameter `--repeat until-pass:3`. That way, CI would still catch a situation when test becomes broken. But that is for possible future enhancement.

jiridanek:
> ```diff
> - Coverage   84.31%   84.30%   -0.01%     
> ```

This is weird, I'd expect a bigger change resulting from this....

kgiusti (approved):

PR #1192, opened 2021-05-06
Dispatch 2100 self test for TCP half-closed handling

ChugR:
Echo server was failing half-closed in its own right. Repair that to flush received data before closing.
Conditionally, upon presence of ncat utility, run ncat as described in https://issues.apache.org/jira/browse/DISPATCH-1878 to trigger half-closed handling in TCP adaptor.


ChugR:
Fixed at commit 561306a4

PR #1196, opened 2021-05-07
DISPATCH-1595: Remove all python test on_modified event handlers

ChugR:
There is no on_modified event in proton python MessagingHandler.
Proton conflates Delivery.MODIFIED and Delivery.RELEASED state handling
and delivers both states in the on_released handler.
    
This patch makes the dispatch handler model precisely reflect the
proton handler model.


ChugR:
Whoops. bad pr for sure

PR #1197, opened 2021-05-07
DISPATCH-1595: Remove all python test on_modified event handlers

ChugR:
There is no on_modified event in proton python MessagingHandler.
Proton conflates Delivery.MODIFIED and Delivery.RELEASED state handling
and delivers both states in the on_released handler.

This patch makes the dispatch handler model precisely reflect the
proton handler model.

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -1107,15 +1107,10 @@ class AsyncTestSender(MessagingHandler):
         event.delivery.settle()
 
     def on_released(self, event):
-        # for some reason Proton 'helpfully' calls on_released even though the
-        # delivery state is actually MODIFIED
         if event.delivery.remote_state == Delivery.MODIFIED:
-            return self.on_modified(event)
-        self.released += 1
-        event.delivery.settle()
-
-    def on_modified(self, event):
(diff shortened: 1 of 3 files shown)


ganeshmurthy (approved):

tests/system_test.py
```diff
@@ -1107,15 +1107,10 @@ def on_accepted(self, event):
         event.delivery.settle()
 
     def on_released(self, event):
-        # for some reason Proton 'helpfully' calls on_released even though the
-        # delivery state is actually MODIFIED
         if event.delivery.remote_state == Delivery.MODIFIED:
-            return self.on_modified(event)
```
jiridanek, line 1113: This makes the handlers work the way they "should've been" done in Proton, imo. I thought that it was a reasonable improved thing to do. But ok, the Proton orthodoxy prevails.

ChugR: You know what, I agree with you. Leave it as is.

ChugR:
The handling in system_test makes a nice on_modified callback for dispatch self tests. Leave it as-is.

PR #1198, opened 2021-05-07
NO-JIRA Workaround apt-get issues and update clang to 12 in Travis CI

jiridanek:


ganeshmurthy:
With this change, the build that we marked as allowed to fail has passed. By that, I mean that there was a test failure but the build got past the past clang failure. I think we can go ahead and delete this build from the allowed failures and then push up this PR

PR #1200, opened 2021-05-07
NO-JIRA: Added info about the -j flag so tests can execute faster

ganeshmurthy:


README
```diff
@@ -134,6 +134,13 @@ option.  Example:
 
 ctest --timeout 1500 -VV
 
+The Dispatch test suite has a number of tests. To shorten execution time of the test suite, use the -j2 flag. Example:
+
+ctest -j2 -VV
+
+The integer following the -j specifies the maximum number of concurrent jobs that are used to execute the tests.
+Higher numbers could lead to unexpected test failures.
```
jiridanek, line 142: "... due to machine overload, which then causes missed timeouts in tests." Users should know that the tests themselves run concurrently just fine (or they should), it's only the hardware which is not keeping up.

"Reasonable value to use on a powerful developer laptop which provides good speed while does not suffer from spurious failures is `-j6`"

PR #1203, opened 2021-05-07
NO-JIRA: produce a zip file with router logs if GitHub Action fails

jiridanek:


jiridanek:
See the `Artifacts (4)` dropdown at https://github.com/jiridanek/qpid-dispatch/runs/2531478762 for an example of failed job with downloadable logs. The size in the list (1.12 GB) is a size after extracting the archive. The zip is about 50 MB only.

kgiusti:
This is really useful!
Does github limit the amount of space we can use for artifacts?  Will we run into issues if there's too many artifacts being stored by all our pull requests?

jiridanek:
https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts

> By default, GitHub stores build logs and artifacts for 90 days, and this retention period can be customized. For more information, see "Usage limits, billing, and administration". The retention period for a pull request restarts each time someone pushes a new commit to the pull request.

https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy

The limit is on the retention time, apparently. Looks like we should be ok, even if we produce gigabytes of logs (which dispatch currently does). Also, the logs are only stored if there is a failing test. Otherwise, the step is skipped.

PR #1210, opened 2021-05-11
DISPATCH-2113: Removed all relative path references in system tests

ganeshmurthy:


tests/system_tests_protocol_settings.py
```diff
@@ -83,8 +84,8 @@ def test_max_sessions(self):
         bc = BlockingConnection(self.router.addresses[0])
         bc.create_receiver("xxx")
         bc.close()
-
-        with open('../setUpClass/MaxSessions.log', 'r') as router_log:
+        logfile = os.path.join(self.router.outdir, self.router.logfile)
```
kgiusti, line 87: why not just use router.logfile_path like you did in other places?

kgiusti (approved):

PR #1219, opened 2021-05-17
DISPATCH-2068: Check endinaness before copying bytes into array

ganeshmurthy:


src/message.c
```diff
@@ -220,11 +229,17 @@ static void print_parsed_field(qd_parsed_field_t *parsed_field, char **begin, ch
            // 64-bit two’s-complement integer representing milliseconds since the unix epoch
            int timestamp_length = 8;
            pn_timestamp_t creation_timestamp = 0;
-
-           //qd_iterator_t* iter = qd_message_field_iterator(msg, field);
            qd_iterator_t *iter = qd_parse_raw(parsed_field);
-           while (!qd_iterator_end(iter) && timestamp_length > 0) {
-               timestamp_bytes[--timestamp_length] = qd_iterator_octet(iter);
+           if (lil_endian()) {
```
jiridanek, line 233: I'm trying to see if advice from https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html?m=1 is applicable. So far, I think yes. The code given there does not have to ask about the native byteorder, only about the byteorder used by the network protocol.

ganeshmurthy:
@jiridanek I have pushed a simple patch that does away with endianness check. Works in both Big and Little Endian systems. That was @kgiusti suggestion

PR #1221, opened 2021-05-17
DISPATCH-1986: Retry creating TestServer instance if port is already …

ganeshmurthy:
…in use

### DIFF
--- a/tests/http1_tests.py
+++ b/tests/http1_tests.py
@@ -30,6 +30,7 @@ except ImportError:
     from httplib import HTTPConnection, HTTPException
 
 from system_test import TestCase, TIMEOUT, Logger, Qdrouterd
+TEST_SERVER_ERROR = "TestServer failed to start due to port %s already in use issue"
 
 
 class RequestHandler(BaseHTTPRequestHandler):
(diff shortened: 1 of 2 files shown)


tests/http1_tests.py
```diff
@@ -272,13 +273,11 @@ def http1_ping(sport, cport):
         ]
     }
 
-    server = TestServer(server_port=sport,
-                        client_port=cport,
-                        tests=TEST)
+    server = wait_server_starting(sport, cport, TEST)
```
jiridanek, line 276: This problem would go away entirely if the test started the server first, before routers, with `TestServer(server_port=0, ...` and then queried the server for an actual port when starting the routers. But if the approach of the PR works reliably, then I see no reason to do yet another change to it.

ganeshmurthy: The tests bring up and bring down the TestServer (in some cases 2 test servers). The TestServer(s) seem to shut down successfully but the port is not released on time before the next test comes along and starts the next TestServer on the same port(s). This fix will wait for the few seconds until the port is released. Could take more than a few seconds for the port to be released but that is something I have not seen yet.

tests/http1_tests.py
```diff
@@ -367,6 +366,27 @@ def check_response(self, rsp):
         return body
 
 
+def wait_server_starting(server_port, client_port, tests, handler_cls=None):
```
kgiusti, line 369: This is a factory method.  I would suggest a more "pythonic" approach would make this a class method of the TestServer class and call it everywhere we need to instantiate a TestServer.

kgiusti (approved):

PR #1229, opened 2021-05-20
DISPATCH-2142: buffering changes for tcp adaptor

grs:


ChugR (approved):

src/adaptors/tcp_adaptor.c
```diff
@@ -216,59 +226,39 @@ void qdr_tcp_q2_unblocked_handler(const qd_alloc_safe_ptr_t context)
...
+    conn->read_buffer.size = 0;
+    conn->read_buffer.offset = 0;
+    conn->read_pending = false;
+    if (result > 0) {
         // account for any incoming bytes just read
         conn->last_in_time = tcp_adaptor->core->uptime_ticks;
-        conn->bytes_in      += count;
-        conn->bytes_unacked += count;
+        conn->bytes_in      += result;
+        conn->bytes_unacked += result;
```
kgiusti, line 254: Question:  if the received bytes are being flushed (buffers == NULL) they will never be sent to the destination - in this case should conn->bytes_unacked not be incremented?

kgiusti: To be fair I noticed the original code also increments conn->bytes_unacked in the same circumstances.

kgiusti (approved):

src/adaptors/tcp_adaptor.c
grs: I think if bytes_in is incremented, bytes_unacked should be also (even if there is never any possibility of them being acked), just for consistency.

Incrementing bytes_in makes sense in one way, as it indicates the amount of data available to read. However I don't feel strongly about that as when the connection is closed (which it will be if it is being flushed), then at present the record of it is deleted.

PR #1232, opened 2021-05-24
DISPATCH-1757: Added code to print out sender stats so we can see wha…

ganeshmurthy:
…t is going on

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -947,6 +947,20 @@ class AsyncTestReceiver(MessagingHandler):
     """
     Empty = Queue.Empty
 
+    class MyQueue(Queue.Queue):
+        def __init__(self, receiver):
+            self._async_receiver = receiver
+            super(AsyncTestReceiver.MyQueue, self).__init__()
+
+        def get(self, timeout=TIMEOUT):
+            self._async_receiver.num_queue_gets += 1
+            msg = super(AsyncTestReceiver.MyQueue, self).get(timeout=timeout)
+            return msg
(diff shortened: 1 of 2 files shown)


tests/system_test.py
```diff
@@ -1013,8 +1018,16 @@ def on_link_closing(self, event):
             rcv = event.container.create_receiver(event.connection,
                                                   **kwargs)
 
+    def queue_put(self, msg):
```
kgiusti, line 1021: You can avoid changing the API (and all those other files) by creating an internal class to the AsyncTestReceiver that inherits from Queue and overrides get() and put() to do what you want.

Something like:

class _MyQueue(Queue):                                                                                                                            
   def __init__(self, receiver):                                                                                                                  
       self._async_receiver = receiver                                                                                                            
       super(_MyQueue, self).__init__()                                                                                                           
                                                                                                                                                  
   def get(self, timeout=TIMEOUT):                                                                                                                
       self._async_receiver.num_queue_gets +=1                                                                                                    
       super(_MyQueue, self).get(timeout)                                                                                                         
                                                                                                                                                  
   def put(self, msg):                                                                                                                            
       self._async_receiver.num_queue_puts +=1                                                                                                    
       super(_MyQueue, self).put(msg)                                                                                                             
                                                                                                                                                  
                                                                                                                                                  
then in the async receiver's __init__:                                                                                                
    self.queue = _MyQueue(self)

kgiusti (approved):

PR #1235, opened 2021-05-26
DISPATCH-2136: Fix tsan race: convert message aborted flag to atomic

ChugR:
The message aborted flag is read and written by I/O and core threads.
Conversion to a sys_atomic variable ensures consistent state.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -1004,6 +1004,7 @@ qd_message_t *qd_message()
     ZERO(msg->content);
     msg->content->lock = sys_mutex();
     sys_atomic_init(&msg->content->ref_count, 1);
+    sys_atomic_init(&msg->content->aborted, 0);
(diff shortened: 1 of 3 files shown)


kgiusti (approved):

src/message.c
```diff
@@ -45,6 +45,10 @@
 #define LOCK   sys_mutex_lock
 #define UNLOCK sys_mutex_unlock
 
+// Implement bool flags with atomic variables
+#define SET_FLAG(flag)  sys_atomic_set(flag, 1)
+#define  IS_FLAG(flag) (sys_atomic_get(flag) == 1)
```
ganeshmurthy, line 50: IMO, I think the #define IS_FLAG(flag) can be called #define IS_FLAG_SET(flag)

ganeshmurthy (approved):

ganeshmurthy:
Good to go !!

PR #1239, opened 2021-06-01
DISPATCH-1956: Fix TSan-reported lock-inversion

mgoulish:


src/log.c
```diff
@@ -379,7 +398,14 @@ qd_log_source_t *qd_log_source(const char *module)
 {
     sys_mutex_lock(log_source_lock);
     qd_log_source_t* src = qd_log_source_lh(module);
+    // Note to Mick: Mick! Please don't access log_source fields
+    // outside of the log_source lock. That's why there's a lock, you know.
+    bool add_me = src->add_me;
```
ganeshmurthy, line 403: Can we please call this flag add_log_stats or something instead of add_me ?

mgoulish: How about    add_me_or_else     ?

src/log.c
```diff
@@ -518,6 +544,7 @@ void qd_log_initialize(void)
     default_log_source->includeTimestamp = true;
     default_log_source->includeSource = 0;
     default_log_source->sink = log_sink_lh(SINK_STDERR);
+    default_log_source->add_log_stats = false;
```
ganeshmurthy, line 547: This line might not be necessary because the call to qd_log_source() already sets the add_log_stats flag to false.

mgoulish: Leave it. It's a nanosecond.

src/log.c
```diff
@@ -648,8 +685,12 @@ qd_error_t qd_log_entity(qd_entity_t *entity)
             src->includeSource = include_source;
         }
 
+        bool add_log_stats = src->add_log_stats;
+        src->add_log_stats = false;
         sys_mutex_unlock(log_source_lock);
-
+        if(add_log_stats) {
+            qd_entity_cache_add(QD_LOG_STATS_TYPE, src);
```
ganeshmurthy, line 692: I am unsure why we need to add the QD_LOG_STATS_TYPE here ?
I commented this line and the logStats was still created because qd_log_source() is creating it.

src/log.c
mgoulish: And anyway -- this function initializes the log. To initialize the log, that flag should be set to false.

ganeshmurthy: If you look at https://github.com/apache/qpid-dispatch/blob/main/src/log.c#L368, the ZERO(log_source); is called. This sets all bits to 0 on the log_source which then automatically sets the add_log_stats to false, so no need to set the add_log_stats explicitly to false in qd_log_source_defaults() which is called right after the call to ZERO(log_source); 
But yes, if ZERO(log_source);  was not called in the future, then qd_log_source_defaults() would take care of defaulting the add_log_stats flag.

ganeshmurthy: I wonder if the call to ZERO should be moved inside the qd_log_source_defaults() function. If we do that, the qd_log_source_defaults() can be called from the place where ZERO is currently being called from. And then, there would be no need to set the add_log_stats to false.

kgiusti (changes requested):
As part of this patch, can you comment out the suppression for DISPATCH-1956 in the tsan.supp file?

https://github.com/apache/qpid-dispatch/blob/main/tests/tsan.supp#L12

mgoulish:
Closing this PR in favor of a new one, coming soon.

PR #1240, opened 2021-06-02
NO-JIRA: Uncomment http tests in CI

ganeshmurthy:


.github/workflows/build.yaml
```diff
@@ -182,7 +182,7 @@ jobs:
       DispatchBuildDir: ${{github.workspace}}/qpid-dispatch/build
       InstallPrefix: ${{github.workspace}}/install
       # TODO(DISPATCH-2078) re-enable system_tests_authz_service_plugin when the GHA failure is understood and fixed
-      DispatchCTestExtraArgs: "-E 'system_tests_authz_service_plugin|system_tests_http1_adaptor|system_tests_http2|system_tests_grpc'"
+      DispatchCTestExtraArgs: "-E 'system_tests_authz_service_plugin'"
```
jiridanek, line 185: This is duplicated in two places in the same file. Once for ubuntu, and second time for Fedora. Sorry about that, I knew it's going to cause trouble...

jiridanek: The other occurrence is line 285.

ganeshmurthy: thanks, I missed that. In any case, I just want to uncomment the http tests in this PR. Can we remove the duplicated system_tests_authz_service_plugin in a separate commit ? thx

kgiusti (approved):

PR #1242, opened 2021-06-03
DISPATCH-1896 - Propagate the GOAWAY frame received from the http2 se…

ganeshmurthy:
…rver all the way back to the client

### DIFF
--- a/README
+++ b/README
@@ -88,7 +88,6 @@ Test Suite Code Coverage (GNU tools only)
 
 Use coverage analysis to ensure that all code paths are exercised by
 the test suite. To run the tests and perform code coverage analysis:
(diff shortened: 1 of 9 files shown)


README
```diff
@@ -90,6 +88,7 @@ Test Suite Code Coverage (GNU tools only)
 
 Use coverage analysis to ensure that all code paths are exercised by
 the test suite. To run the tests and perform code coverage analysis:
+the test suite. To run the tests and perform code coverage analysis:
```
kgiusti, line 91: duplicate?

kgiusti (approved):
Other than the duplication - it looks fine to me.

PR #1243, opened 2021-06-04
DISPATCH-1956: reduce locking around log src lookup

mgoulish:
I noticed that almost all calls to   qd_log_source_lh()  were very simple:  take the lock, call the function, release the lock.

But that means that we could move the lock take-and-release inside of that function, with no change in behavior.

And then that means I don't have to ask the caller to add my new log_source (if it is new) to the list -- I can do that myself from within the function and simplify everything -- but do it after I drop the log_source_lock, so we don't get the lock inversion.

SO I did that, and removed the _lh from the end of the function name -- but that made it collide with the name of another function. So I changed it to qd_log_source_find_or_make().
I.e. "if you don't find one of this name, then make it."

The only slight disturbance was the function qd_log_entity() which took the lock and used it to cover this function and a couple other calls to a different  _lh() function.    SO for that one I just inserted new take-and-releases of the lock, right around the other function calls.   Now there is only one line of code between each take-and-release, which seems much nicer.

Also qd_log_source_reset() bugged me because it called my new function but then immediately had to take the lock just so it could set the log_source to default values. SO I just added a flag to my new fn so that  qd_log_source_reset()  could tell me to do that for him right after I retrieve it.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -358,18 +358,32 @@ static void qd_log_source_defaults(qd_log_source_t *log_source) {
     memset ( log_source->severity_histogram, 0, sizeof(uint64_t) * (N_LEVEL_INDICES) );
 }
 
-/// Caller must hold the log_source_lock
-static qd_log_source_t *qd_log_source_lh(const char *module)
+/// Find a preexisting source of this name, or make a new one.
+/// Optionally, set default values even if the source was prexisting.
+static qd_log_source_t *qd_log_source_find_or_make(const char *module, bool set_defaults)
 {
+    bool add = false;
+    sys_mutex_lock(log_source_lock);
     qd_log_source_t *log_source = lookup_log_source_lh(module);
-    if (!log_source)
-    {
+    if (log_source) {
+        if ( set_defaults )
+            qd_log_source_defaults(log_source);
+    } else {
         log_source = NEW(qd_log_source_t);
         ZERO(log_source);
         log_source->module = (char*) malloc(strlen(module) + 1);
         strcpy(log_source->module, module);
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -592,24 +601,22 @@ qd_error_t qd_log_entity(qd_entity_t *entity)
...
                 error_in_output = true;
-                sys_mutex_unlock(log_source_lock);
                 break;
             }
 
             // DEFAULT source may already have a sink, so free the old sink first
             if (src->sink) {
+                sys_mutex_lock(log_source_lock);
                 log_sink_free_lh(src->sink);
+                sys_mutex_unlock(log_source_lock);
```
kgiusti, line 619: This could be a problem: src->sink has been freed.  Once the lock is dropped some other thread may try to access that stale sink pointer.

src/log.c
```diff
@@ -592,24 +601,22 @@ qd_error_t qd_log_entity(qd_entity_t *entity)
...
-        //
-        sys_mutex_lock(log_source_lock);
-
-        qd_log_source_t *src = qd_log_source_lh(module); /* The original(already existing) log source */
+        qd_log_source_t *src = qd_log_source_find_or_make(module, false); /* The original (already existing) log source */
 
         if (has_output_file) {
+            sys_mutex_lock(log_source_lock);
             log_sink_t* sink = log_sink_lh(outputFile);
+            sys_mutex_unlock(log_source_lock);
```
kgiusti, line 609: What prevents another thread from calling this function and freeing the same sink (line 618) once the first thread drops the lock?

kgiusti (changes requested):

PR #375, opened 2018-09-13
DISPATCH 1096 - store priority in links

mgoulish:
TED -- I was not able to reproduce the problem you saw, after 1000 iterations. So -- take a look at this code and if this is what you had in mind, pull away. 

I am going to do some massive testing with python tests -- comparing this code to pre-priority code.

PR #1248, opened 2021-06-08
DISPATCH-1940: Added code to handle http2 flow control window updates

ganeshmurthy:


src/adaptors/http2/http2_adaptor.c
```diff
@@ -449,6 +469,17 @@ static qdr_http2_stream_data_t *create_http2_stream_data(qdr_http2_session_data_
 }
 
 
+/**
+ * This callback function  is invoked when the nghttp2 library tells the application about the error code, and error message.
+ */
+static int on_error_callback(nghttp2_session *session, int lib_error_code, const char *msg, size_t len, void *user_data)
+{
+	qdr_http2_connection_t *conn = (qdr_http2_connection_t *)user_data;
+	qd_log(http2_adaptor->protocol_log_source, QD_LOG_TRACE, "[C%"PRIu64"] Error generated in the on_error_callback, lib_error_code=%i, error_msg=%s", conn->conn_id, lib_error_code, msg);
```
kgiusti, line 478: If this results in a user-visible failure, should the log level be warn/error/etc?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -576,25 +609,27 @@ static int snd_data_callback(nghttp2_session *session,
         memcpy(qd_http2_buffer_cursor(http2_buff), framehd, HTTP2_DATA_FRAME_HEADER_LENGTH);
         qd_http2_buffer_insert(http2_buff, HTTP2_DATA_FRAME_HEADER_LENGTH);
         pn_raw_buffer_t pn_raw_buffs[stream_data->qd_buffers_to_send];
-        qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, 0, stream_data->qd_buffers_to_send);
+        qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, stream_data->curr_stream_data_qd_buff_offset, stream_data->qd_buffers_to_send);
```
kgiusti, line 612: Shouldn't you use the return value here to determine how many buffers have actually been moved?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -576,25 +609,27 @@ static int snd_data_callback(nghttp2_session *session,
         memcpy(qd_http2_buffer_cursor(http2_buff), framehd, HTTP2_DATA_FRAME_HEADER_LENGTH);
         qd_http2_buffer_insert(http2_buff, HTTP2_DATA_FRAME_HEADER_LENGTH);
         pn_raw_buffer_t pn_raw_buffs[stream_data->qd_buffers_to_send];
```
kgiusti, line 611: Ah - a VLA?  Thought you were avoiding these...

src/adaptors/http2/http2_adaptor.c
```diff
@@ -576,25 +609,27 @@ static int snd_data_callback(nghttp2_session *session,
...
+        qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, stream_data->curr_stream_data_qd_buff_offset, stream_data->qd_buffers_to_send);
 
         int idx = 0;
+        size_t bytes_to_send = length;
         while (idx < stream_data->qd_buffers_to_send) {
             if (pn_raw_buffs[idx].size > 0) {
-                //int bytes_remaining = length - bytes_sent;
-                //if (bytes_remaining > pn_raw_buffs[idx].size) {
+            	if (bytes_to_send < pn_raw_buffs[idx].size) {
+            		memcpy(qd_http2_buffer_cursor(http2_buff), pn_raw_buffs[idx].bytes + stream_data->curr_stream_data_offset, bytes_to_send);
```
kgiusti, line 619: Q: can this loop overrun the space available in http2_buff?   Where is the remaining capacity of http2_buff tracked?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -359,11 +364,25 @@ static void free_http2_stream_data(qdr_http2_stream_data_t *stream_data, bool on
...
+    	if (stream_data->curr_stream_data)
+    		qd_message_stream_data_release(stream_data->curr_stream_data);
+    	if (stream_data->next_stream_data)
+    		qd_message_stream_data_release(stream_data->next_stream_data);
         qdr_delivery_decref(http2_adaptor->core, stream_data->in_dlv, "HTTP2 adaptor in_dlv - free_http2_stream_data");
     }
 
     if (stream_data->out_dlv && !stream_data->out_dlv_decrefed) {
+    	if (stream_data->curr_stream_data)
+    		qd_message_stream_data_release(stream_data->curr_stream_data);
```
kgiusti, line 383: could this have already been freed on line #375?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -359,11 +364,25 @@ static void free_http2_stream_data(qdr_http2_stream_data_t *stream_data, bool on
...
+    	if (stream_data->next_stream_data)
+    		qd_message_stream_data_release(stream_data->next_stream_data);
```
kgiusti, line 385: same Q here (line 377)?

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: Good point, I am going to set the stream_data->curr_stream_data and stream_data->next_stream_data once we call qd_message_stream_data_release. That way the succeeding lines cannot double free

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: Same answer as for the previous question

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: Changed log level to ERROR

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: I will an an assert to the code --

int written = qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, stream_data->curr_stream_data_qd_buff_offset, stream_data->qd_buffers_to_send);

assert (written == stream_data->qd_buffers_to_send);

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: We pre-determine how much we can stuff into a http2 buffer on line 1247 below. nghttp2 allows you to send a max for 16k per callback. So we determine how many qd buffers can fit into that 16k. We use buff offsets and offsets within each buffer to track where we are in the body data object.

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: This VLA is in the fast path. It seems to me that the VLA is the fastest approach in this situation.  Thoughts?

In this case, stream_data->qd_buffers_to_send does not exceed 32.
32 is the number of qd_buffers that will fit in one 16k http2 buffer.

PR #1249, opened 2021-06-08
DISPATCH-2105: Removed variable length arrays wherever possible

ganeshmurthy:


ChugR (approved):

kgiusti:
I think the changes are fine, but the travis test run shows up a particularly worrisome error:

https://travis-ci.com/github/apache/qpid-dispatch/jobs/512555656#L10222

I don't think it's related to this change however - seems like the refcount on the address is in error....

ganeshmurthy:
> I don't think it's related to this change however - seems like the refcount on the address is in error....
That does seem unrelated. That use after free comes from the http2 adaptor test. I have raised a JIRA for it - https://issues.apache.org/jira/browse/DISPATCH-2168

kgiusti (approved):

PR #1252, opened 2021-06-09
DISPATCH-2097: Add management status for hiding connections on console gui

ChugR:
The first use of this feature is to signal to the console not
to display TCP adaptor egress_dispatcher connections.


ChugR:
A better approach is to insert a key/value pair into the connection properties map. Something like

  qd.adaptor: tcp

PR #1253, opened 2021-06-09
DISPATCH-1977: Introduced atomics raw_closed_read and raw_closed_writ…

ganeshmurthy:
…e in http2 adaptor. Also made raw_closed_write and raw_closed_read as atomics in tcp_adaptor"

### DIFF
--- a/include/qpid/dispatch/atomic.h
+++ b/include/qpid/dispatch/atomic.h
@@ -205,7 +205,9 @@ static inline void sys_atomic_destroy(sys_atomic_t *ref)
 
 #endif
 
-#define    SET_ATOMIC_FLAG(flag)  sys_atomic_set(flag, 1)
+#define    SET_ATOMIC_FLAG(flag)   sys_atomic_set(flag, 1)
+#define    CLEAR_ATOMIC_FLAG(flag) sys_atomic_set(flag, 0)
+
 #define IS_ATOMIC_FLAG_SET(flag) (sys_atomic_get(flag) == 1)
 
 /** Atomic increase: NOTE returns value *before* increase, like i++ */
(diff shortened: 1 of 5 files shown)


include/qpid/dispatch/atomic.h
```diff
@@ -205,6 +205,9 @@ static inline void sys_atomic_destroy(sys_atomic_t *ref)
 
 #endif
 
+#define    SET_ATOMIC_FLAG(flag)  sys_atomic_set(flag, 1)
```
kgiusti, line 208: Why not a CLEAR_ATOMIC_FLAG?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -2426,6 +2436,8 @@ static void handle_connection_event(pn_event_t *e, qd_server_t *qd_server, void
             send_settings_frame(conn);
             qd_log(log, QD_LOG_INFO, "[C%"PRIu64"] Accepted Ingress ((PN_RAW_CONNECTION_CONNECTED)) from %s", conn->conn_id, conn->remote_address);
         } else {
+            sys_atomic_init(&conn->raw_closed_read, 0);
```
kgiusti, line 2439: Why re-init?  Won't this destroy the existing state?

include/qpid/dispatch/atomic.h
ganeshmurthy: Introduced a CLEAR_ATOMIC_FLAG macro and the http adaptor now uses it to clear the flags once a new connection to the server is established.

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: Re-init was wrong. Now uses CLEAR_ATOMIC_FLAG

kgiusti (approved):

PR #1255, opened 2021-06-10
DISPATCH-1797: Added a retry to qdmanage to make sure we give enough …

ganeshmurthy:
…time for addresses to be added to the router address table

### DIFF
--- a/tests/system_tests_qdmanage.py
+++ b/tests/system_tests_qdmanage.py
@@ -542,17 +542,23 @@ class QdmanageTest(TestCase):
         # Try fetching all 10,000 addresses
         # This qdmanage query command would fail without the fix
         # for DISPATCH-974
+        i = 0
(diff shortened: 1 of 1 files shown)


tests/system_tests_qdmanage.py
```diff
@@ -542,17 +542,23 @@ def test_yy_query_many_links(self):
         # Try fetching all 10,000 addresses
         # This qdmanage query command would fail without the fix
         # for DISPATCH-974
+        i = 0
         query_command = 'QUERY --type=org.apache.qpid.dispatch.router.address'
-        outs = json.loads(self.run_qdmanage(query_command))
-
-        sender_addresses = 0
-        receiver_addresses = 0
+        while i < 3:
```
jiridanek, line 547: `for _ in range(3):` would make this be two lines shorter ;)

kgiusti (approved):

PR #1260, opened 2021-06-16
DISPATCH-1956: rewrite of log.c to avoid most locking

mgoulish:
Please see large comment at start of file that describes approach.

I will remove that comment -- as well as all the "1956" comments -- before this code goes to main.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -79,6 +79,7 @@ typedef struct log_sink_t {
     char *name;
     bool syslog;
     FILE *file;
+    sys_mutex_t *sink_lock;
     DEQ_LINKS(struct log_sink_t);
 } log_sink_t;
 
@@ -110,19 +111,14 @@ static const char* SINK_STDERR = "stderr";
 static const char* SINK_SYSLOG = "syslog";
 static const char* SOURCE_DEFAULT = "DEFAULT";
 
-static log_sink_t* find_log_sink_lh(const char* name) {
-    log_sink_t* sink = DEQ_HEAD(sink_list);
-    DEQ_FIND(sink, strcmp(sink->name, name) == 0);
-    return sink;
-}
-
-// Must hold the log_source_lock
-static void log_sink_free_lh(log_sink_t* sink) {
+static void log_sink_free(log_sink_t* sink) {
     if (!sink) return;
     assert(sink->ref_count);
 
     if (sys_atomic_dec(&sink->ref_count) == 1) {
+        sys_mutex_lock(sink->sink_lock);
         DEQ_REMOVE(sink_list, sink);
+        sys_mutex_unlock(sink->sink_lock);
         free(sink->name);
         if (sink->file && sink->file != stderr)
             fclose(sink->file);
@@ -132,8 +128,10 @@ static void log_sink_free_lh(log_sink_t* sink) {
     }
 }
 
-static log_sink_t* log_sink_lh(const char* name) {
-    log_sink_t* sink = find_log_sink_lh(name);
+static log_sink_t* log_sink(const char* name) {
+    // Find the named sink if it exists.
+    log_sink_t* sink = DEQ_HEAD(sink_list);
+    DEQ_FIND(sink, strcmp(sink->name, name) == 0);
     if (sink) {
         sys_atomic_inc(&sink->ref_count);
     }
@@ -170,8 +168,8 @@ static log_sink_t* log_sink_lh(const char* name) {
         sink->name = strdup(name);
         sink->syslog = syslog;
         sink->file = file;
+        sink->sink_lock = sys_mutex();
         DEQ_INSERT_TAIL(sink_list, sink);
-
     }
     return sink;
 }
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -1,3 +1,49 @@
...
+      time, and it won't let anybody call that function again.
+
+    * Each log-source gets its own lock, which is used only to protect
+      operations on that source's 'sink'.  We still have to do this
+      because otherwise, management could call in at a random moment
+      and replace a sink out from under us, causing untold havoc,
+      destruction, and loss of life.
+
+    * The old global lock is only used in one place now, so I've
+      renamed it for that use. It's a lock for a global long entry
```
ChugR, line 28: /long/log/

src/log.c
```diff
@@ -1,3 +1,49 @@
...
+      list.
+
+    * I think the net effect of all this is to greatly reduce the
+      scope of locking.
+
+    * This code eliminates the deadlock that was happening befoe, with
```
ChugR, line 34: /befoe/before/

ChugR: sorry. these were pre-deleted comments.

ChugR:
The two fedora asan builds melted down and hardly any tests passed. This PR branch runs all tests fine for me.

src/log.c
```diff
@@ -187,16 +234,19 @@ struct qd_log_source_t {
...
     char *module;
     int mask;
-    int includeTimestamp;       /* boolean or -1 means not set */
-    int includeSource;          /* boolean or -1 means not set */
+    int includeTimestamp;       // boolean or -1 means not set
+    int includeSource;          // boolean or -1 means not set
     bool syslog;
     log_sink_t *sink;
     uint64_t severity_histogram[N_LEVEL_INDICES];
+    sys_mutex_t * sink_lock;    // 1956: This lock protects operations on this source's sink.
```
kgiusti, line 242: formatting: no space between * and sink_lock

src/log.c
```diff
@@ -354,41 +403,42 @@ static void qd_log_source_defaults(qd_log_source_t *log_source) {
...
+    sys_mutex_lock(log_source->sink_lock);
+    log_source->sink = 0;
+    sys_mutex_unlock(log_source->sink_lock);
 }
 
-/// Caller must hold the log_source_lock
-static qd_log_source_t *qd_log_source_lh(const char *module)
+static bool log_initialized = false;
+
+static qd_log_source_t *
```
kgiusti, line 415: formatting: return type and function signature on the same line please.

src/log.c
```diff
@@ -354,41 +403,42 @@ static void qd_log_source_defaults(qd_log_source_t *log_source) {
...
-        qd_log_source_defaults(log_source);
-        DEQ_INSERT_TAIL(source_list, log_source);
-        qd_entity_cache_add(QD_LOG_STATS_TYPE, log_source);
-    }
+    if(log_initialized)
+      return NULL;
+
+    qd_log_source_t *log_source = NEW(qd_log_source_t);
+    ZERO(log_source);
+    log_source->module = (char*) malloc(strlen(module) + 1);
```
kgiusti, line 423: While we're at it, we should move away from the malloc+strcpy pattern to simply using qd_strdup()

src/log.c
```diff
@@ -499,8 +552,19 @@ PyObject *qd_log_recent_py(long limit) {
...
+
+// 1956: At initialization time, we create one log-source for each module.
+//       If a new module is written, include its name in the list.
+//       This fn can only be called once.
+
+
 void qd_log_initialize(void)
 {
+    if(log_initialized) {
+      // TODO -- log something here?
```
kgiusti, line 564: Just assert(!log_initialized) here - it's a coding bug to call it twice.

src/log.c
```diff
@@ -562,11 +658,14 @@ qd_error_t qd_log_entity(qd_entity_t *entity)
...
-        // also holds the log_source_lock when calling write_log().
-        //
+        qd_log_source_t *src = qd_log_source(module); /* The original(already existing) log source */
+
+        // 1956: Since we are messing with this source's sink, here we
+        //       can use the local sink-lock.
+        //       Previously, we needed to obtain these attributes outside of the lock, because
+        //       of a deadlock issue.
+
+        sys_mutex_lock(src->sink_lock);
```
kgiusti, line 668: I don't think this is correct: we shouldn't be protecting the sink because the sink is not changing.  The source is being modified - and if any sink attributes need to be modified we simply create a new one and drop the reference to the old sink (which can still be in use by other sources).

kgiusti: For this I think maybe a per-source read/write lock would be a better choice:  take the read lock while logging.  Take the write lock during this call.

src/log.c
```diff
@@ -429,22 +479,25 @@ void qd_vlog_impl(qd_log_source_t *source, qd_log_level_t level, bool check_leve
...
+    // Obtain the log_entries_lock. We need to do this, if not, the qd_log_entity() function
     // could free the log_source->sink from underneath you and replace it with a new sink.
     // Once we obtain this lock, we only release the lock once the log line is written to the sink.
     //
-    sys_mutex_lock(log_source_lock);
     entry->module = source->module ? strdup(source->module) : 0;
     entry->level  = level;
     entry->file   = file ? strdup(file) : 0;
     entry->line   = line;
     gettimeofday(&entry->time, NULL);
```
kgiusti, line 490: We may need to move the gettimeofday into write_log so we're holding the lock while the timestamp is taken to reduce likelyhood of re-ordering log events.

PR #1261, opened 2021-06-16
Thread debug cr4 - DO NOT MERGE

ChugR:
This commit provides a framework for analyzing mutex locking patterns. Each mutex gets a name when it is created. Then when the mutex is locked and unlocked a per-thread stack of locked mutexes is printed to stderr.

- The first commit by kgiusti provides the foundation instrumentation to show the lock nesting. 
- The second commit by chugr adds timestamps and prints lock acquisition and hold times in microseconds.

This package is pertty verbose. A full run of self tests on my laptop consumes 34Gbytes of disk storage to hold the lock log output. That of course makes useful data a little hard to find. With a little patience one can find instances of locks being held for > 1 second. (Check edge_router test, EA1.out file) Analysis tools to crunch the data are forthcoming. 

Note that the second commit also prints timestamped data to stderr out-of-order! There is a tradeoff for the ordering problem which is corrected by a simple sort vs. adding an extra mutex lock for every line printed. The commit chooses the out-of-order solution for speed and simplicity.

### DIFF
--- a/include/qpid/dispatch/server.h
+++ b/include/qpid/dispatch/server.h
@@ -569,6 +569,35 @@ void qd_connection_invoke_deferred(qd_connection_t *conn, qd_deferred_t call, vo
 
 
 /**
+ * Schedule a call to be invoked on a thread that has ownership of this connection
+ * when it will be safe for the callback to perform operations related to this connection.
+ * A qd_deferred_call_t object has been allocated before hand to avoid taking
+ * the ENTITY_CACHE lock.
+ *
+ * @param conn Connection object
+ * @param call The function to be invoked on the connection's thread
+ * @param context The context to be passed back in the callback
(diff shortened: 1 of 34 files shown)


ChugR:
The analysis is not done yet. Improvements planned for immediate future:

- per-lock statistic CSV tables similar to per-thread tables
- automatic harvesting of core thread id and start/stop timestamps to exclude enormous lock times during router startup and shutdown
- log file emitter so that lock events can be integrated with normal router logs and viewed with Scraper. This will let a researcher know what router events are associated with long lock acquisition and hold times.

ChugR:
See result at
https://issues.apache.org/jira/secure/attachment/13027168/DISPATCH-2183_85f805_topology-disposition-routerA.txt

PR #1265, opened 2021-06-17
NO-JIRA: parallel priority test fix - wait for senders to complete first

kgiusti:


ganeshmurthy:
Will this fix take care of this JIRA - https://issues.apache.org/jira/browse/DISPATCH-2116  ?

kgiusti:
> Will this fix take care of this JIRA - https://issues.apache.org/jira/browse/DISPATCH-2116 ?

This fix doesn't apparently fix anything :) 
I'm going to try instrumenting debug code - see what gives.

ganeshmurthy (approved):

kgiusti:
> Will this fix take care of this JIRA - https://issues.apache.org/jira/browse/DISPATCH-2116 ?

I wouldn't think so - 2116 indicates failure to see any framing activity during the idle-timeout period.  That means - for whatever reason - the test client just "hung" hard: no message output.

This fix simply reduces the size of the message.  If the client has gone silent.... I'd imagine that's something else.   Could be the GHA simply did not schedule the client(s) to run for a few seconds (a stretch, but not impossible).

PR #1266, opened 2021-06-17
NO-JIRA: update GHA to use same timeouts as travis

kgiusti:


jiridanek (approved):
Makes sense; I somehow did not think of applying the Travis config to GHA as well. Maybe because it was not needed, until the slower TSan Fedora jobse were added.

There is an interesting CTest test timeout at https://github.com/apache/qpid-dispatch/pull/1266/checks?check_run_id=2852874425. Logs from that job were archived correctly (I noticed some log archiving failures on GitHub's side recently).

kgiusti:
> Makes sense; I somehow did not think of applying the Travis config to GHA as well. Maybe because it was not needed, until the slower TSan Fedora jobse were added.
> 
> There is an interesting CTest test timeout at https://github.com/apache/qpid-dispatch/pull/1266/checks?check_run_id=2852874425. Logs from that job were archived correctly (I noticed some log archiving failures on GitHub's side recently).

Yeah I noticed that - it appears as if routes never completed propagation, even given the longer timeouts.  That's a problem - the amount of time given by the test should be waaay more than necessary to get the routes propagated.

I'm going ahead with the patch but I'll ping the tcp guys to take a look at these logs.

PR #1267, opened 2021-06-20
DISPATCH-2178: Do not build separate libqpid-dispatch.so

jiridanek:
The change would've been nice and easy in CMake 3.12+. Without support for using OBJECT library in `target_link_libraries`, it is necessary to 1) use a STATIC library and then handle the `-Wl,--whole-archive` flag (so that symbols used only from Python are not eliminated by linker) or 2) propagate the dependent libraries through the build some other way. I choose option 2) here.

@astitcher Is this what you had in mind, or are there some improvements that I've missed?

This works on Centos 7, Ubuntu and Fedora. I wonder what Travis will say.


tests/c_unittests/CMakeLists.txt
```diff
@@ -22,8 +22,8 @@ file(GLOB unittest_SOURCES
 )
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_STANDARD_FLAGS} ${SANITIZE_FLAGS}")
-add_executable(c_unittests ${unittest_SOURCES})
-target_link_libraries(c_unittests pthread qpid-dispatch)
+add_executable(c_unittests ${unittest_SOURCES}  $<TARGET_OBJECTS:qpid-dispatch>)
```
jiridanek, line 25:
```suggestion
add_executable(c_unittests ${unittest_SOURCES} $<TARGET_OBJECTS:qpid-dispatch>)
```

jiridanek:
MacOS seems to have failed on the empty STATIC library https://travis-ci.com/github/apache/qpid-dispatch/jobs/516966091

```
ar: no archive members specified
usage:  ar -d [-TLsv] archive file ...
	ar -m [-TLsv] archive file ...
	ar -m [-abiTLsv] position archive file ...
	ar -p [-TLsv] archive [file ...]
	ar -q [-cTLsv] archive file ...
	ar -r [-cuTLsv] archive file ...
	ar -r [-abciuTLsv] position archive file ...
	ar -t [-TLsv] archive [file ...]
	ar -x [-ouTLsv] archive [file ...]
make[2]: *** [src/libqpid-dispatch-libraries.a] Error 1
```

Other than that, it looks good (so far).

As a quick fix, I can add some dummy symbol into the library which will be then eliminated by linker during linking anyways, so the effect will be the same, and it should work even on macOS.

src/server.c
```diff
@@ -1738,7 +1738,7 @@ void qd_connector_decref(qd_connector_t* connector)
     }
 }
 
-__attribute__((noinline)) // permit replacement by dummy implementation in unit_tests
+__attribute__((weak)) // permit replacement by dummy implementation in unit_tests
```
astitcher, line 1741: [Not a comment on this PR really!]
This whole replacing a function by redefining in the unit tests without an injection API makes me queazy! I'm not entirely sure why - it just seems error-prone.

tests/CMakeLists.txt
```diff
@@ -40,8 +40,8 @@ set(unit_test_SOURCES
     thread_test.c
     )
 
-add_executable(unit_tests ${unit_test_SOURCES})
-target_link_libraries(unit_tests  qpid-dispatch)
+add_executable(unit_tests ${unit_test_SOURCES} $<TARGET_OBJECTS:qpid-dispatch>)
+target_link_libraries(unit_tests qpid-dispatch-libraries)
```
astitcher, line 44: In this and the subsequent similar changes I don't understand why you need both to add the object files and to add a static library. Surely you are just effectively adding the same object twice like this? I think you only need to do the add_executable() change and remove the target_link_libraries() unless there is another library to add.

astitcher (commented):
@jiridanek This is exactly what I had in mind. And I agree this would be more comprehensible in a more recent version of CMake - I had the same issue when decreasing the compile time of proton using a cmake object library.

tests/CMakeLists.txt
jiridanek: The static library is empty, it is only CMake target that carries dependencies of qpid-dispatch. Like INTERFACE library, but that is not supported in CMake 3.8.12.2. I have to propagate the PUBLIC/INTERFACE dependencies of the object files in `$<TARGET_OBJECTS:qpid-dispatch>` somehow.

If I had CMake 3.12, I could do `target_link_libraries(unit_tests qpid-dispatch)` and be done with it. For older CMake, I have to add the object files into `add_executable`, but that does nothing about link dependencies. So I'd have to have e.g. a variable.... I thought that a `-libraries` static library containing no symbols of its own is cleaner. (The library will actually have to contain something due to the macOS error I mentioned above.)

src/server.c
jiridanek: I'd love to do something about it, but there isn't really an existing universal solution (in the form of a library) that I'd be happy with. And I never spent the time.

Doing linking substitution like this actually is not so bad. It is even recommended (in it's original version, with a shared library) in http://blog.wingman-sw.com/linker-substitution-in-c-limitations-and-workarounds. When switching from shared library to a static one, I could've either kept building the so in addition, just for use in tests (that is a good idea, AFAIK, for the future), or change the tests to use the other technique mentioned in the article (the pointer substitution one). Or use weak symbols, which produces the smallest diff.

I'm actually happy that we can get rid of `noinline`, that was bugging me. Clang by default wants to inline one of the functions on -O2 and we have to prevent that because of our tests. I feel better about having weak symbols around, but that might be just me. Maybe there is some harm in them that I am just not seeing. (Besides the obvious: you might redefine it when you don't mean to.) Noinline is btw necessary with the runtime injection APIs. Or compile for tests extra and turn off optimizations altogether.

Regarding the function pointer substitution technique from the blog. It is the first time I saw it and it is actually a very good solution, also it uses standard C without any platform magic (either linker trickery or ELF file manipulation). It looks really simple to understand, but it's actually quite clever. I wouldn't 've come up with it myself.

src/empty.c
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+typedef int qd_make_this_translation_unit_not_be_empty;
```
jiridanek, line 20: This translation unit is deliberately left blank ;P

src/dispatch.c
```diff
@@ -122,9 +122,9 @@ STATIC_ASSERT(sizeof(long) >= sizeof(void*), pointer_is_bigger_than_long);
 
 qd_error_t qd_dispatch_load_config(qd_dispatch_t *qd, const char *config_path)
 {
-    qd->dl_handle = dlopen(QPID_DISPATCH_LIB, RTLD_LAZY | RTLD_NOLOAD);
+    qd->dl_handle = dlopen(NULL, RTLD_LAZY | RTLD_NOLOAD);
```
kgiusti, line 125: This dlopen is done for backward compatibility, right?  Can you drop a comment here explaining that?

src/empty.c
kgiusti: Can you add that comment about macOS support here as well?

kgiusti (approved):
Just needs a couple of comments so we don't break things in the future.

src/dispatch.c
jiridanek: I was thinking whether I could move the `dlopen`  to the python side, in `configure_dispatch`, but I was not exactly sure I can make the things work exactly the way they are now.

PR #1268, opened 2021-06-21
DISPATCH-2179: Set SSLDomain.ANONYMOUS_PEER wherever necessary. Also …

ganeshmurthy:
…fixed python code to set correct scheme

### DIFF
--- a/python/qpid_dispatch_internal/tools/command.py
+++ b/python/qpid_dispatch_internal/tools/command.py
@@ -261,9 +261,8 @@ def opts_url(opts):
     url = Url(opts.bus)
 
     # If the options indicate SSL, make sure we use the amqps scheme.
-    if opts.ssl_certificate or opts.ssl_trustfile:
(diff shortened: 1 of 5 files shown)


kgiusti:
I'd recommend getting review input from some of the proton folk on this - possibly astitcher and/or cjansen.

python/qpid_dispatch_internal/tools/command.py
```diff
@@ -261,9 +261,8 @@ def opts_url(opts):
     url = Url(opts.bus)
 
     # If the options indicate SSL, make sure we use the amqps scheme.
-    if opts.ssl_certificate or opts.ssl_trustfile:
+    if opts.ssl_certificate or opts.ssl_trustfile or opts.bus.startswith("amqps"):
```
astitcher, line 264: This looks wrong shouldn't it be ...startswith("amqps:")? Otherwise what happens if you connect to amqps.com?

PR #1270, opened 2021-06-22
DISPATCH-2179: Set SSLDomain.VERIFY_PEER or SSLDomain.VERIFY_PEER_NAM…

ganeshmurthy:
…E based on the ssl_disable_peer_name_verify flag. Removed --ssl-disable-peer-name-verify from run_qdstat in system_tests_qdstat

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -711,7 +711,6 @@ static void on_connection_bound(qd_server_t *server, pn_event_t *e) {
             pn_ssl_domain_t* plugin_ssl_domain = NULL;
             if (config->sasl_plugin_config.use_ssl) {
                 plugin_ssl_domain = pn_ssl_domain(PN_SSL_MODE_CLIENT);
-                pn_ssl_domain_set_peer_authentication(plugin_ssl_domain, PN_SSL_ANONYMOUS_PEER, NULL);
                 if (config->sasl_plugin_config.ssl_certificate_file) {
                     if (pn_ssl_domain_set_credentials(plugin_ssl_domain,
                                                       config->sasl_plugin_config.ssl_certificate_file,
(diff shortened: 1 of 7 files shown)


src/server.c
```diff
@@ -711,7 +711,6 @@ static void on_connection_bound(qd_server_t *server, pn_event_t *e) {
             pn_ssl_domain_t* plugin_ssl_domain = NULL;
             if (config->sasl_plugin_config.use_ssl) {
                 plugin_ssl_domain = pn_ssl_domain(PN_SSL_MODE_CLIENT);
-
```
kgiusti, line 714: Does it make sense to patch this file simply to remove whitespace?

ganeshmurthy: Sorry about that. Fixed now ! Thanks

tests/system_tests_user_id.py
```diff
@@ -242,7 +242,7 @@ def create_ssl_domain(self, ssl_options_dict, mode=SSLDomain.MODE_CLIENT):
         domain = SSLDomain(mode)
         if trustfile:
             domain.set_trusted_ca_db(str(trustfile))
-            domain.set_peer_authentication(SSLDomain.VERIFY_PEER, str(trustfile))
+        domain.set_peer_authentication(SSLDomain.VERIFY_PEER, str(trustfile))
```
kgiusti, line 245: if trustfile is false or None, what does 'str(trustfile)' return?    Should there instead be a "else" to the trustfile test where None is passed as the second arg to set_peer_authentication?

ganeshmurthy: Fixed to send in None instead of str(trustfile)

kgiusti (approved):

PR #1273, opened 2021-06-24
DISPATCH-2184: invoke Q2 unblock handler on Q2 disable

kgiusti:


src/router_node.c
```diff
@@ -2108,10 +2112,6 @@ static void CORE_delivery_update(void *context, qdr_delivery_t *dlv, uint64_t di
                 // and if it is blocked by Q2 holdoff, get the link rolling again.
                 //
                 qd_message_Q2_holdoff_disable(msg);
-
-                qd_link_t_sp *safe_ptr = NEW(qd_link_t_sp);
-                set_safe_ptr_qd_link_t(link, safe_ptr);
-                qd_connection_invoke_deferred(qd_conn, deferred_AMQP_rx_handler, safe_ptr);
```
ganeshmurthy, line 2114: So no need to do invoke the deferred_AMQP_rx_handler anymore

kgiusti: The change to qd_message_Q2_holdoff_disable() now calls it automagically only if necessary (actually in Q2 block).  If you look at qd_message_receive() you'll see where qd_message_set_q2_unblocked_handler() is called to register the deferred_AMQP_rx_handler when the message is first created.

With this change we don't have to take the extra step to explicitly schedule the handler every time we want to disable Q2.

ganeshmurthy (approved):

PR #1275, opened 2021-06-28
DISPATCH-1539: remove Python 2 idioms and compatibiltiy constructs

jiridanek:


jiridanek:
I tried to use `2to3` script, but  many additional changes were manual.

I'll have to go over this later; there is lots of little changes that I should look over few times, to try to catch any error I might've introduced in the manual fixes.

kgiusti (approved):

PR #1276, opened 2021-06-28
DISPATCH-2186 Mark test helpers so that PyTest does not attempt test collection on them

jiridanek:


ganeshmurthy:
So, this change is to suppress test collection on those classes so that the following warning does not show up ?
` /home/runner/work/qpid-dispatch/qpid-dispatch/qpid-dispatch/tests/system_test.py:1259: PytestCollectionWarning: cannot collect test class 'TestTimeout' because it has a __init__ constructor (from: )`

jiridanek:
> So, this change is to suppress test collection on those classes so that the following warning does not show up ?

Exactly. And the test collection should be suppressed, because those classes contain no unittests methods of their own.

ganeshmurthy (approved):

PR #1279, opened 2021-06-28
DISPATCH-1539: Remove the remainder of py2 code from Dispatch

jiridanek:


.travis.yml
```diff
@@ -118,8 +119,6 @@ jobs:
     os: linux
     dist: focal
     compiler: clang
-    python:
-      - "3.9"
```
jiridanek, line 122: this setting was ignored anyways (requires `language: python`, not `language: c`)

tests/management_test/schema.py
```diff
@@ -22,19 +22,9 @@
...
 from qpid_dispatch_internal.management.schema import Schema, BooleanType, EnumType, AttributeType, ValidationError, EnumValue, EntityType
 
 
-def replace_od(thing):
-    """Replace OrderedDict with dict"""
-    if isinstance(thing, OrderedDict):
-        return dict((k, replace_od(v)) for k, v in thing.items())
-    if isinstance(thing, list):
-        return [replace_od(t) for t in thing]
-    return thing
```
jiridanek, line 35: dead code

gemmellr:
Did this perhaps miss https://github.com/apache/qpid-dispatch/blob/44c14b799811a0898a47eee94ac7babc7fbb30dd/python/qpid_dispatch_internal/compat/__init__.py? Seems to contain related code to that noted dead above.

jiridanek:
@gemmellr The OrderedDict function? Yes; on outward inspection, that should not be needed, but it is not actually dead, it is still being called. The UNICODE and BINARY functions are actually used and as far as I can tell they make sense even in Python 3 world. There are some uses that I can eliminate! Calling `UNICODE('literal string')` does not make sense. Good catch. Also, files can be opened as utf-8 text files, removing need for later conversion there. But I expect some uses will still remain.

Maybe the package should be renamed from `compat` to something like `utils`?

gemmellr:
Yep, it was mainly the name and description, and the fact it also did similar dict things to those removed, that made me raise it....if the code makes sense it should be kept, just maybe not there.

PR #1280, opened 2021-06-29
DISPATCH-1619 Close connections created in system_test by wait_router_connected

jiridanek:


ChugR:
+1, Yes, use 'expected'.

jiridanek:
The "obvious" patch did not fix macOS; I'll have to investigate more.

jiridanek:
At least now I know that reverting f42ea178a876a6367372daaa444d6c4287fb4b57 fixes the macOS build. Now I just need to find where FDs are leaking or are used in excessive numbers. Or maybe I need to increase the limit ;P

ganeshmurthy (approved):

PR #1281, opened 2021-07-01
DISPATCH-2189: fix 100-Continue bug, add curl tests

kgiusti:


tests/http1_tests.py
```diff
@@ -1353,3 +1356,189 @@ def client_response_close_test(self, server_port, client_port):
...
         client.wait()
         server.wait()
+
+
+class Http1CurlTestsMixIn(object):
+    """
+    Test cases using curl as the command line client
+    """
+
+    def _run_curl(self, args, input=None, timeout=TIMEOUT):
```
ganeshmurthy, line 1366: Can we put this run_curl in say system_test.py so the http2 can use it as well. Otherwise this will go the way of run_qdstat, there are so many of those all over the place, most of them doing the same thing.

kgiusti: Done.  Updated http2 tests as well.

ganeshmurthy (approved):

PR #1300, opened 2021-07-16
DISPATCH-2199: Function cleanup_link should use qd_alloc_deref_safe_ptr

ChugR:


kgiusti (approved):
+1, but I recommend going a bit further and replacing the the code on lines 1228+1229 with at call to qd_alloc_set_safe_ptr() rather than mucking with the structure directly....

PR #1303, opened 2021-07-17
DISPATCH-2200 Add basic TCP adaptor microbenchmarks

jiridanek:


tests/c_benchmarks/bm_tcp_adapter.cpp
```diff
@@ -103,3 +221,143 @@ static void BM_TCPEchoServerLatencyWithoutQDR(benchmark::State &state)
...
+};
+
+class DispatchRouterSubprocessTcpLatencyTest
+{
+    int pid;
+
+   public:
+    DispatchRouterSubprocessTcpLatencyTest(std::string configName)
+    {
+        pid = fork();
```
jiridanek, line 262: I really like this! Since dispatch is compiled in, it can be exec'd by forking the test/benchmark process. With (considerable) work, it would be possible to even have two-way interactions with the forked process, using some Linux IPC... Just running qdmanage is simpler and probably better, though.

jiridanek:
I think it's best to merge this now, and do the pass/fail part of the jira in a next PR. I've been having problems with TCP stalls when I have more than 2 routers in the network. There are already some jiras for when that bug may happen...

PR #1305, opened 2021-07-20
DISPATCH-2204 Fix "dictionary changed size during iteration" errors

jiridanek:
I've reviewed #1275 (again) and I found nothing more. Hopefully this will be the last fixup.

### DIFF
--- a/src/adaptors/tcp_adaptor.c
+++ b/src/adaptors/tcp_adaptor.c
@@ -1340,7 +1340,7 @@ void qd_dispatch_delete_tcp_connector(qd_dispatch_t *qd, void *impl)
     qd_tcp_connector_t *ct = (qd_tcp_connector_t*) impl;
     if (ct) {
         //need to close the pseudo-connection used for dispatching
-        //deliveries out to live connnections:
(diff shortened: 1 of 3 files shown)


src/adaptors/tcp_adaptor.c
```diff
@@ -1340,7 +1340,7 @@ void qd_dispatch_delete_tcp_connector(qd_dispatch_t *qd, void *impl)
     qd_tcp_connector_t *ct = (qd_tcp_connector_t*) impl;
     if (ct) {
         //need to close the pseudo-connection used for dispatching
-        //deliveries out to live connnections:
+        //deliveries out to live connections:
```
ChugR, line 1343: I'd prefer to see changes like this typo fix pushed as a separate commit. It could be NO-JIRA even. Don't let it stop this PR though.

ChugR (approved):

src/adaptors/tcp_adaptor.c
jiridanek: allright; next time, then

PR #1308, opened 2021-07-21
Dispatch 2166 - Thread safety improvements for message flags accessed by multiple threads

ChugR:
This PR is created with a series of commits where each commit changes one message variable. Small commits make it easier to see how each variable was handled.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -1020,10 +1020,11 @@ qd_message_t *qd_message()
 
     ZERO(msg->content);
     msg->content->lock = sys_mutex();
-    sys_atomic_init(&msg->content->ref_count, 1);
     sys_atomic_init(&msg->content->aborted, 0);
     sys_atomic_init(&msg->content->discard, 0);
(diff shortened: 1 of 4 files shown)


src/message.c
```diff
@@ -1225,7 +1241,7 @@ int qd_message_get_phase_annotation(const qd_message_t *in_msg)
 void qd_message_set_stream_annotation(qd_message_t *in_msg, bool stream)
 {
     qd_message_pvt_t *msg = (qd_message_pvt_t*) in_msg;
-    msg->content->ma_stream = stream;
+    sys_atomic_set(&msg->content->ma_stream, (stream ? 1 : 0));
```
ganeshmurthy, line 1244: just a thought. Instead of calling sys_atomic_set can we do the following - 
if (stream) {
SET_ATOMIC_FLAG(&msg->content->ma_stream)
}
else {
CLEAR_ATOMIC_FLAG(&msg->content->ma_stream)
}

kgiusti (approved):

PR #1310, opened 2021-07-21
DISPATCH-2207: fix qd_connection_t leak

kgiusti:


src/server.c
```diff
@@ -1352,9 +1352,12 @@ static void try_open_cb(void *context)
         // else deleted or failed - on failed wait until after connection is freed
         // and state is set to CXTR_STATE_CONNECTING (timer is rescheduled then)
         try_open_lh(ct, ctx);
+        ctx = 0;  // owned by ct
     }
 
     sys_mutex_unlock(ct->lock);
+
+    free_qd_connection_t(ctx);  // noop if ctx == 0
```
ganeshmurthy, line 1360: My comment is not about this fix but about the readability of this code. A previous commit moved the creation of qd_connection_t from inside of try_open_lh() to try_open_cb() and this is to avoid the lock issues caused by the connector-entity cache lock inversion. From my point of view the previous code was super easy to understand. This code is not. Instead of getting rid of the entity cache concept completely, we are keeping it and making the code much less readable.

ganeshmurthy (approved):

PR #1311, opened 2021-07-22
Mick log rewrite 2

mgoulish:


kgiusti:
This patch LGTM, but I'd like to land this right after we drop the freeze for 1.17.0 with the goal of having a complete fix for [DISPATCH-1956](https://issues.apache.org/jira/browse/DISPATCH-1956) in 1.18.0.

kgiusti (approved):

PR #1312, opened 2021-07-22
DISPATCH-2209: HTTP/1.x use same meta-data keys as HTTP/2 adaptor

kgiusti:


grs (approved):
Do we need :method and :status as separate properties if they are also conveyed in subject for request and response respectively?

kgiusti:
> Do we need :method and :status as separate properties if they are also conveyed in subject for request and response respectively?

@grs you're correct - the app properties are duplicate.  However IIUC the HTTP/2 adaptor also puts :method and :status in the app properties BUT doesn't set a Subject property at all.

@ganeshmurthy what do you think: in the HTTP/1 adaptor, the AMQP Subject property contains the request method (e.g. GET) in the AMQP mapping for a request method, and sets Subject to the response code (:status, e.g. 200) in the response AMQP message.

Should both adaptors remove the :status and :method app properties and just use the Subject field instead?

ganeshmurthy (approved):

PR #1314, opened 2021-07-22
DISPATCH-2037: Removed ping feature

ganeshmurthy:


kgiusti:
Conditional approval: the buffer leak error should be fixed when DISPATCH-2211.  Rebase this after 2211 lands and feel free to land if no more buffer leaks.  
Thanks

kgiusti (approved):

PR #1316, opened 2021-07-23
DISPATCH-2210 clarify scale

pwright:


docs/books/modules/user-guide/network-topologies.adoc
```diff
@@ -62,11 +61,22 @@ of the configuration:
...
 
+[NOTE]
+====
+Performance of your router network is determined by various factors:
+
+* topology
+* number of routers
+* underlying infrastructure (host resources, network speed, etc)
+
+It is recommended to implement an architecture that minimises the number
```
kgiusti, line 76: I'm a little uncomfortable with this recommendation - it can be interpreted as discouraging redundancy, which may be a key benefit to a particular use case.

The topic of router "performance" is a complex one - I don't think a simple note like this adds much value.  Probably better to simply remove this note.

pwright: done

ChugR (approved):

kgiusti (changes requested):
The built documentation is still referencing the 128 interior scale numbers.  Apparently that text is also in docs/books/modules/user-guide/router-operating-modes.adoc, which needs to be changed as well.

pwright:
@kgiusti yipes, I'd been editing the wrong file (but with similar content). That file wasn't used, so removing.

kgiusti (approved):
Looks great - thanks pwright!

PR #1322, opened 2021-07-26
NO-JIRA: update github actions

kgiusti:
- Add Debug build type
  - Disable debug/assertions on TSAN tests

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,13 +29,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        buildType: [RelWithDebInfo]
+        buildType: [Debug]
         runtimeCheck: [asan]
     env:
       BuildType: ${{matrix.buildType}}
       ProtonBuildDir: ${{github.workspace}}/qpid-proton/build
       DispatchBuildDir: ${{github.workspace}}/qpid-dispatch/build
       InstallPrefix: ${{github.workspace}}/install
+      VERBOSE: 1
 
       ProtonCMakeExtraArgs: >
         -DCMAKE_C_COMPILER_LAUNCHER=ccache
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -173,7 +173,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        buildType: [RelWithDebInfo]
+        buildType: [Debug]
```
jiridanek, line 176: Using a Debug build makes the CI run maybe ~1 minute longer ;P Users are not running debug builds in production, therefore it would make sense to me to test mostly with RelWithDebInfo.

I think there should be maybe one debug build in the CI, not all of them. BTW, Coverity warns if there are C asserts that have side-effects, so I presume debug builds only serve to introduce some timing randomness into tests...

.github/workflows/build.yaml
```diff
@@ -378,14 +378,22 @@ jobs:
       - name: Display ccache stats
         run: ccache -s
 
+      - name: enable asserts on asan build
+        if: matrix.runtimeCheck == 'asan' || matrix.runtimeCheck == 'OFF'
+        run: echo "DispatchCMakeAsserts=ON" >> $GITHUB_ENV
+
+      - name: disable asserts on tsan build
+        if: matrix.runtimeCheck == 'tsan'
+        run: echo "DispatchCMakeAsserts=OFF" >> $GITHUB_ENV
```
jiridanek, line 387: This does not work. Debug builds always have asserts enabled; `-DQD_ENABLE_ASSERTIONS` is only a way to turn them on where they would be otherwise disabled.

Notice that on Debug builds, the option does not undefine anything (which would be necessary to disable asserts):

https://github.com/apache/qpid-dispatch/blob/main/CMakeLists.txt#L218

jiridanek (changes requested):

jiridanek (approved):
Ok, never mind what I wrote above. Ubuntu build is a debug builds, the rest are still RelWithDebInfo. So turning the asserts on and off on the Fedora builds does work.

.github/workflows/build.yaml
```diff
@@ -29,13 +29,14 @@ jobs:
...
         os: [ubuntu-20.04]
-        buildType: [RelWithDebInfo]
+        buildType: [Debug]
         runtimeCheck: [asan]
     env:
       BuildType: ${{matrix.buildType}}
       ProtonBuildDir: ${{github.workspace}}/qpid-proton/build
       DispatchBuildDir: ${{github.workspace}}/qpid-dispatch/build
       InstallPrefix: ${{github.workspace}}/install
+      VERBOSE: 1
```
jiridanek, line 39: :+1:

PR #1324, opened 2021-07-26
DISPATCH-2198: Improvement - Call qd_message_stream_data_buffers only…

ganeshmurthy:
… once so buffers need to iterated only once

### DIFF
--- a/src/adaptors/http2/http2_adaptor.c
+++ b/src/adaptors/http2/http2_adaptor.c
@@ -625,15 +625,17 @@ static int snd_data_callback(nghttp2_session *session,
         // Insert the framehd of length 9 bytes into the buffer
         memcpy(qd_http2_buffer_cursor(http2_buff), framehd, HTTP2_DATA_FRAME_HEADER_LENGTH);
         qd_http2_buffer_insert(http2_buff, HTTP2_DATA_FRAME_HEADER_LENGTH);
-        pn_raw_buffer_t pn_raw_buffs[stream_data->qd_buffers_to_send];
-        int written = qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, stream_data->curr_stream_data_qd_buff_offset, stream_data->qd_buffers_to_send);
+
+        size_t diff = qd_message_stream_data_buffer_count(stream_data->curr_stream_data) - stream_data->curr_stream_data_qd_buff_offset;
+        pn_raw_buffer_t pn_raw_buffs[diff];
+        int written = qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, stream_data->curr_stream_data_qd_buff_offset, diff);
         (void)written;
-        assert (written == stream_data->qd_buffers_to_send);
+        assert(diff == written);
 
         int idx = 0;
         size_t bytes_to_send = length;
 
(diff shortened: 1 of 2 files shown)


src/adaptors/http2/http2_adaptor.c
```diff
@@ -625,15 +625,17 @@ static int snd_data_callback(nghttp2_session *session,
         // Insert the framehd of length 9 bytes into the buffer
         memcpy(qd_http2_buffer_cursor(http2_buff), framehd, HTTP2_DATA_FRAME_HEADER_LENGTH);
         qd_http2_buffer_insert(http2_buff, HTTP2_DATA_FRAME_HEADER_LENGTH);
-        pn_raw_buffer_t pn_raw_buffs[stream_data->qd_buffers_to_send];
-        int written = qd_message_stream_data_buffers(stream_data->curr_stream_data, pn_raw_buffs, stream_data->curr_stream_data_qd_buff_offset, stream_data->qd_buffers_to_send);
+
+        size_t diff = qd_message_stream_data_buffer_count(stream_data->curr_stream_data) - stream_data->curr_stream_data_qd_buff_offset;
+        pn_raw_buffer_t pn_raw_buffs[diff];
```
kgiusti, line 630: I thought we're avoiding VLA's, no?

src/adaptors/http2/http2_adaptor.c
```diff
@@ -1330,14 +1314,12 @@ ssize_t read_data_callback(nghttp2_session *session,
...
                 		bytes_to_send = length;
-                		set_buffers_to_send(stream_data, bytes_to_send);
                 		stream_data->full_payload_handled = false;
                 	}
                 	else {
                 		bytes_to_send = remaining_payload_length;
-                        set_buffers_to_send(stream_data, bytes_to_send);
                 		stream_data->full_payload_handled = true;
-                		qd_log(http2_adaptor->protocol_log_source, QD_LOG_TRACE, "[C%"PRIu64"][S%"PRId32"] read_data_callback remaining_payload_length (%zu) <= QD_HTTP2_BUFFER_SIZE(16384), bytes_to_send=%zu, stream_data->qd_buffers_to_send=%zu", conn->conn_id, stream_data->stream_id, remaining_payload_length, bytes_to_send, stream_data->qd_buffers_to_send);
+                		qd_log(http2_adaptor->protocol_log_source, QD_LOG_TRACE, "[C%"PRIu64"][S%"PRId32"] read_data_callback remaining_payload_length (%zu) <= QD_HTTP2_BUFFER_SIZE(16384), bytes_to_send=%zu", conn->conn_id, stream_data->stream_id, remaining_payload_length, bytes_to_send);
```
kgiusti, line 1322: should "remaining_payload_length" covert using %i instead?

ganeshmurthy: Yes, that is correct. https://github.com/apache/qpid-dispatch/commit/6d7e8c2d8bbf1645826064c3bf0074da89366071 took care of that already. I will rebase this PR

src/adaptors/http2/http2_adaptor.c
ganeshmurthy: Having VLAs is ok here I think. The max value for diff is 32 which is the number of qd_buffers in one 16k http2_buffer

PR #1326, opened 2021-07-26
DISPATCH-2168: Fix error that showed up in Coverity. Check for addr b…

ganeshmurthy:
…efore dereferencing it


kgiusti (approved):
LGTM.  Let's get this on main and cherry pick it over to 1.17.x

PR #1327, opened 2021-07-27
DISPATCH-2214: Reduce logging in tcp_adaptor self test

ChugR:
* Put restraints on echo tests startup wait loop
   Only poll N times; delay between polls
 * Print qdsdat address displays if polling fails

### DIFF
--- a/tests/system_tests_tcp_adaptor.py
+++ b/tests/system_tests_tcp_adaptor.py
@@ -322,7 +322,7 @@ class TcpAdaptor(TestCase):
 
         # define logging levels
         cls.print_logs_server = False
-        cls.print_logs_client = True
+        cls.print_logs_client = False
         parent_path = os.path.dirname(os.getcwd())
         cls.logger = Logger(title="TcpAdaptor-testClass",
                             print_to_console=True,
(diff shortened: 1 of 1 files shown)


ganeshmurthy (approved):

tests/system_tests_tcp_adaptor.py
```diff
@@ -528,6 +530,15 @@ def router(name, mode, connection, extra=None):
                     unseen = [srv for srv in cls.router_order if "ES_" + srv not in seen]
                     cls.logger.log("TCP_TEST Router %s sees only %d of %d addresses. Waiting for %s" %
                                    (rtr, len(server_lines), len(cls.router_order), unseen))
+                if poll_loops == 1:
+                    # last poll loop
+                    for line in lines:
+                        cls.logger.log("TCP_TEST Router %s : %s" % (rtr, line))
+            poll_loops -= 1
+            if poll_loops == 0:
+                assert False, "TCP_TEST TCP_Adaptor test setup failed. Echo tests never executed."
```
jiridanek, line 539:
```suggestion
                self.fail("TCP_TEST TCP_Adaptor test setup failed. Echo tests never executed.")
```
https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail

PR #1346, opened 2021-08-23
DISPATCH-2235: Dump time delay before lock is acquired

kgiusti:
- top five total waits:
  $ sort --reverse --numeric --key 3 <output-file> | head -5
- top five with waits > 1000 nsc:
  $ sort --reverse --numeric --key 7 <output-file> | head -5
- top five maximum single-call wait times:
  sort --reverse --numeric --key 9 <output-file> | head -5

### DIFF
--- a/src/router_core/forwarder.c
+++ b/src/router_core/forwarder.c
@@ -154,7 +154,7 @@ qdr_delivery_t *qdr_forward_new_delivery_CT(qdr_core_t *core, qdr_delivery_t *in
     out_dlv->delivery_id = next_delivery_id();
     out_dlv->link_id     = out_link->identity;
     out_dlv->conn_id     = out_link->conn_id;
-    out_dlv->dispo_lock  = sys_mutex();
+    out_dlv->dispo_lock  = sys_mutex("dlv-dispo");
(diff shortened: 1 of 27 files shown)


src/router_core/forwarder.c
```diff
@@ -154,7 +154,9 @@ qdr_delivery_t *qdr_forward_new_delivery_CT(qdr_core_t *core, qdr_delivery_t *in
     out_dlv->delivery_id = next_delivery_id();
     out_dlv->link_id     = out_link->identity;
     out_dlv->conn_id     = out_link->conn_id;
-    out_dlv->dispo_lock  = sys_mutex();
+    char name[64];
+    snprintf(name, sizeof(name), "delivery-%"PRIu32, out_dlv->delivery_id);
```
ganeshmurthy, line 158: This snprintf is in the fast path. Is this a problem ?

kgiusti: It's probably overkill, especially for this lock which is hardly used.
In fact we probably want to avoid storing the mutex name (strdup call) when we're not explicitly debugging the locks.
Update coming soon...

ganeshmurthy (approved):

PR #1351, opened 2021-08-26
NO-JIRA: remove those travis test combinations that duplicate GHA tests

kgiusti:


jiridanek:
@kgiusti Can you please remind me what is the problem we are solving here? What is the motivating principle? "Tests that are deleted do not fail"? "Frugal with ASF resources you shall be, and you will have long and satisfied life"? "Travis runs so slow that I don't even look at the results anymore"? "Travis jobs are not useful because we cannot debug the failures anyways"?

kgiusti:
> @kgiusti Can you please remind me what is the problem we are solving here? What is the motivating principle? "Tests that are deleted do not fail"? "Frugal with ASF resources you shall be, and you will have long and satisfied life"? "Travis runs so slow that I don't even look at the results anymore"? "Travis jobs are not useful because we cannot debug the failures anyways"?

Pretty much all of the above to some degree, with real emphasis on reducing the overlap with GHA in order to lessen our consumption of ASF resources.  Environments like macos are consistently failing - let's avoid using any ASF resources there until someone who needs macos support and is willing to invest time to fix and maintain it going forward.

TSAN is valuable - seems to be more sensitive on travis (timing?) and the errors TSAN logs are easy to access.
Code Coverage is also useful (assuming people actually check the analysis - which I highly encourage folks to do)

kgiusti:
@jiridanek:

And to your point (re: usefulness):  it takes way too long to get feedback from travis for _pull request verification_.  Checking the history we're talking ~5 hours plus to get through travis.  GHA runs the _same tests_ in under 30 minutes - which gives me enough confidence that my PR is solid enough for review.

Where I think travis can give us value is testing alternate platforms and tooling, like ARM and clang, which aren't currently covered in GHA.  Also travis seems to have a knack for finding TSAN issues better than GHA - for whatever reason.  Let's leverage that strength as well.

So here's what this patch proposes: 
- reduce redundancies with GHA, favoring GHA over travis given the fast turn around time.
- reduce the load on travis by trimming the test cases back to those that provide alternative test configurations from what we're already covering in GHA.  This results in a quicker test run as well as lessening the burden on ASF resources.

The patch does this by reducing the travis test runs to the following:
- Coverage (which generates reports on our test effectiveness)
- Default Router build using clang
- TSAN checking, esp. on non-x86_64 architectures: arm64 and s390x

.travis.yml
```diff
@@ -20,83 +20,53 @@
...
-    - PATH="/usr/bin:$PATH" PROTON_VERSION=0.36.0 BUILD_TYPE=RelWithDebInfo
-    - DISPATCH_CMAKE_ARGS='-DRUNTIME_CHECK=asan -DCMAKE_C_FLAGS=-DQD_MEMORY_DEBUG'
-  - name: "qdrouterd:RelWithDebInfo+MemoryDebug (clang on focal)"
-    os: linux
-    dist: focal
-    before_install:
-    - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key 2>/dev/null | sudo apt-key add -
-    - sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main' -y
-    - sudo apt-get update -q
-    - sudo apt-get install -y clang-13 llvm-13-dev
```
jiridanek, line 77: there used to be clang 13 for the amd64 linux tsan job

jiridanek:
> And to your point (re: usefulness): it takes way too long to get feedback from travis for pull request verification. Checking the history we're talking ~5 hours plus to get through travis. GHA runs the same tests in under 30 minutes - which gives me enough confidence that my PR is solid enough for review.

Meaning that Travis is useless for PRs, regardless of what we do or do not do in this PR. The wait queue is always going to be log.

> Where I think travis can give us value is testing alternate platforms and tooling, like ARM and clang, which aren't currently covered in GHA. Also travis seems to have a knack for finding TSAN issues better than GHA - for whatever reason. Let's leverage that strength as well.

Agreed.

> TSAN checking, esp. on non-x86_64 architectures: arm64 and s390x

I wonder whether the Tsan on non-x86 is somehow different than the x86 variant. I don't think so. The checks the software performs are still the same, what's different is that the synchronization issues may also have more visible actual consequences, but that is irrespective of the sanitizer. In fact, tsan can make e.g. arm less sensitive to synch issues:

> *force_seq_cst_atomics* This flag turns all atomic operations into sequentially consistent operations. This is useful to check whether a race is caused by incorrect memory ordering or not.
> -- https://github.com/google/sanitizers/wiki/ThreadSanitizerFlags

There is a fundamentally different ASAN for arm64, https://stackoverflow.com/search?q=user%3A1047788+aarch64+deleted%3Ayes

jiridanek (approved):
lgtm, but maybe undo the clang downgrade

in the future I really should revisit hwasan, sometimes; it was broken the last time I checked, that was before clang 12, I think

PR #1357, opened 2021-08-27
DISPATCH-840 Properly escape the token `${user}` in asciidoc

jiridanek:
CC @pwright


jiridanek:
I haven't figured out how to get any kind of diagnostics when `'${user}'` is encountered. Having this would be useful, so we spot the issue if something like this ever appears again.

bhardesty (approved):
Looks good! Just a minor suggestion: since you're changing ${user} to be styled as a literal in the adoc, you could also change the other entity names to be styled the same way. For example, `targetPattern` instead of 'targetPattern'.

jiridanek:
`${user}` is not entity name, it is, ... something else; part of the string which has special interpretation

I think I'm only going to fix the doc rendering error in this way, and I will leave formatting changes to people who edit with the docs more and have good overview of the thing in its entirety. Thanks for checking the PR!

PR #1358, opened 2021-08-29
DISPATCH-413 Reference the AMQP Management spec in docs in a less ambiguous manner

jiridanek:
The spec in question is [AMQP Management v1.0 WD09](https://www.oasis-open.org/committees/document.php?document_id=54441&wg_abbrev=amqp). I managed to find a newer working draft WD12 of the spec, but I did not manage to download a readable copy of it. It looks like the spec never got past working drafts anyways.

CC @pwright


pwright:
@jiridanek looks good, note that we normally[1] refer to 1.0 spec at http://www.amqp.org/resources/download 

[1] https://qpid.apache.org/releases/qpid-dispatch-1.12.0/user-guide/index.html#overview-of-amqp-qdr

jiridanek:
This is not part of AMQP 1.0 spec. It is some kind of extension on top, if I understand the situation correctly.

PR #1365, opened 2021-09-08
 DISPATCH-2246 Use port 0 in systemtests when starting TCP_echo_server, then query for actual port

jiridanek:


tests/system_tests_tcp_adaptor.py
```diff
@@ -347,12 +342,12 @@ def router(name, mode, connection, extra=None):
                                    ofilename=os.path.join(parent_path, "setUpClass/TcpAdaptor_echo_server_%s.log" % rtr))
             cls.logger.log("TCP_TEST Launching echo server '%s'" % server_prefix)
             server = TcpEchoServer(prefix=server_prefix,
-                                   port=cls.tcp_server_listener_ports[rtr],
+                                   port=0,
                                    logger=server_logger)
             assert server.is_running
```
jiridanek, line 347: Previously, this suffered from a race condition; what if the new thread took its time setting this variable?

Not that I ever saw this specific assert to fail, to be honest.

kgiusti (approved):

PR #1366, opened 2021-09-09
Rewrite of log source locking.

mgoulish:
* Individual locks for each log source -- no more global lock for log sources.
* New global lock for recent log entries list.
* A big comment explaining the new overall locking approach.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -26,6 +26,7 @@
 #include "entity_cache.h"
 #include "log_private.h"
 #include "server_private.h"
+#include "schema_enum.h"
 
 #include "qpid/dispatch/alloc.h"
 #include "qpid/dispatch/atomic.h"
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -59,12 +88,11 @@ struct qd_log_entry_t {
     struct timeval  time;
     char            text[TEXT_MAX];
 };
-
 ALLOC_DECLARE(qd_log_entry_t);
 ALLOC_DEFINE(qd_log_entry_t);
-
 DEQ_DECLARE(qd_log_entry_t, qd_log_list_t);
 static qd_log_list_t         entries = {0};
+sys_mutex_t * entries_lock = 0;
```
kgiusti, line 95: code convention: remove space between * and entries_lock please

kgiusti (changes requested):
One of the travis tests (system_tests_qdmanage)  hit a use-after-free associated with a log_sink instance:

https://app.travis-ci.com/github/apache/qpid-dispatch/jobs/536456829#L4880

Not sure exactly why that happened - locking seems correct to me.  Maybe toss in some debug code to print out the sink reference counts to see where the issue is.

kgiusti (approved):

PR #1376, opened 2021-09-29
DISPATCH-1956: log.c rewrite to reduce locking scope

mgoulish:
WIP: disagreement with tsan.


kgiusti (changes requested):
This patch causes the system_tests_management.py::test_log() test to fail.
Best I can tell, at this line: 
https://github.com/apache/qpid-dispatch/blob/main/tests/system_tests_management.py#L260
The log is re-configured to set the log level to debug only, but the output log has error logs in it as well.

mgoulish:
Closing in favor of new PR.

PR #1378, opened 2021-09-30
DISPATCH-1487: optimize view_initialize for NODE_HASH views

kgiusti:
Avoid needlessly searching the iterator for ":" when in NODE_HASH view:  Node hash is used on node identifiers which do not contain ":" - they are not URL formatted.

### DIFF
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -287,6 +287,15 @@ static void view_initialize(qd_iterator_t *iter)
         return;
 
     //
+    // ITERN_VIEW_NODE_HASH has no scheme or leading slash - see iterator.h
+    // Do not walk the entire iterator trying to find one.
+    //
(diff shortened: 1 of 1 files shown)


ganeshmurthy (approved):

src/iterator.c
```diff
@@ -286,6 +286,15 @@ static void view_initialize(qd_iterator_t *iter)
     if (iter->view == ITER_VIEW_ALL)
         return;
 
+    //
+    // ITERN_VIEW_NODE_HASH has no scheme or leading slash - see iterator.h
```
gemmellr, line 290:
```suggestion
    // ITER_VIEW_NODE_HASH has no scheme or leading slash - see iterator.h
```
Wayward N?

kgiusti: oops! nice catch.  Will add to patch.

PR #1409, opened 2021-10-31
DISPATCH-2270 Use PyLong_FromVoidPtr to reliably pass void* values into Python regardless of platform type sizes

jiridanek:


jiridanek:
This is IMO useful on all platforms, so it can be merged even if Windows is to remain unsupported.

PR #1413, opened 2021-10-31
DISPATCH-2272 Mark all functions called from Python with QD_EXPORT

jiridanek:


kgiusti (changes requested):
Please raise a discussion on the qpid-dispatch apache mailing list regarding win32 support.
Adding win32 support is a big deal and needs to be discussed at the project level.

jiridanek:
> Please raise a discussion on the qpid-dispatch apache mailing list regarding win32 support. Adding win32 support is a big deal and needs to be discussed at the project level.

It has been discussed multiple times, see the "Previous discussions" section on https://issues.apache.org/jira/browse/DISPATCH-2027. Why do you think there is something more to say than what was already mentioned?

Actually, from my experience, it is not such a big deal. It greatly depends on what do you mean by "Windows support". What @ppatierno did in the tweet linked in Jira (running Dispatch in Windows Subsystem for Linux) was AFAIK quite simple user-level effort, which nevertheless got you there. My goal is to be able to build a native Windows binary in Visual Studio, but using the Clang compiler (VS offers a choice of MSVC and Clang on Windows; Clang obviously requires less effort). That is somewhat bigger deal, but nothing too crazy.

@kgiusti Do I need to create a new thread, or can I reuse one of the existing ones?

jiridanek:
Regarding the usefulness of QD_EXPORT even on Linux: It would provide a way to build dispatch code into a static library. Currently that is not workable because compiler/linker would eliminate functions it seems unused. Adding a suitable ` __attribute__` would mark the functions so that they are not dropped.

kgiusti:
Interesting point re: static builds.  I'd like to get the opinions of @ted-ross and @ganeshmurthy since this seems - correct me if I'm wrong - that we'd need devs to be consistent with the use of the new QD_EXPORT qualifier when adding/modifying new declarations in the codebase.

include/qpid/dispatch/internal/export.h
```diff
@@ -0,0 +1,27 @@
...
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+
+/// @cond INTERNAL
+
+/// import/export macros
+#if defined(_WIN32)
```
ganeshmurthy, line 23: I suggest we remove any code that accommodates WIN32. In fact, I would recommend that we go thru the already existing Dispatch code and remove any code that has this kind of #if defined(_WIN32)
The reason I say this is that we don't want to have any code that even hints at anything to do with WIN32

include/qpid/dispatch/internal/export.h
```diff
@@ -0,0 +1,27 @@
...
+#  define QD_EXPORT __declspec(dllexport)
+#else
+#  define QD_EXPORT __attribute__((visibility("default"))) __attribute__((used))
```
ganeshmurthy, line 26: I was looking at this - https://developer.arm.com/documentation/dui0491/f/Compiler-specific-Features/--attribute----used---function-attribute?lang=en

It seems to say that "This function attribute informs the compiler that a static function is to be retained in the object file, even if it is unreferenced."

Is it true that __attribute__((used)) can *only* be used on *static* functions ?

I also understand from reading this - https://developer.arm.com/documentation/dui0491/f/BABCJJID - that __attribute__((used)) can also be used with  *static* variables, although I understand that this PR is not attempting to do so.

include/qpid/dispatch/internal/export.h
jiridanek: In this PR, I agree.

In the entire codebase, I think it is pointless, since there are (as far as I know only two instances: in the doctest.h framework, which we should not edit, and in CMakeLists.txt which is code "borrowed" from LLVM build scripts, which is also rather unprofitable to modify, in my view).

include/qpid/dispatch/internal/export.h
jiridanek: In GCC docs it says nothing about the functions having to be static, https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

> This attribute, attached to a function, means that code must be emitted for the function even if it appears that the function is not referenced. This is useful, for example, when the function is referenced only in inline assembly.

I do remember I had the dead code elimination issue when I tried to switch from libqpiddispatch.so to a different way of compiling the code. Because I couldn't easily have the .a library due to dead code elimination by the linker, I had to do the CMake object library, which is cumbersome in old versions of CMake.

jiridanek:
> @kgiusti that we'd need devs to be consistent with the use of the new QD_EXPORT qualifier when adding/modifying new declarations in the codebase.

Yes, that is true. There is a chunk of code in dispatch which IMO tries to resolve all C functions during startup to discover any possible issues. There are functions missing here, mostly things added for protocol adaptors.

https://github.com/apache/qpid-dispatch/blob/main/python/qpid_dispatch_internal/dispatch.py#L51-L79

include/qpid/dispatch/internal/export.h
ganeshmurthy: > In the entire codebase, I think it is pointless, since there are (as far as I know only two instances: in the doctest.h framework, which we should not edit, and in CMakeLists.txt which is code "borrowed" from LLVM build scripts, which is also rather unprofitable to modify, in my view).

Make sense. Let's remove it from the PR. Thanks.

jiridanek:
> > @kgiusti that we'd need devs to be consistent with the use of the new QD_EXPORT qualifier when adding/modifying new declarations in the codebase.
> 
> Yes, that is true. There is a chunk of code in dispatch which IMO tries to resolve all C functions during startup to discover any possible issues. There are functions missing here, mostly things added for protocol adaptors.
> 
> https://github.com/apache/qpid-dispatch/blob/main/python/qpid_dispatch_internal/dispatch.py#L51-L79

For example, `qd_entity_refresh_allocator` is not in the list above, where function signatures get configured. It is not immediately obvious from grepping the codebase this gets called from Python, because the function name is computed at runtime:

```python
fname = "qd_entity_refresh_" + entity_type.short_name.replace('.', '_')
```

include/qpid/dispatch/internal/export.h
jiridanek: Removed, thanks.

PR #1414, opened 2021-11-01
DISPATCH-2247: Added router id to log message

ganeshmurthy:


src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -690,8 +690,24 @@ static void qcm_mobile_sync_on_mau_CT(qdrm_mobile_sync_t *msync, qd_parsed_field
...
             qdr_post_set_mobile_seq_CT(msync->core, router->mask_bit, mobile_seq);
-        } else
-            qd_log(msync->log, QD_LOG_ERROR, "Received MAU from an unknown router");
+        } else {
+            //
+            // There is a possibility here that router_id is null but that is fine. We want to print it out either way
+            // which will help us in debugging.
+            //
+            char *r_id = 0;
+            bool free_rid = false;
```
ted-ross, line 699: You don't need the free_rid thing.  It is safe to call 'free' on a null pointer.

src/router_core/modules/mobile_sync/mobile.c
```diff
@@ -690,8 +690,24 @@ static void qcm_mobile_sync_on_mau_CT(qdrm_mobile_sync_t *msync, qd_parsed_field
...
+            if (id_field) {
+                qd_iterator_t *id_iter = qd_parse_raw(id_field);
+                if (id_iter) {
+                    r_id = (char *)qd_iterator_copy(id_iter);
+                    free_rid = true;
+                }
+            }
+            qd_log(msync->log, QD_LOG_ERROR, "Received MAU from an unknown router with router id %s", r_id);
```
ted-ross, line 707: There is a similar log for MAR messages.  For completeness, you should probably add the router_id to that one too.

PR #1416, opened 2021-11-02
DISPATCH-1958: Fixed error in qdstat when memory stats are turned off…

ganeshmurthy:
… (-DQD_MEMORY_STATS=OFF)


kgiusti:
hmmm...  I wonder if it would be simpler to make QD_MEMORY_STATS non-configurable?  IOW, get rid of the flag and just always provide memory statistics.  It's pretty obvious we haven't tested QD_MEMORY_STATS=OFF as part of our CI.

Just an opinion, thoughts?

ganeshmurthy:
> hmmm... I wonder if it would be simpler to make QD_MEMORY_STATS non-configurable? IOW, get rid of the flag and just always provide memory statistics. It's pretty obvious we haven't tested QD_MEMORY_STATS=OFF as part of our CI.
> 
> Just an opinion, thoughts?

I am all for getting rid of the flag once and for all and always provide memory statistics but first I want to understand why this flag was introduced in the first place ? 
If we decide to keep this flag, we need to augment CI (preferably GHA) to test with QD_MEMORY_STATS=OFF. That will add *yet another* build type to the already long list of tests we run, so I am inclined to say let's just get rid of this flag.

@ted-ross  what do you think ?

ted-ross:
> > hmmm... I wonder if it would be simpler to make QD_MEMORY_STATS non-configurable? IOW, get rid of the flag and just always provide memory statistics. It's pretty obvious we haven't tested QD_MEMORY_STATS=OFF as part of our CI.
> > Just an opinion, thoughts?
> 
> I am all for getting rid of the flag once and for all and always provide memory statistics but first I want to understand why this flag was introduced in the first place ? If we decide to keep this flag, we need to augment CI (preferably GHA) to test with QD_MEMORY_STATS=OFF. That will add _yet another_ build type to the already long list of tests we run, so I am inclined to say let's just get rid of this flag.
> 
> @ted-ross what do you think ?

+1 on removing the flag.  This was first introduced due to concern that the stats collection might be costly.  I don't believe that it has been identified as a performance/memory bottleneck in any of the investigations we've done.

ted-ross (changes requested):
Let's do this the simple way and remove the flag.

ganeshmurthy:
> Let's do this the simple way and remove the flag.

Sounds good. I will add a commit to this PR removing the flag. Thanks.

ted-ross (approved):

kgiusti (approved):

kgiusti (approved):

PR #1420, opened 2021-11-03
DISPATCH-2065: Added timer to check for deleted connection

ganeshmurthy:


tests/system_tests_two_routers.py
```diff
@@ -439,16 +447,35 @@ def bail(self, error):
 
     def on_link_opened(self, event):
         if event.receiver == self.mgmt_receiver:
+            self.mgmt_receiver_link_opened = True
+        elif event.receiver == self.mgmt_receiver_1:
+            self.mgmt_receiver_1_link_opened = True
+        elif event.receiver == self.mgmt_receiver_2:
+            self.mgmt_receiver_2_link_opened = True
+
```
kgiusti, line 455: shouldn't this check also ensure that "receiver_to_kill" link is also opened?  That ensures the "connection_to_kill" exists before the testing starts.

tests/system_tests_two_routers.py
```diff
@@ -470,19 +497,17 @@ def on_message(self, event):
                                 'adminStatus': 'deleted'}
                             request.reply_to = self.mgmt_receiver_1.remote_source.address
                             self.mgmt_sender.send(request)
+                            self.n_sent += 1
```
kgiusti, line 500: If you add the additional check to wait for "receiver_to_kill", then the test should immediately fail if this loop exits without finding the connection to delete.

tests/system_tests_two_routers.py
```diff
@@ -470,19 +497,17 @@ def on_message(self, event):
...
         elif event.receiver == self.mgmt_receiver_1:
+            self.n_received += 1
             if event.message.properties['statusDescription'] == 'OK' and event.message.body['adminStatus'] == 'deleted':
```
kgiusti, line 503: I would argue if this response is not OK and deleted the test should immediately fail.

tests/system_tests_two_routers.py
```diff
@@ -409,6 +409,14 @@ def __init__(self, address):
...
+        self.timer = None
+        self.n_sent = 0
+        self.n_received = 0
+        self.mgmt_receiver_link_opened = False
+        self.mgmt_receiver_1_link_opened = False
+        self.mgmt_receiver_2_link_opened = False
+        self.query_timer = None
 
     def on_start(self, event):
         self.timer = event.reactor.schedule(TIMEOUT, TestTimeout(self))
```
kgiusti, line 422: Sorry but github won't let me comment on line 438:
The timeout(self) method should not simply close self.mgmt_conn - it should call self.bail() with the given error message

tests/system_tests_two_routers.py
```diff
@@ -439,16 +447,35 @@ def bail(self, error):
```
kgiusti, line 447: bail() needs to cancel the query timer as well if it is not None

tests/system_tests_two_routers.py
```diff
@@ -470,19 +497,17 @@ def on_message(self, event):
...
+                self.query_timer = event.reactor.schedule(3.0, PollTimeout(self))
 
         elif event.receiver == self.mgmt_receiver_2:
+            self.n_received += 1
             attribute_names = event.message.body['attributeNames']
             property_index = attribute_names .index('properties')
-            identity_index = attribute_names .index('identity')
 
             for result in event.message.body['results']:
                 if result[property_index]:
```
kgiusti, line 513: If you're super paranoid (like me) and assume that CI will be sadistically slow (it will be), instead of calling bail() on line 516, you'd re-schedule the query_timer and try again.  If the connection never gets deleted then the test will fail when the TIMEOUT timer expires.

kgiusti (changes requested):
I realize some of my comments apply to pre-existing parts of that test, but you'll probably want to consider fixing them if you agree they are also potential race conditions.

kgiusti (approved):

PR #1429, opened 2021-11-06
DISPATCH-2277 Aligned memory allocation and freeing

jiridanek:
This is still messy, but it seems to work well on Windows. The main change is that ALIGNED allocations now require ALIGNED free. This is only checked on Windows, on other platforms it does not matter.

There used to be an inconsitency of allocating `tpool` both aligned and unaligned. I am changing this to be aligned always. This could possibly be fixed as its own separate bug...

### DIFF
--- a/include/qpid/dispatch/ctools.h
+++ b/include/qpid/dispatch/ctools.h
@@ -41,7 +41,7 @@
 //
 
 #ifdef __sun
-#define NEW_CACHE_ALIGNED(t,p)                  \
+#define NEW_CACHE_ALIGNED(t,p) \
 do { \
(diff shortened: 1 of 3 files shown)


kgiusti (changes requested):
+1 to the FREE_CACHE addition.

I'd remove the windows code - native windows is not a supported platform.  Until someone steps up to maintain a windows port I don't think we should head down that path.

Speaking personally I've don't have any windows systems that I can use to support this, and I'm not planning on getting any.

jiridanek:
> I'd remove the windows code

Ok, then, I'll go ahead merging it without the Windows part

PR #1430, opened 2021-11-09
Dispatch 2275

kgiusti:


tests/system_tests_link_routes.py
```diff
@@ -1797,110 +1814,227 @@ def _fake_broker(self, cls):
         self.router.wait_connectors()
         return fake_broker
 
+    def _find_frame(self, data, code) -> Optional[list]:
```
jiridanek, line 1817: This could be typed more. What is the type of `data` is quite obvious. I have little idea what is the type of the `code` parameter (I suspect `int`) and I am not sure list of _what_ is being returned. I'd venture a guess it is an `int` as well.

```suggestion
    def _find_frame(self, data: Data, code: int) -> Optional[List[int]]:
```

tests/system_tests_link_routes.py
```diff
@@ -1797,110 +1814,227 @@ def _fake_broker(self, cls):
...
+        """
+        frame.rewind()
+        fbytes = frame.encode()
+        flen = len(fbytes) + 8
+        # AMQP FRAME HEADER: 4 byte length, DOFF, TYPE, CHANNEL
+        sock.sendall(flen.to_bytes(4, "big"))
+        sock.sendall(bytes([2, 0, 0, 0]))
+        sock.sendall(fbytes)
+
+    def _construct_transfer(self, delivery_id, tag, more=False, add_ma=False,
```
jiridanek, line 1850: I can imagine that these functions will in the future need to be moved up the scope so that other tests can use them too. They look a bit complicated, maybe even enough to deserve their own test, maybe just a small one?

tests/system_tests_link_routes.py
```diff
@@ -1797,110 +1814,227 @@ def _fake_broker(self, cls):
...
+        flow_frame = self._find_frame(data[8:], self.FLOW_DESCRIPTOR)
+        self.assertIsNotNone(flow_frame)
+
+        # First send a message on link routed address "examples" with a small
+        # message body. Verify the the sent message has been accepted.
+        t1_frame = self._construct_transfer(0, b'\x01', add_ma=True, add_body=True)
+        self._send_frame(t1_frame, s)
+
+        # We expect to get a disposition frame that accepted the message
         sleep(0.5)
```
jiridanek, line 1977: Maybe use a retry function? Alan added some back in the day. They would take a lambda and retry until lambda returns True. Also, there was one that would return while an expected exception was being thrown. So maybe use it to repeat `self.assertIsNotNone(flow_frame)` in a loop while either AssertionError is being thrown, or 0.5 seconds is not yet up.

tests/system_tests_link_routes.py
```diff
@@ -1797,110 +1814,227 @@ def _fake_broker(self, cls):
...
         data = s.recv(1024)
-        # The delivery has been accepted.
-        self.assertIn("x00S$E", repr(data))
+        dispo_frame = self._find_frame(data, self.DISPO_DESCRIPTOR)
+        self.assertIsNotNone(dispo_frame, "expected a disposition (none arrived!)")
```
jiridanek, line 1980: This is fragile in case s.recv() receives say only half of the AMQP frame the first time it runs, and the other half the second time, and the expected data is found on the border of both frames. So the suggestion with the retry I tried to give above does not apply, it would make the test flaky here. (Well, the s.recv()s would have to be appending to a buffer, which gets checked in the entirety each time, I guess, so solutions exist.)

ganeshmurthy (approved):

jiridanek (approved):

PR #1435, opened 2021-11-12
DISPATCH-2267 - IO-thread facility to watch for changes to address reachability

ted-ross:
Added API to the core header file, implemented the feature, and added a test.

### DIFF
--- a/include/qpid/dispatch/router_core.h
+++ b/include/qpid/dispatch/router_core.h
@@ -68,6 +68,15 @@ void qdr_process_tick(qdr_core_t *core);
 
 
 /**
+ * Return true iff the test hooks option is enabled for this process.
+ * 
+ * @param core Pointer to the core object returned by qd_core()
+ * @return true iff test hooks are enabled
+ */
+bool qdr_core_test_hooks_enabled(const qdr_core_t *core);
+
+
+/**
  ******************************************************************************
  * Route table maintenance functions (Router Control)
  ******************************************************************************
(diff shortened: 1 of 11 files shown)


src/router_core/address_watch.c
```diff
@@ -0,0 +1,154 @@
...
+        watch = DEQ_NEXT(watch);
+    }
+}
+
+void qdr_address_watch_shutdown(qdr_core_t *core)
+{
+    qdr_address_watch_t *watch = DEQ_HEAD(core->addr_watches);
+    while (!!watch) {
+        DEQ_REMOVE(core->addr_watches, watch);
+        free(watch->address_hash);
```
ganeshmurthy, line 100: can we introduce a function called qdr_free_address_watch() or qdr_address_watch_free() where we free watch->address_hash and then call free_qdr_address_watch_t(watch); ? 
free_qdr_address_watch_t is called from two places and so the new function can be called instead of that.

ganeshmurthy (approved):

src/router_core/address_watch.c
ted-ross: This is a good idea.  I'll add it.

include/qpid/dispatch/router_core.h
```diff
@@ -160,6 +169,64 @@ void qdr_send_to2(qdr_core_t *core, qd_message_t *msg, const char *addr,
...
+                                          const char                 *address,
+                                          char                        aclass,
+                                          char                        phase,
+                                          qdr_address_watch_update_t  on_watch,
+                                          void                       *context);
+
+/**
+ * qdr_core_unwatch_address
+ * 
+ * Cancel an address watch subscription.  It is safe to invoke this function from an IO thread.
```
kgiusti, line 223: It is possible for the update handler to be invoked during or after this call on another I/O thread. That brings up a possible race:  whatever the context pointer addresses has to remain valid until we're sure that the handler will no longer be invoked, which isn't possible to tell from the caller's point of view.

src/router_core/address_watch.c
```diff
@@ -0,0 +1,160 @@
...
+
+
+static void qdr_watch_invoker(qdr_core_t *core, qdr_general_work_t *work)
+{
+    work->watch_handler(work->context,
+                        work->local_consumers, work->in_proc_consumers, work->remote_consumers, work->local_producers);
+}
+
+
+static void qdr_core_watch_address_CT(qdr_core_t *core, qdr_action_t *action, bool discard)
```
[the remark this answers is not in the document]
ted-ross: Yes, it should.  I'll make this change.

include/qpid/dispatch/router_core.h
ted-ross: Yes, this is a good catch.  It will require a bit of re-work, but I'll make it so that this cannot happen.

kgiusti (approved):
Approved modulo requested changes.

PR #1437, opened 2021-11-15
DISPATCH-2266: Increase container timeout to 3 seconds on AsyncTestRe…

ganeshmurthy:
…ceiver and AsyncTestSender

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -928,13 +928,19 @@ class AsyncTestReceiver(MessagingHandler):
     def _main(self):
         self._container.timeout = 0.5
         self._container.start()
-        self._logger.log("Starting reactor")
+        self._logger.log("AsyncTestReceiver Starting reactor")
         while self._container.process():
             if self._stop_thread:
                 if self._conn:
                     self._conn.close()
                     self._conn = None
-        self._logger.log("reactor thread done")
+        self._logger.log("AsyncTestReceiver reactor thread done")
+
+    def on_connection_error(self, event):
+        self._logger.log("AsyncTestReceiver on_connection_error=%s" % event.connection.remote_condition.description)
+
+    def on_link_error(self, event):
+        self._logger.log("AsyncTestReceiver on_link_error=%s" % event.link.remote_condition.description)
 
     def stop(self, timeout=TIMEOUT):
         self._stop_thread = True
@@ -1028,10 +1034,10 @@ class AsyncTestSender(MessagingHandler):
     def _main(self):
         self._container.timeout = 0.5
         self._container.start()
-        self._logger.log("Starting reactor")
+        self._logger.log("AsyncTestSender Starting reactor")
         while self._container.process():
             self._check_if_done()
-        self._logger.log("reactor thread done")
+        self._logger.log("AsyncTestSender reactor thread done")
 
     def get_msg_stats(self):
         return self.msg_stats % (self.sent, self.accepted, self.released, self.modified, self.rejected)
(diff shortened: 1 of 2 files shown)


tests/system_test.py
```diff
@@ -926,15 +926,15 @@ def get_queue_stats(self):
         return self.queue_stats % (self.num_queue_puts, self.num_queue_gets)
 
     def _main(self):
-        self._container.timeout = 0.5
+        self._container.timeout = 3.0
```
jiridanek, line 929: Maybe use a global constant for the three seconds? So that the timeouts can be unified across tests, based on what they "mean". Something like `DEFAULT_TEST_CLIENT_CONTAINER_TIMEOUT = 3`?

jiridanek: In addition to system_tests.TIMEOUT which is there already, https://github.com/apache/qpid-dispatch/blob/138d084d7a0d9dbeed2991affffc8f669563be68/tests/system_test.py#L124-L126

tests/system_tests_edge_router.py
```diff
@@ -1895,6 +1895,10 @@ def test_51_link_route_proxy_configured(self):
         if self.skip['test_51'] :
             self.skipTest("Test skipped during development.")
 
+        # The previous test waited for address CfgLinkRoute1 on router INT_B
+        # We will wait for address CfgLinkRoute1 to be unsubscribed
+        self.INT_B.wait_address_unsubscribed("CfgLinkRoute1")
```
jiridanek, line 1900: Depending on the specific order in which the tests are run sounds dangerous. I think that this line should be part of the previous test.

It is probably sufficient to put it at the end. That way it would be skipped if the test failed before it got to it, but that should be rare,...

There is currently not a good way to have a per-test tear-down in system-tests. One could be developed... Maybe a decorator on the test method? Or go all-in with pytest and use fixtures, which do have their teardowns? The least disruptive thing would be to have a list in the test class, `self.addresses_requiring_unsubscribe`, which would be cleaned in the shared tear_down code. Tests would populate it with addresses that require this treatment.

tests/system_tests_edge_router.py
```diff
@@ -1887,6 +1887,8 @@ def test_50_link_topology(self):
         bc_a.close()
         bc_b.close()
 
+        self.INT_B.wait_address_unsubscribed("CfgLinkRoute1")
```
kgiusti, line 1890: It would be better to do this check in test_51_link_route_proxy_configured to keep it all in one place and prevent accidental breakage if - for whatever reason - tests need to be re-ordered.  Eventually all tests that enable a link route(s) should perform that type of check at the start of the test.

kgiusti: Update: Jiri's approach is much cleaner - require all tests to verify that any addresses it introduced are removed from the router at the end of the individual test.  But it would need to be done to all applicable tests, which means... more work.

tests/system_tests_edge_router.py
jiridanek: Another way, maybe not quick, but certainly dirty, would be to require that each test suffixes all addresses it uses with it's own name. The name can be found by introspecting the Python stack, or maybe the unittest module already knows the name of the currently running test. Randomly generated string would work as well.

The problem is that these unique addresses cannot be preconfigured ar the beginning, each test has to create them through management.

tests/system_tests_edge_router.py
ganeshmurthy: For now, I am going to commit this fix to see what is happening on the main branch. If we see more of these kinds of failures, let's then think of an approach to solving the issue in a more generic fashion

PR #1441, opened 2021-11-16
DISPATCH-2258: Cleaned up log messages, created senders after receive…

ganeshmurthy:
…rs are created

### DIFF
--- a/tests/system_tests_distribution.py
+++ b/tests/system_tests_distribution.py
@@ -1635,8 +1635,8 @@ class DistributionTests (TestCase):
                                     self.A_addr,
                                     self.B_addr,
                                     self.C_route_container_addr,
-                                    self.waypoint_prefix_3 + '.waypoint'
-                                    )
+                                    self.waypoint_prefix_3 + '.waypoint',
+                                    debug=True)
         test.run()
         self.assertIsNone(test.error)
 
@@ -1951,9 +1951,9 @@ class LinkAttachRoutingCheckOnly (MessagingHandler):
 
         self.debug = False
 
-    def debug_print(self, message) :
-        if self.debug :
-            print(message)
+    def debug_print(self, message):
+        if self.debug:
+            print(message, flush=True)
 
     def timeout(self):
         self.bail("Timeout Expired")
@@ -2889,9 +2889,9 @@ class RoutingTest (MessagingHandler):
         self.status = 'start up'
         self.container_id = container_id
 
-    def debug_print(self, message) :
-        if self.debug :
-            print(message)
+    def debug_print(self, message):
+        if self.debug:
+            print(message, flush=True)
 
     # If this happens, the test is hanging.
     def timeout(self):
@@ -3112,7 +3112,7 @@ class RoutingTest (MessagingHandler):
     # =====================================================
     def check_receiver_distribution(self, expected_receiver_counts) :
         self.debug_print("check_receiver_distribution expecting: %s" % str(expected_receiver_counts))
-        if self.debug :
+        if self.debug:
             self.print_receiver_distribution()
         for router in range(len(self.router_cnx_counts)) :
             cnx_dict = self.router_cnx_counts[router]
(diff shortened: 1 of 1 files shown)


ganeshmurthy:
This PR enables debug logging so we can figure out what is going on in the Travis environment when this test fails.

tests/system_tests_distribution.py
```diff
@@ -3826,6 +3823,7 @@ def __init__(self,
         self.error                = None
         self.messages_per_sender  = 100
```
kgiusti, line 3824: Non-patch observation: is 100 messages *really* necessary?  CI systems tend to be painfully slow.  Can the same degree of confidence that this works be achieved by say 20 messages?

tests/system_tests_distribution.py
```diff
@@ -3884,7 +3882,7 @@ def __init__(self,
         n_links_per_message = 4
         self.n_expected_transitions = len(self.senders) * self.messages_per_sender * n_links_per_message
 
-        self.debug = False
+        self.debug = debug
 
         self.test_name = test_name
```
kgiusti, line 3888: Since github won't let me comment on lines that haven't been changed: this is regards to the debug_print() function: add a flush=True to the print statement:  print(message, flush=True).   unittest has a tendency to not flush stdout if a failure is hit.

tests/system_tests_distribution.py
```diff
@@ -3910,11 +3908,9 @@ def send_from_client(self, sender, n_messages, sender_index):
             n_sent             += 1
             self.n_sent        += 1
             self.n_transitions += 1
-            self.debug_print("send_from_client -- sender: %d n_sent: %d" % (sender_index, n_sent))
+        self.debug_print("send_from_client -- sender: %d n_sent: %d" % (sender_index, n_sent))
```
kgiusti, line 3911: Did you also want to do the same thing (move the debug print out of the loop) for the send_from_waypoint() method below?  Just asking...

tests/system_tests_distribution.py
```diff
@@ -3972,19 +3958,25 @@ def on_link_opening(self, event):
...
     def on_sendable(self, event):
-        self.debug_print("on_sendable ------------------------------")
         for index in range(len(self.senders)) :
             sender = self.senders[index]
             if event.sender == sender['sender'] :
-                self.debug_print("    client sender %d" % index)
                 if sender['n_sent'] < sender['to_send'] :
                     self.debug_print("    sending %d" % sender['to_send'])
                     self.send_from_client(sender['sender'], sender['to_send'], index)
                     sender['n_sent'] = sender['to_send']  # n_sent = n_to_send
```
kgiusti, line 3979: wait a sec... can't send_from_client exit with < to_send if credit runs out?   This seems wrong.  Maybe send_from_client should be incrementing sender['n_sent'] rather than having this line assume sender['to_send'] was actually sent.

And don't get me started on why the heck this test is using maps (sender['n_sent']) rather than a class (sender.n_sent) all over this test.  It's like this test was purposely written to be easy to break and maximize reader confusion.

kgiusti (changes requested):
I give up - this test is a mess and no amount of careful debug printing is going to fix it.
Either disable it or re-write it, and if rewriting it please use the FakeBroker for the waypoint - that's why it was written in the first place.

tests/system_tests_distribution.py
ganeshmurthy: I am curious to find out why this test is failing. I am sure the failure has nothing to do with the number of messages sent. I want to push this logging change to main and analyze the debug output when the test fails.

tests/system_tests_distribution.py
ganeshmurthy: Actually, no. The loop in send_from_waypoint prints the message_content_number, so I want to keep the print inside the loop

tests/system_tests_distribution.py
ganeshmurthy: If send_from_client() exited because of lack of credit, the debug_print we added will print the n_sent. This additional logging will tell us if the client could not send the desired number of messages due to the lack of credit

kgiusti (approved):

PR #1449, opened 2021-11-20
NO-JIRA Consistently use Python 3 in all Travis CI jobs

jiridanek:


jiridanek:
Travis issue reported in https://issues.apache.org/jira/browse/DISPATCH-2292. Let's see how frequent it is going to be.

PR only fixes Travis CI, so it couldn't be responsible for the fail. Will merge.

PR #1451, opened 2021-11-23
DISPATCH-2221 Proposal for automated e2e webconsole test in MS Playwright

jiridanek:
Hi, I've put together the minimal amount of things that shows how this could look like

https://www.youtube.com/watch?v=NM_C899M5MQ

The video shows a sample test (which opens the Playwright homepage) and then a trivial assert-nothing test that clicks through the main sections of the dispatch console one by one.

My goal is to have all the major parts in the PR, while not spending too much time on the tests and through that getting attached to something that has not yet been approved. I am hoping that Playwright will be acceptable automation technology, given its characteristics (see e.g. https://playwright.dev/docs/browsers/). I feel that the JavaScript/TypeScript API is the most appropriate one, given that console itself is written in that, and the console developers might not necessarily be too friendly with Python.

One (minor) thing I haven't figured out is how to use Yarn 2 the right way. Other aspects, such as installing the headless browser, starting dispatch, and running the test in GitHub Actions, are there. One thing missing is storing artifacts on failure. Playwright can take browser event dumps for replay, and also record videos if a test failed.

Next up, we have to agree on the test framework (see the Jira, where I considered Selenium, Playwright and Cypress) and decide what kind of tests we need. I am thinking for every CRUD form, try to create/delete/update router entities.

CC @fgiorgetti @granzoto Thanks!


jiridanek:
In a subsequent discussion it was decided that we want to test websites with JUnit + Selenium instead of this.

PR #1452, opened 2021-12-01
Dispatch-1403/1487  Message Annotations refactor part 1

kgiusti:
On my laptop this change improves small message throughput in a two-hop deployment by 10-12% over baseline.


src/buffer_field_api.h
```diff
@@ -0,0 +1,311 @@
...
+#include "qpid/dispatch/iterator.h"
+
+/* qd_buffer_field_memcpy
+ *
+ * Copy up to n octets from bfield to dest, advance bfield by the number of
+ * octets copied
+ *
+ * @return total of octets copied - may be < n if len(bfield) < n
+ */
+static inline size_t qd_buffer_field_memcpy(qd_buffer_field_t *bfield, uint8_t *dest, size_t n)
```
ganeshmurthy, line 38: can we call this function qd_buffer_field_nmemcpy()

src/buffer_field_api.h
```diff
@@ -0,0 +1,311 @@
...
+{
```
ganeshmurthy, line 39: `if (!bfield)
     return 0;`

src/buffer_field_api.h
kgiusti: I'll change it to qd_buffer_field_ncopy() - that's consistent with similar APIs in the code.

src/buffer_field_api.h
kgiusti: Would you mind if instead we treated calling this with a null bfield as an error?  I'd like to keep these functions as fast as possible due to their use for inbound message parsing.
The proton API has done this as well by adding "assert(required-non-0-pointer)" to those APIs which expect the caller has already sanitized the parameters.  example: https://github.com/apache/qpid-proton/blob/main/c/src/core/engine.c#L99

ganeshmurthy: > Would you mind if instead we treated calling this with a null bfield as an error?
Sounds good to me.

src/buffer_field_api.h
```diff
@@ -0,0 +1,311 @@
...
+ *
+ * Move the cursor of bfield forward by amount octets.
+ *
+ * @return total of octets skipped - may be < amount if len(bfield) < amount
+ */
+static inline size_t qd_buffer_field_advance(qd_buffer_field_t *bfield, size_t amount)
+{
+    size_t blen = bfield->length;
+    size_t count = MIN(amount, blen);
+    if (bfield->buffer) {
```
ganeshmurthy, line 95: Can we assume that all bfields MUST have an associated buffer. It is after all a qd_buffer_field. Don't use this data structure if you cannot associate this bfield with a buffer

src/buffer_field_api.h
```diff
@@ -0,0 +1,311 @@
...
+                    bfield->length = 0;
+                    count = 0;
+                    assert(false);  // TODO(KAG): is this fixed?
+                }
+            }
+        } while (count);
+
+        return dest - start;
+
+    } else {    // string/binary data
```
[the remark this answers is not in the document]
kgiusti: Here's the problem (which is totally non-obvious) - qd_iterator_t (from which the qd_buffer_field_t api was copied) supports buffer fields that don't actually have an associated qd_buffer_list_t!  When an iterator is created via qd_iterator_string() or qd_iterator_binary() there's no qd_buffer_t involved at all.

I didn't realize that until parse tests started failing since the code was attempting to parse out an iterator that was created on a string/binary non-buffer.  So I had to carry this over to minimize the patch.

I agree this is non-obvious and messy.

To remove the non-buffer code from the qd_buffer_field_t will require modifying parse() to check the iterator parameter for a buffer.  If not present parse() _could_ copy the iterator's content into a temp buffer list.  That may not be as bad as it sounds if it only occurs for non-message parsing - I'll take a look at that.

src/buffer_field_api.h
kgiusti: I'd like that.

kgiusti:
Going to re-submit these patches separately - closing for now.

src/parse.c
```diff
@@ -23,27 +23,37 @@
...
+#include "buffer_field_api.h"
+
 #include <assert.h>
 #include <inttypes.h>
 #include <stdio.h>
 
 DEQ_DECLARE(qd_parsed_field_t, qd_parsed_field_list_t);
 
+
+typedef struct qd_amqp_field_t {
```
ganeshmurthy, line 35: I am thinking aloud here but can qd_amqp_field_t and qd_buffer_field_t just be combined into just a qd_amqp_field_t ?

PR #1457, opened 2021-12-13
DISPATH-2298: Avoid rescheduling qd_timer if we are already servicing timer

astitcher:
If we are in the middle of running through the list of timers in
qd_timer_visit() we don't need to wake the timer as anything we add to
the list will get serviced immediately in any case.

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -249,7 +249,11 @@ void qd_timer_schedule(qd_timer_t *timer, qd_duration_t duration)
     }
 
     qd_timer_t *first = DEQ_HEAD(scheduled_timers);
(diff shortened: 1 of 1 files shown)


ganeshmurthy (approved):

kgiusti (approved):
A nice clean optimization - well done!
Please fix typo in commit log "DISPATH" before commit.

Let's hold off committing this for now - until we stabilize the CI a bit.  I don't think this patch is the cause of the failures on CI btw.

astitcher:
Fixed the commit message typo

astitcher:
Rebase and merged to main

PR #1458, opened 2021-12-14
DISPATCH-1487: inter-router annotations re-factor

kgiusti:


kgiusti:
Nevermind - will re-submit with proper fixes.

PR #1463, opened 2021-12-16
DISPATCH-2300: Fixed test to not look for contents of transfer messag…

ganeshmurthy:
…e. Also a minor fix to the connection creation code

### DIFF
--- a/tests/system_tests_log_level_update.py
+++ b/tests/system_tests_log_level_update.py
@@ -254,12 +254,21 @@ class EnableConnectionLevelInterRouterTraceTest(TestCase):
 
         cls.address = cls.routers[1].addresses[0]
 
+    def _get_transfer_frame_count(self, conn_id):
+        inter_router_cid = "[C" + conn_id + "]"
+        num_transfers = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            for log_line in router_log:
+                log_components = log_line.split(" ")
+                if len(log_components) > 8 and 'PROTOCOL' in log_components[3]:
+                    if inter_router_cid in log_components[5] and '@transfer' in log_components[8]:
+                        num_transfers += 1
+        return num_transfers
+
     def test_inter_router_protocol_trace(self):
         qd_manager = QdManager(self, self.address)
-        # Turn off trace logging on all connections for Router B.
-        qd_manager.update("org.apache.qpid.dispatch.log", {"enable": "info+"},
-                          name="log/DEFAULT")
 
+        # The router already has trace logging turned on for all connections.
         # Get the connection id of the inter-router connection
         results = qd_manager.query("org.apache.qpid.dispatch.connection")
         conn_id = None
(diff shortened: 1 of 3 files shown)


tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,69 @@ def router(name, connection):
...
         for result in results:
             if result['role'] == 'inter-router':
                 conn_id = result['identity']
 
-        # Turn on trace logging for the inter-router connection
-        qd_manager.update("org.apache.qpid.dispatch.connection", {"enableProtocolTrace": "true"}, identity=conn_id)
+        # Turn off trace logging for the inter-router connection. This update command is run async by the router
+        # so we need to sleep a bit before the operation is actually completed.
+        qd_manager.update("org.apache.qpid.dispatch.connection", {"enableProtocolTrace": "false"}, identity=conn_id)
+        time.sleep(1)
```
jiridanek, line 271: Is there a way to either wait for completion of the async operation or possibly repeatedly query for the current state until the change happens? Just asking.

tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,69 @@ def router(name, connection):
...
 
-        # Create a receiver and make sure the MAU update is seen on the inter-router connection log
+        inter_router_cid = "[C" + conn_id + "]"
+
+        num_transfers = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            log_lines = router_log.read().split("\n")
```
jiridanek, line 277:
```suggestion
            log_lines = router_log.readlines()
```

?

jiridanek:
```suggestion
            for log_line in router_log:
```

Actually, this is what you want. Open file is an iterator over its lines, in Python.

tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,69 @@ def router(name, connection):
...
-        logs = qd_manager.get_log()
-        mau_found = False
-        for log in logs:
-            if 'PROTOCOL' in log[0]:
-                if "@transfer" in log[2] and TEST_ADDR_1 in log[2] and "MAU" in log[2]:
-                    mau_found = True
-                    break
+        num_transfers_after_update = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            log_lines = router_log.read().split("\n")
```
jiridanek, line 291: same as before, this helper variable is not necessary

tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,69 @@ def router(name, connection):
...
-        logs = qd_manager.get_log()
-        mau_found = False
-        for log in logs:
-            if 'PROTOCOL' in log[0]:
-                if "@transfer" in log[2] and TEST_ADDR_2 in log[2] and "MAU" in log[2]:
-                    mau_found = True
-                    break
+        num_transfers_after_update = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            log_lines = router_log.read().split("\n")
```
jiridanek, line 313: and ditto here, if you decide to try the iterator-over-lines

tests/system_tests_log_level_update.py
ganeshmurthy: Unfortunately  there is not. We could query the connection again to find out but even if that has the correct value of enableProtocolTrace, it does not guarantee that the action actually has executed. See here - https://github.com/apache/qpid-dispatch/blob/main/src/router_core/agent_connection.c#L644

tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,72 @@ def router(name, connection):
...
-        for log in logs:
-            if 'PROTOCOL' in log[0]:
-                if "@transfer" in log[2] and TEST_ADDR_2 in log[2] and "MAU" in log[2]:
-                    mau_found = True
-                    break
+        num_transfers_after_update = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            for log_line in router_log:
+                log_components = log_line.split(" ")
+                if len(log_components) > 2 and 'PROTOCOL' in log_components[3]:
```
kgiusti, line 317: and here as well (maybe make the count code a private method to this test so you don't have to copy & pasta it multiple times).

tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,72 @@ def router(name, connection):
...
-                if "@transfer" in log[2] and TEST_ADDR_1 in log[2] and "MAU" in log[2]:
-                    mau_found = True
-                    break
-
-        self.assertTrue(mau_found)
+        num_transfers_after_update = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            for log_line in router_log:
+                log_components = log_line.split(" ")
+                if len(log_components) > 2 and 'PROTOCOL' in log_components[3]:
```
kgiusti, line 294: ditto len(log_components) here as well.

tests/system_tests_log_level_update.py
```diff
@@ -256,56 +256,72 @@ def router(name, connection):
...
+        time.sleep(1)
 
-        # Create a receiver and make sure the MAU update is seen on the inter-router connection log
+        inter_router_cid = "[C" + conn_id + "]"
+
+        num_transfers = 0
+        with open(self.routers[1].logfile_path) as router_log:
+            for log_line in router_log:
+                log_components = log_line.split(" ")
+                if len(log_components) > 2 and 'PROTOCOL' in log_components[3]:
```
kgiusti, line 279: I think you meant to check len(log_components) > 8, so indexing log_components[8] won't barf?  or at minimum > 3 if you're willing to trust the format of the PROTOCOL log statement.

kgiusti (changes requested):
See inline

tests/system_tests_log_level_update.py
kgiusti: Uh, apparently my comments are in reverse order from above (?)

kgiusti (approved):

PR #1465, opened 2021-12-16
DISPATCH-1403: define a common lightweight buffer field API

kgiusti:


include/qpid/dispatch/buffer_field.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <stdbool.h>
+
+/* descriptor for a sequence of bytes in a buffer list
+ */
+typedef struct qd_buffer_field_t qd_buffer_field_t;
+struct qd_buffer_field_t {
+    qd_buffer_t   *buffer;  // hold start of data
+    const uint8_t *cursor;  // first octet of data
+    size_t         length;  // of data
```
ganeshmurthy, line 39: Should length be called remaining_length ? The reason I ask is that the length field continue to decrease as we move thru the buffer and *length* sounds like something that does not change.

kgiusti: Yes - the original source for the buffer field code is a set of file-private inline functions in interator.c; that used the term "remaining" instead of "length".  No reason to change it - let's go back to "remaining".

ganeshmurthy (approved):

kgiusti:
Closed.  This patch is included as a commit in https://github.com/apache/qpid-dispatch/pull/1515 instead.

PR #1475, opened 2022-01-07
DISPATCH-2305: prevent policy defaults overriding configuration

kgiusti:


python/qpid_dispatch/management/qdrouter.json
```diff
@@ -2358,21 +2358,18 @@
                 "maxFrameSize": {
                     "type": "integer",
                     "description": "The largest frame, in bytes, that may be sent on this connection. Non-zero policy values overwrite values specified for a listener object (AMQP Open, max-frame-size).",
-                    "default": 16384,
```
ganeshmurthy, line 2361: Bear with me here but when I look at this from a layman's perspective, I like seeing the defaults in the json schema and this PR seems to remove the defaults from the schema. Is there a way that these defaults can be still there and if the policy does not overide these values, the defaults can be used? Is that too much work from the policy side ?

kgiusti: Why have two sets of defaults then (one in the entity's definition in the schema [e.g. listener], and one in the policy)?   My point is defining a policy should not force another set of defaults when the entity itself defines its defaults.

Otherwise we're maintaining two sets of defaults which need to be kept in sync else we trigger this bug.

ganeshmurthy: Good point. I did notice that maxSessionWindow is defaulted to 1638400 in vhostUserGroupSettings and is not "directly" specified anywhere else.

ganeshmurthy (approved):

ted-ross (approved):

PR #1476, opened 2022-01-10
DISPATCH-1974 flake8 python-checker checks and fixes

jiridanek:


jiridanek:
@ganeshmurthy, @kgiusti I think this is overall desirable, but there are some caveats that I'll have to fix. First, the flake8 and pylint things should each have separate PR.

Then, some of the flake8 warnings only appear if you have extra flake8 modules installed in your python environment, so I'm actually surprising things that flake8 in tox will not report (because I was running flake8 installed by yum. Meaning the docstring warnings do not actually appear. There the question is, do we want them? Flake8 lines e.g. one-line docstrings if they can be put on one line. That is imo good check, makes docstrings more compact. Enforcing quotes instead of apostrophes in docstrings is also helpful, imo. The other checks, not so much useful.

kgiusti:
> @ganeshmurthy, @kgiusti I think this is desirable overall, but there are some caveats that I'll have to fix.
> 
> First, the flake8 and pylint things should each have separate PR.
> 
> Then, some of the flake8 warnings only appear if you have extra flake8 modules installed in your python environment, so I'm actually suppressing things that flake8 in tox will not report (because I was running flake8 installed by yum. Meaning the docstring warnings do not actually appear. There the question is, do we want them? Flake8 likes e.g. one-line docstrings if they can be put on one line. That is imo good check, makes docstrings more compact. Enforcing quotes instead of apostrophes in docstrings is also helpful, imo. The other checks, not so much useful and actually often quite annoying.

Agreed on all points!

jiridanek:
I've merged the pylint addition. I probably will not do anything about flake8 doc comment stuff because it just introduces large diffs to commit history, which is undesirable at this point.

Maybe later.

PR #1498, opened 2022-01-26
DISPATCH-2318: fix double-free of subscription on shutdown

kgiusti:


src/router_core/route_tables.c
```diff
@@ -671,8 +671,7 @@ static void qdr_subscribe_CT(qdr_core_t *core, qdr_action_t *action, bool discar
             DEQ_INSERT_TAIL(addr->subscriptions, sub);
             qdr_addr_start_inlinks_CT(core, addr);
         }
-    } else
-        free(sub);
+    }
```
jiridanek, line 674: I was not sure if this is safe thing to do, but it probably is: if this causes problems (=leak), that would mean there is bug somewhere else (maybe causing the "unsubscribe" action not to be scheduled).

ganeshmurthy (approved):

PR #1052, opened 2021-02-20
DISPATCH-848, DISPATCH-1962 Fix leak of IoAdapter_init

jiridanek:
This PR depends on PRs to fix leaks in `qdr_subscription_t` (https://github.com/apache/qpid-dispatch/pull/1051) and `qdr_core_t` (https://github.com/apache/qpid-dispatch/pull/1049). The former is still WIP, so I won't carry the commit here, because it would make review harder.

The original problem is

```
9: Direct leak of 56 byte(s) in 1 object(s) allocated from:
9:     #0 0x7f78a3606e8f in __interceptor_malloc (/nix/store/g40sl3zh3nv52vj0mrl4iki5iphh5ika-gcc-10.2.0-lib/lib/libasan.so.6+0xace8f)
9:     #1 0x7f78a2d64afb in qd_malloc ../include/qpid/dispatch/ctools.h:229
9:     #2 0x7f78a2d657da in qdr_core_subscribe ../src/router_core/route_tables.c:149
9:     #3 0x7f78a2c83072 in IoAdapter_init ../src/python_embedded.c:711
9:     #4 0x7f78a2353a6c in type_call (/nix/store/r85nxfnwiv45nbmf5yb60jj8ajim4m7w-python3-3.8.5/lib/libpython3.8.so.1.0+0x165a6c)
```

The solution consists of
1) implementing support for GC in IoAdapter so that Python can free the cycle described in the Jira.
2) Fixing Decreffing so that Python does free
3) Calling PyGC_Collect() at an opportune place where finalizers can be called without accessing previously already freed memory.

PR #1499, opened 2022-01-27
DISPATCH-2144 Use python3-dbg in the Debug GHA build

jiridanek:


jiridanek:
@kgiusti This won't work as of yet, for several reasons

1. https://github.com/apache/qpid-dispatch/pull/1052 has to be fixed, but that PR itself is broken, because I can't do proper shutdown, which is hopefully addressed in followup work to https://github.com/skupperproject/skupper-router/pull/50
2. pytest released a bug in 7.0.0, which causes the `INTERNALERROR>` stacktraces, see https://github.com/skupperproject/skupper-router/pull/49 for link to pytest bug and attempt at workarounding that

I think that is all the reasons. I'll add more if I can think of something.

PR #1500, opened 2022-01-27
DISPATCH-2126: mask TSAN warning on delivery settled flag

kgiusti:


ganeshmurthy (approved):

tests/tsan.supp
```diff
@@ -29,6 +29,7 @@ race:qdr_link_stalled_outbound
 
 # DISPATCH-2126
 race:qdr_delivery_mcast_inbound_update_CT
+race:qdr_delivery_mcast_outbound_settled_CT
```
jiridanek, line 32: The entries are regexes, so they would also match any method that contains this as a substring.

`race:^qdr_delivery_mcast_outbound_settled_CT$` is imo preferable.

I see that the entries in the file above and bellow don't use the regex anchors, so it's probably ok to leave them out here as well

PR #1506, opened 2022-01-30
DISPATCH-2323 Try binding to a port before handing it out as "free"

jiridanek:
This has three more-or-less independent commits. I did not understand the full problem at first, so I tried to get the ports in a better way. Which I still think is worthwhile to do, but it does not fix all the fails.

Second commit avoids the daemon threads and fudge sleeps, which only obfuscate the problems and don't really do all that much good.

Third commit adds log dump to Git Hub Actions (or whenever the tests are being run by pytest). See for example https://github.com/apache/qpid-dispatch/runs/4998230289?check_suite_focus=true#step:27:703 to see what it looks like when test fails. There are logging messages from proton and from the test itself.

### DIFF
--- a/tests/http1_tests.py
+++ b/tests/http1_tests.py
@@ -77,8 +77,8 @@ class RequestHandler(BaseHTTPRequestHandler):
             self.end_headers()
             self.wfile.write(b'Server Closed')
             self.wfile.flush()
-            self.close_connection = True
-            self.server.server_killed = True
+            self.close_connection = True  # server will close connection from router
+            self.server.server_killed = True  # server will not accept a reconnect
             return
         self._execute_request(self.server.system_tests["POST"])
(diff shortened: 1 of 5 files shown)


kgiusti (approved):
Good improvement - these changes will need to be ported to skupper as well.

jiridanek:
I tried posting a SO question about ungraceful shutdown of the `http.server` in Python. There was no reply, but I believe it can be implemented the way (linked) Go server does it: keep track of incoming sockets and `.shutdown()` them when quitting. That should unblock any ongoing request processing stuck on `read()` and let it to finish.

https://stackoverflow.com/questions/70916335/how-do-i-forcibly-disconnect-all-currently-connected-clients-to-my-tcp-or-http-s

PR #1511, opened 2022-02-07
DISPATCH-2310: limit router identifiers to 255 characters

kgiusti:
This pulls over Ted's router name fix from skupper-router

### DIFF
--- a/include/qpid/dispatch/router.h
+++ b/include/qpid/dispatch/router.h
@@ -33,6 +33,8 @@
 
 #include <stdbool.h>
 
+#define QD_ROUTER_ID_MAX 127  // max length of router id in chars
+
 typedef struct qdr_core_t   qdr_core_t;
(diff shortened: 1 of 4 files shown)


ganeshmurthy (approved):

include/qpid/dispatch/router.h
```diff
@@ -33,6 +33,8 @@
 
 #include <stdbool.h>
 
+#define QD_ROUTER_ID_MAX 127  // max length of router id in chars
```
jiridanek, line 36: It's inconvenient that there is a C version and Python version of the constant. Is the #define actually used anywhere? I can't find any usage in this PR

jiridanek (commented):
Feel free to ignore the Python suggestion; it's coming too late in the review cycle and it is a small nit anyways

PR #1512, opened 2022-02-10
DISPATCH-2327: clean up link routes at end of tests

kgiusti:


tests/system_tests_multi_tenancy.py
```diff
@@ -683,12 +704,8 @@ def __init__(self, first_host, second_host, first_address, second_address, dynam
...
 
     def timeout(self):
-        self.error = "Timeout Expired: n_sent=%d n_rcvd=%d n_settled=%d" % (self.n_sent, self.n_rcvd, self.n_settled)
-        self.first_conn.close()
-        self.second_conn.close()
-        self.lookup_conn.close()
-        if self.poll_timer:
-            self.poll_timer.cancel()
+        self.fail("Timeout Expired: n_sent=%d n_rcvd=%d n_settled=%d" %
+                  (self.n_sent, self.n_rcvd, self.n_settled))
```
jiridanek, line 708: All the other tests scrupulously avoid throwing in the Proton handlers. I presume there was/is a good reason for it. (Other tests have the `error` field, and always do `self.assertIsNone(test.error)` at the end). So, is throwing actually alright?

tests/system_tests_multi_tenancy.py
```diff
@@ -229,6 +231,17 @@ def test_16_two_router_anonymous_sender_tenant_on_both(self):
         test.run()
         self.assertIsNone(test.error)
 
+    def _cleanup_link_routes(func):
```
jiridanek, line 234: The 'traditional' way of using `unittest` would be to do your `setUp()` and `tearDown()` in these two methods. (With classes being understood as a way to group together tests that share setup and teardown.)

I'd probably try to put the teardown into `LinkRouteTest` itself... but that does not have access to `self.routers`, so it is not such a good idea. Maybe add method `def tearDown(routers)` to it and call it in every test...

Decorators are too much magic for my simple tastes, I guess.

It's a matter of preference. The decorator is not too hard to understand... And if I had unlimited time, enthusiasm, and money to redesign this test "the best way", I would not actually know what to do with it right away... So, lgtm, just go ahead, I guess?

tests/system_tests_multi_tenancy.py
jiridanek: @kgiusti Actually, you don't even have `self.fail` available here! This is instance of `proton.handlers.MessagingHandler`, not `unittest.TestCase`. (At least, If I look right).

kgiusti: Yeah - not the best choice of method name, but it's local to the LinkRouteTest class itself.  I've changed it from "fail" to "done" so it won't be as confusing.

tests/system_tests_multi_tenancy.py
jiridanek: > Yeah in the end I'm punting on the decorator approach: too many issues with the python linters.

If the only problem is the linter, that should not be hard to put right. Put the decorator as a standalone top level method in the module, and the warning you got should be gone! You were not relying on it being an instance method anyway.

```python
def _cleanup_link_routes(func):
    @functools.wraps(func)
    def wrapper(self, *args, **kwargs):
        result = func(self, *args, **kwargs)
        # TODO something here
        return result
    return wrapper

class TestClass(unittest.TestCae):
    def __init__(self, a=1, b=2):
    @decorate
    def test_something(self):
        pass
```

I'm not sure what other problems there might be with it. I'd have to actually try it, and I don't want to poke into tests that I don't understand.

PR #1513, opened 2022-02-10
DISPATCH-2327: use decorator to clean up link routes at end of tests

jiridanek:
I like to think I could do something much nicer and more wow with unlimited money and enthusiasm, but this is what I got right now

```diff
> The decorator shall not fall into ruins
> despite the pain of loss upon all of us
> In our hearts of hearts the fallen leader
> is mightier than when alive
```

Watching CI, hoping I did not screw up too badly.

### DIFF
--- a/tests/system_tests_multi_tenancy.py
+++ b/tests/system_tests_multi_tenancy.py
@@ -17,6 +17,8 @@
 # under the License.
 #
 
+import functools
+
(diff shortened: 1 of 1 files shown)


jiridanek:
Ok, older Pythons don't like the `@staticmethod`, apparently. Very sad.

PR #1517, opened 2022-02-13
DISPATCH-2234: Update JavaScript console packages for the 1.19.0 release

jiridanek:
* Error: A `<Route>` is only ever to be used as the child of `<Routes>` element, never rendered directly. Please wrap your `<Route>` in a `<Routes>`.

* Error: You cannot render a `<Router>` inside another `<Router>`. You should never have more than one in your app.

See https://reactrouter.com/docs/en/v6/upgrading/v5, and therein linked https://gist.github.com/mjackson/d54b40a094277b7afdd6b81f51a0393f

* BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
* Module not found: Error: Can't resolve 'os' in '/home/jdanek/repos/qpid/qpid-dispatch/console/react/node_modules/rhea/lib'
* https://github.com/facebook/create-react-app/issues/11756
* filled https://github.com/amqp/rhea/issues/379


console/react/src/App.js
```diff
@@ -28,26 +29,22 @@ import "@patternfly/patternfly/components/Nav/nav.css";
...
+function App(props) {
+
+  // service is passed in to make testing easier
+  let service = new QDRService();
+
+  // also, a router is used here to provide PageLayout with a history property
+  let history = createHashHistory({ window });
+
+  return (
+    <HistoryRouter history={history}>
```
bartoval, line 42: Why we are using HistoryRouter here? Before this refactoring, we used the HashRouter.

console/react/package.json
```diff
@@ -1,25 +1,25 @@
...
+    "express": "^4.17.2",
     "font-awesome": "^4.7.0",
-    "prop-types": "^15.7.2",
+    "prop-types": "^15.8.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-fontawesome": "^1.7.1",
-    "react-router-dom": "^5.2.0",
-    "rhea": "^2.0.4",
+    "react-router-dom": "^6.2.1",
```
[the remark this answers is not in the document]
jiridanek: I think so, but there is not a thoughtfully developed policy on this in the dispatch-router project.

Normally, for the native code components (the qdrouterd itself), the versions of dependencies are dictated by the minimum version present in Linux distributions that needs to be supported; that is, centos/rhel/fedora and ubuntu. This is how C projects usually work. There are some packaging systems for C, similar to npm, but the traditional Linux practice is to rely on your distro.

In Javascript, all dependencies come from npm, so there the only limitation coming from distros is version of nodejs.

I believe that the most sensible policy is to keep up with the latest upstream version and to upgrade often. Here are my reasons for that

- security: minimizes the amount of `npm audit` warnings
- ease of update: when required (due to npm audit warnings), it helps if we are on still actively supported recent version
- small diffs on upgrade: when the upgrade is done soon, the amount of changes will be smaller than when having to upgrade over multiple major versions

console/react/src/App.js
```diff
@@ -28,26 +29,22 @@ import "@patternfly/patternfly/components/Nav/nav.css";
...
-          />
-        </div>
-      </Router>
-    );
-  }
+
+function App(props) {
+
+  // service is passed in to make testing easier
+  let service = new QDRService();
```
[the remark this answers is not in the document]
jiridanek: :+1:

console/react/src/overview/dashboard/layout.test.js
```diff
@@ -21,6 +21,7 @@ import React from "react";
 import { render } from "@testing-library/react";
 import { service } from "../../serviceTest";
 import PageLayout from "./layout";
+import {MemoryRouter} from "react-router-dom";
```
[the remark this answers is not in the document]
jiridanek: `MemoryRouter` is intended for testing in nodejs (outside of browser). `HistoryRouter` is meant to allow injecting your own `history` instance. (https://reactrouter.com/docs/en/v6/api#setup)

console/react/src/App.js
jiridanek: Because the code of some components nested further in needs to have `history` instance in properties. That is not available in react-router-dom v6 unless I put it there. And `HistoryRouter` is the way to inject `history`.

bartoval: Here I see that the child component PageLayout pass explicitly the pro ```history={history}``` and history coming from ``` createHashHistory({ window });```

jiridanek: Yeah, that is how you are supposed to use HistoryRouter, according to docs, https://reactrouter.com/docs/en/v6/api#unstable_historyrouter

Honestly, there are two things in the console, related to react-router-dom, that I think can be now done better

- navigating between pages; which is now done inconsistently, either with `<Navigate>`, or `navigate()`, or through `history`.
- the page structure which is supposed to use the `<Outlet>` feature but it does not

bartoval: Good point +1.

Probably one of the things to do is refactor the PageLayout component as a function (but this is out of the scope of this PR honestly) to use directly hooks. So we can avoid passing history and use directly useNavigate and useLocation. Then probably we can avoid passing {...props } because without typescript we lose a bit the control of what we inject in the component. Moreover the High order component 'WithRoute'  will be deleted

bartoval:
Great job +1 !

bartoval (approved):

console/react/src/App.js
jiridanek: Noted in https://issues.apache.org/jira/browse/DISPATCH-2331

PR #1550, opened 2022-04-06
DISPATCH-2338 - dockerfiles: Introduce Debian-based slim image w/ CI tests

Jmennius:
Use Debian stable release slim image with multistage build to
keep image size reasonable (143MB compared to 1.39GB Ubuntu based image).

Other changes compared to Ubuntu based image:
* ~~libuv is enabled~~
* a bit more careful work with dependencies
* allow building web console
* faster builds with Ninja
* build Proton from release, not master; configurable
* build Dispatch Router from checked out source
* minor changes in build process, Dockerfile syntax, etc

TODO:
- [x] build image on CI

---
It's not the prettiest in some ways.. but what can I do :)

~~I did not run any real workloads with it yet, but the router daemon starts and qdmanage can query from it.~~ tested

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -630,3 +630,15 @@ jobs:
       - name: Output
         if: ${{ ! cancelled() }}
         run: cat target/rat.txt || echo "Unable to print output"
+
+  container-images:
+    name: Container Images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Build Debian image
+        uses: docker/build-push-action@v2
+        with:
+          file: dockerfiles/Dockerfile-debian
+          context: .
--- a/dockerfiles/Dockerfile-debian
+++ b/dockerfiles/Dockerfile-debian
@@ -25,6 +25,7 @@ LABEL org.opencontainers.image.authors="dev@qpid.apache.org"
 
 # UPGRADE: proton > 0.37.0: replace `python3/dist-packages` with `python3.9/site-packages/`
 ARG proton_version=0.37.0
+# Enable web console: use 'ON' or 'OFF'
 ARG enable_console=OFF
 
 
@@ -32,7 +33,8 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends build-essential cmake pkg-config git ninja-build ca-certificates \
-        libssl-dev libsasl2-dev swig python3-dev libwebsockets-dev npm && \
+        libssl-dev libsasl2-dev swig python3-dev libwebsockets-dev \
+        $(if [ "${enable_console}" = "ON" ]; then echo "npm"; fi) && \
     rm -rf /var/lib/apt/lists/*
 
 # first, build Proton since there is no recent version packaged in Debian
--- a/dockerfiles/Dockerfile-debian
+++ b/dockerfiles/Dockerfile-debian
@@ -17,13 +17,14 @@
 # under the License.
 #
 
+# build with specifying root of repository as build context, e.g. `buildah -f dockerfiles/Dockerfile-debian .`
+
 FROM debian:bullseye-slim AS build
 
 LABEL org.opencontainers.image.authors="dev@qpid.apache.org"
 
 # UPGRADE: proton > 0.37.0: replace `python3/dist-packages` with `python3.9/site-packages/`
 ARG proton_version=0.37.0
-ARG dispatch_version=1.19.0
 ARG enable_console=OFF
 
 
@@ -34,8 +35,8 @@ RUN apt-get update && \
         libssl-dev libsasl2-dev swig python3-dev libwebsockets-dev npm && \
     rm -rf /var/lib/apt/lists/*
 
-RUN git clone --depth 1 -b ${dispatch_version} https://github.com/apache/qpid-dispatch.git && \
-	git clone --depth 1 -b ${proton_version} https://github.com/apache/qpid-proton.git
+# first, build Proton since there is no recent version packaged in Debian
+RUN git clone --depth 1 -b ${proton_version} https://github.com/apache/qpid-proton.git
 
 # build and install into system for laster usage by dispatch, but also install discretely so that it's easier to copy
 RUN mkdir /qpid-proton/build /qpid-proton/install-prefix && cd /qpid-proton/build && \
@@ -43,6 +44,8 @@ RUN mkdir /qpid-proton/build /qpid-proton/install-prefix && cd /qpid-proton/buil
 	cmake --build . --target install && \
 	cmake --install . --prefix /qpid-proton/install-prefix
 
+# build Dispatch Router
+ADD . /qpid-dispatch/
 RUN mkdir /qpid-dispatch/build /qpid-dispatch/install-prefix && cd /qpid-dispatch/build && \
 	cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=/usr -DUSE_VALGRIND=NO -DCONSOLE_INSTALL=${enable_console} && \
 	cmake --build . && \
--- a/dockerfiles/Dockerfile-debian
+++ b/dockerfiles/Dockerfile-debian
@@ -31,7 +31,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends build-essential cmake pkg-config git ninja-build ca-certificates \
-        libssl-dev libsasl2-dev libuv1-dev swig python3-dev libwebsockets-dev npm && \
+        libssl-dev libsasl2-dev swig python3-dev libwebsockets-dev npm && \
     rm -rf /var/lib/apt/lists/*
 
 RUN git clone --depth 1 -b ${dispatch_version} https://gitbox.apache.org/repos/asf/qpid-dispatch.git && \
@@ -39,7 +39,7 @@ RUN git clone --depth 1 -b ${dispatch_version} https://gitbox.apache.org/repos/a
 
 # build and install into system for laster usage by dispatch, but also install discretely so that it's easier to copy
 RUN mkdir /qpid-proton/build /qpid-proton/install-prefix && cd /qpid-proton/build && \
-	cmake .. -GNinja -DPROACTOR=libuv -DSYSINSTALL_BINDINGS=ON -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_PYTHON=ON && \
+	cmake .. -GNinja -DSYSINSTALL_BINDINGS=ON -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_PYTHON=ON && \
 	cmake --build . --target install && \
 	cmake --install . --prefix /qpid-proton/install-prefix
 
@@ -52,6 +52,11 @@ RUN mkdir /qpid-dispatch/build /qpid-dispatch/install-prefix && cd /qpid-dispatc
 #RUN cd /qpid-dispatch/build && ctest -VV
 
 FROM debian:bullseye-slim
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libsasl2-2 libsasl2-modules libwebsockets16 \
+        python3 libpython3.9 ca-certificates && \
+    rm -rf /var/lib/apt/lists/*
 
 # proton
 COPY --from=build /qpid-proton/install-prefix /usr/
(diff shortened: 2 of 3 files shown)


dockerfiles/Dockerfile-debian
```diff
@@ -0,0 +1,79 @@
...
+    apt-get install -y --no-install-recommends build-essential cmake pkg-config git ninja-build ca-certificates \
+        libssl-dev libsasl2-dev libuv1-dev swig python3-dev libwebsockets-dev npm && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN git clone --depth 1 -b ${dispatch_version} https://gitbox.apache.org/repos/asf/qpid-dispatch.git && \
+	git clone --depth 1 -b ${proton_version} https://gitbox.apache.org/repos/asf/qpid-proton.git
+
+# build and install into system for laster usage by dispatch, but also install discretely so that it's easier to copy
+RUN mkdir /qpid-proton/build /qpid-proton/install-prefix && cd /qpid-proton/build && \
+	cmake .. -GNinja -DPROACTOR=libuv -DSYSINSTALL_BINDINGS=ON -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_PYTHON=ON && \
```
jiridanek, line 42: Why libuv and not the default epoll?

dockerfiles/Dockerfile-debian
```diff
@@ -0,0 +1,79 @@
...
+# leftovers that are not so compliant with prefix, at least with how we manipulate it
+COPY --from=build /etc/qpid-dispatch /etc/qpid-dispatch
+COPY --from=build /usr/share/qpid-dispatch /usr/share/qpid-dispatch
+COPY --from=build /etc/sasl2 /etc/sasl2
+COPY --from=build /usr/lib/python3.9/site-packages/qpid_dispatch /usr/lib/python3.9/site-packages/qpid_dispatch
+COPY --from=build /usr/lib/python3.9/site-packages/qpid_dispatch_site.py /usr/lib/python3.9/site-packages/
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libsasl2-2 libsasl2-modules libuv1 python3 libpython3.9 libwebsockets16 && \
+    rm -rf /var/lib/apt/lists/*
```
jiridanek, line 74: Why is this written as `COPY` first, `RUN apt-get` afterwards. I thought that the opposite order is more reasonable.

dockerfiles/Dockerfile-debian
```diff
@@ -0,0 +1,79 @@
...
+	git clone --depth 1 -b ${proton_version} https://gitbox.apache.org/repos/asf/qpid-proton.git
+
+# build and install into system for laster usage by dispatch, but also install discretely so that it's easier to copy
+RUN mkdir /qpid-proton/build /qpid-proton/install-prefix && cd /qpid-proton/build && \
+	cmake .. -GNinja -DPROACTOR=libuv -DSYSINSTALL_BINDINGS=ON -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_PYTHON=ON && \
+	cmake --build . --target install && \
+	cmake --install . --prefix /qpid-proton/install-prefix
+
+RUN mkdir /qpid-dispatch/build /qpid-dispatch/install-prefix && cd /qpid-dispatch/build && \
+	cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=/usr -DUSE_VALGRIND=NO -DCONSOLE_INSTALL=${enable_console} && \
```
jiridanek, line 47: Is Ninja really such a big win? I do use it interactively, but mostly because it is fast on rebuilds. Using make with `-j` option for first build is IMO quite comparable in speed with ninja.

I'm fine with ninja in that it avoids having to mess with the `-j` make option.

dockerfiles/Dockerfile-debian
```diff
@@ -0,0 +1,79 @@
...
+COPY --from=build /etc/sasl2 /etc/sasl2
+COPY --from=build /usr/lib/python3.9/site-packages/qpid_dispatch /usr/lib/python3.9/site-packages/qpid_dispatch
+COPY --from=build /usr/lib/python3.9/site-packages/qpid_dispatch_site.py /usr/lib/python3.9/site-packages/
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libsasl2-2 libsasl2-modules libuv1 python3 libpython3.9 libwebsockets16 && \
+    rm -rf /var/lib/apt/lists/*
+
+# Add site-packages to the PYTHONPATH environment variable
+ENV PYTHONPATH=/usr/lib/python3.9/site-packages
```
jiridanek, line 77: Yeah, the `site-packages` bug is really annoying...

dockerfiles/Dockerfile-debian
```diff
@@ -0,0 +1,79 @@
...
+#
+
+FROM debian:bullseye-slim AS build
+
+LABEL org.opencontainers.image.authors="dev@qpid.apache.org"
+
+# UPGRADE: proton > 0.37.0: replace `python3/dist-packages` with `python3.9/site-packages/`
+ARG proton_version=0.37.0
+ARG dispatch_version=1.19.0
+ARG enable_console=OFF
```
jiridanek, line 27: :+1:

jiridanek (commented):
I like it, overall. I have a few comments inline.

dockerfiles/Dockerfile-debian
Jmennius: Hard to say, I am not opposed to either one - I just though this was the intention of other images since they installed it (but did not enable).

dockerfiles/Dockerfile-debian
Jmennius: Yes, that was exactly the reason to use it - to avoid hard coding or messing with `-j`

dockerfiles/Dockerfile-debian
Jmennius: It probably is, I will change it. (I will also include `ca-certificates` in the final image)

dockerfiles/Dockerfile-debian
Jmennius: yeah, but minor enough for me to ignore it for now :)

Jmennius:
In case you are willing to accept this - should I add build to CI?

jiridanek:
> In case you are willing to accept this - should I add build to CI?

Yes, that would be quite helpful. In fact, we should probably build all the dockerfiles. But building just this one will be a good start.

Jmennius:
> > In case you are willing to accept this - should I add build to CI?
> 
> Yes, that would be quite helpful. In fact, we should probably build all the dockerfiles. But building just this one will be a good start.

Ok, I'll do that a little bit later.
I have pushed a commit with requested changes (lets squash that later).

dockerfiles/Dockerfile-debian
```diff
@@ -0,0 +1,80 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends build-essential cmake pkg-config git ninja-build ca-certificates \
+        libssl-dev libsasl2-dev swig python3-dev libwebsockets-dev npm && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN git clone --depth 1 -b ${dispatch_version} https://gitbox.apache.org/repos/asf/qpid-dispatch.git && \
+	git clone --depth 1 -b ${proton_version} https://gitbox.apache.org/repos/asf/qpid-proton.git
```
gemmellr, line 38: I'd suggest using the GitHub repo, especially if going to build it in CI.

dockerfiles/Dockerfile-debian
Jmennius: I meant to ignore it but it's Monday and I was curious...
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=998739

jiridanek:
https://github.com/apache/qpid-dispatch/runs/5980336959?check_suite_focus=true#step:2:96

```
/usr/bin/docker buildx build --file dockerfiles/Dockerfile-debian --iidfile /tmp/docker-build-push-8nmE0m/iidfile --metadata-file /tmp/docker-build-push-8nmE0m/metadata-file .
error: could not find dockerfiles: stat dockerfiles: no such file or directory
Error: buildx failed with: error: could not find dockerfiles: stat dockerfiles: no such file or directory
```

The image build job looks broken

Jmennius:
> https://github.com/apache/qpid-dispatch/runs/5980336959?check_suite_focus=true#step:2:96
> 
> ```
> /usr/bin/docker buildx build --file dockerfiles/Dockerfile-debian --iidfile /tmp/docker-build-push-8nmE0m/iidfile --metadata-file /tmp/docker-build-push-8nmE0m/metadata-file .
> error: could not find dockerfiles: stat dockerfiles: no such file or directory
> Error: buildx failed with: error: could not find dockerfiles: stat dockerfiles: no such file or directory
> ```
> 
> The image build job looks broken

Let's try again... (needs approval)

dockerfiles/Dockerfile-debian
jiridanek: Ticket at https://issues.apache.org/jira/browse/DISPATCH-2337, to keep track of the defect.

Jmennius:
> > Let's try again... (needs approval)
> 
> It should be running without approvals on your GH fork, so you should already know if success is expected. I don't know why you don't have the GHA running at https://github.com/Jmennius/qpid-dispatch/actions.

That steps builds now :)

jiridanek:
@ganeshmurthy @kgiusti I think this looks ok, and it has the benefit of coming with a CI job. Furthermore the image is smaller than what can be achieved with CentOS bases (but ubi8 should now work to produce small image, if we either decide to live without fetching in gdb ;P or pull it from CentOS).

In general I'd like to delete some of the dockerfiles, maybe all except one, and leave the Debian one to be the only one Containerfile in `main` branch, WDYT?

Jmennius:
Few additional questions
1. should be web console be enabled by default?
2. we're fine with `RelWithDebInfo` (default) or should we switch to `Release`?
3. is it worth adding `$(if [ "${enable_console}" = "ON" ]; then echo "npm"; fi)` to apt-get invocation on build stage to avoid installing `npm` when it is not required during build?

jiridanek:
> 1. should be web console be enabled by default?

imo no

> 2. we're fine with `RelWithDebInfo` (default) or should we switch to `Release`?

RelWithDebInfo is used everywhere, nobody ever uses Release with dispatch, as far as I know

> 3. is it worth adding `$(if [ "${enable_console}" = "ON" ]; then echo "npm"; fi)` to apt-get invocation on build stage to avoid installing `npm` when it is not required during build? (does not influence runtime image)

maybe, yes. If you put it on its own line, then it makes sense to me; of course, it complicates build... up to you, I think

Jmennius:
> > 1. should be web console be enabled by default?
> 
> imo no
> 
> > 2. we're fine with `RelWithDebInfo` (default) or should we switch to `Release`?
> 
> RelWithDebInfo is used everywhere, nobody ever uses Release with dispatch, as far as I know
> 
> > 3. is it worth adding `$(if [ "${enable_console}" = "ON" ]; then echo "npm"; fi)` to apt-get invocation on build stage to avoid installing `npm` when it is not required during build? (does not influence runtime image)
> 
> maybe, yes. If you put it on its own line, then it makes sense to me; of course, it complicates build... up to you, I think

Thanks. If console stays disabled by default - I'll add that condition then.

jiridanek:
I've reported the failing CentOS 7 GHA job as issue https://issues.apache.org/jira/browse/PROTON-2534

.github/workflows/build.yaml
```diff
@@ -630,3 +630,15 @@ jobs:
...
+    name: Container Images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Build Debian image
+        uses: docker/build-push-action@v2
+        with:
+          file: dockerfiles/Dockerfile-debian
+          context: .
```
jiridanek, line 644: I don't insist on you adding it in this PR, but the dispatch project is missing a `.dockerignore` file. This means that people who have a `build` directory in their checkout and want to build the image will suffer the time penalty while docker copies the entire directory (with all logs from the test runs, ...) into the docker build context.

Good .dockerignore for dispatch can look something like this, https://github.com/skupperproject/skupper-router/blob/9017643e419af523236f643d54b0aca42f109e57/.dockerignore

Jmennius: Added, hope it's fine.

jiridanek:
@ganeshmurthy @kgiusti I'm going to merge this if there aren't any more comments. Thanks.

kgiusti (approved):
@ganeshmurthy @jiridanek looks good to me.  Feel free to commit it if you're both happy.

ganeshmurthy (approved):

PR #1560, opened 2022-04-21
DISPATCH-2339: Update react-scripts to 5; add CRACO to reapply polyfills

jiridanek:
Rhea ships with its sources in `lib/` as well as single file in `dist/`. Using the `lib/` sources with react-scripts 5 fails

```
$ npx react-scripts start
(node:836641) [DEP_WEBPACK_DEV_SERVER_ON_AFTER_SETUP_MIDDLEWARE] DeprecationWarning: 'onAfterSetupMiddleware' option is deprecated. Please use the 'setupMiddlewares' option.
(Use `node --trace-deprecation ...` to show where the warning was created)
(node:836641) [DEP_WEBPACK_DEV_SERVER_ON_BEFORE_SETUP_MIDDLEWARE] DeprecationWarning: 'onBeforeSetupMiddleware' option is deprecated. Please use the 'setupMiddlewares' option.
Starting the development server...
Failed to compile.

Module not found: Error: Can't resolve 'os' in '/home/jdanek/repos/qpid/qpid-dispatch/console/react/node_modules/rhea/lib'
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "os": require.resolve("os-browserify/browser") }'
        - install 'os-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "os": false }
ERROR in ./node_modules/rhea/lib/connection.js 36:9-22
Module not found: Error: Can't resolve 'os' in '/home/jdanek/repos/qpid/qpid-dispatch/console/react/node_modules/rhea/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "os": require.resolve("os-browserify/browser") }'
        - install 'os-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "os": false }

ERROR in ./node_modules/rhea/lib/connection.js 38:11-26
Module not found: Error: Can't resolve 'path' in '/home/jdanek/repos/qpid/qpid-dispatch/console/react/node_modules/rhea/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }

webpack compiled with 2 errors
```

The problem is that while rhea offers browserified bundled `dist/rhea.js` file already, but I could not figure out how to import it from the console's `connection.js` so that it passes through babel/webpack/whatever and it's exports are still visible.

"How do I browserify a file and then include it to be subsequently webpacked?"

I gave up on this and instead configured webpack to do the browserification the same way as it was with react-scripts 4.


jiridanek:
CC @bartoval Did you deal with this problem already?

console/react/package.json
```diff
@@ -58,6 +63,7 @@
     ]
   },
   "devDependencies": {
+    "@craco/craco": "^7.0.0-alpha.3",
```
jiridanek, line 66: Alternative is `react-app-rewired` package, which seems to do the same thing, with pretty much the same configuration format, and apparently does not force usage of alpha version to work with react-scripts 5. (https://blog.alchemy.com/blog/how-to-polyfill-node-core-modules-in-webpack-5)

bartoval: Hi @jiridanek,

yes, in ES6 modules the easiest way is using craco or react-app-rewired

bartoval: If you prefer to use the dist version of the files (dist/rhea.min, js, or rhea.js) you need to load them as a standalone library in the publc/index.html.

1) npm uninstall rhea
2) move the dist lib to the public folder of the project.


3) load the lib and attach it to the window object, so you can reuse it


4) use it

jiridanek:
I wanted to wait for rhea release that would incorporate valerio's fix that makes CRACO (or other such workaround) unnecessary. But maybe best not to wait.

bartoval:
@jiridanek the rhea libs is updated to the version 3.0.1. 
We can remove this modification and add ```import rhea from "rhea";``` in the connection.js file.

Let me know if you want to  update the code here or in case i can create a new PR?

jiridanek:
I looked into this some time ago. I had issues connecting the console to the router, which persisted regardless of versions of console and router that I used. Even older version that worked just fine in the past was unable to connect... I did not investigate in a greater detail then.

Currently I am on a sick leave, expected to last at least two or three weeks. I don't expect to find the time to look into this again this year.

jiridanek:
> I had issues connecting the console to the router, which persisted regardless of versions of console and router that I used. Even older version that worked just fine in the past was unable to connect... I did not investigate in a greater detail then.

The problem was too new version of Proton. With Proton 0.36 everything works fine.

jiridanek:
Obsoleted by the Rhea changes mentioned above.

PR #1567, opened 2022-05-31
Fixed #116: Coordinate port assignment between all parallel tests spawned by ctest -j (#123)

jiridanek:
* https://github.com/skupperproject/skupper-router/issue/116

### DIFF
--- a/tests/system_test.py
+++ b/tests/system_test.py
@@ -29,6 +29,7 @@ Features:
 """
 
 import errno
+import fcntl
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


kgiusti (approved):

tests/system_test.py
```diff
@@ -765,6 +772,46 @@ def setup(self):
...
+            This method always unlocks the lock file, so it should not ever deadlock other tests running in parallel.
+            Even if that happened, the lock is unlocked by the OS when the file is closed, which happens automatically
+            when the process that opened and locked it ends.
+
+            Invalid content in the self.port_file will break this method. Manual intervention is then required.
+        """
+        try:
+            fcntl.flock(self.port_file, fcntl.LOCK_EX)
+
+            # read old value
```
ganeshmurthy, line 792: Minor suggestion - 
Does `next_port = port + 1` need to be done if the file is empty. 
```
            # read old value
            self.port_file.seek(0, os.SEEK_END)
            if self.port_file.tell() != 0:
                # file is not empty, read the old port and increment it to get next possible available port.
                self.port_file.seek(0)
                port = int(self.port_file.read())
                next_port = port + 1
            else:
                # file is empty, so just start with a randint between 20000 and 30000
                port = random.randint(self.port_range[0], self.port_range[1])
```

jiridanek: i think it is needed; next_port has to be one higher than current port, and it does not matter if I get current port from file or as random number

jiridanek: In the beginning (no file), I return the result of random.randint and then store port + 1 to file. Next time, I read number from file and store number + 1 back to file and so on.

If I did not do port + 1 the first time, then I'd return the same port twice.

PR #1643, opened 2023-07-14
Update dependencies and refactoring the code

bartoval:
I have updated the dependencies apart from the ones related to D3.js (bloodbath mode on)

@jiridanek @ganeshmurthy  I have executed the basic application, but I believe it would be better if you could test this pull request to ensure that everything is functioning properly, like to the main branch. (ie login/logout)


jiridanek:
@dependabot rebase

jiridanek:
I've been checking the console pages by hand and they look allright, so let's merge this.

image

I had to use older qpid-proton, with main the console did not connect to the router (it was waiting for a callback from rhea that never came). The version of console before this PR has the same problem. This is where I got stuck trying to check the PR originally, I could not make it work on my machine. This time I thought to downgrade Proton to 0.36 and with that it works.

PR #1705, opened 2024-12-27
DISPATCH-2368: chore(build): add a missing `(void)` into C function declarations

jiridanek:
This is an intermediate step (in C23 this can be handled differently), but for now it's the most reasonable fix.

Previously already fixed in skupper-router
* https://github.com/skupperproject/skupper-router/pull/803

Fixes compilation failures such as this one

```
error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
```


jiridanek:
CI keeps failing the expected ways, so should be fine.

PR #1717, opened 2024-12-29
DISPATCH-2365: chore(gha): patch up the CentOS 7 job so that it runs again

jiridanek:
https://issues.apache.org/jira/browse/DISPATCH-2365


jiridanek:
The CI failures in the failing jobs match the failures before this change, so I think we're good.

PR #1736, opened 2025-01-13
DISPATCH-2368: chore(build): add missing `#include <sys/time.h>`

jiridanek:
Fixes compilation failures such as this one

```
error: call to undeclared function 'gettimeofday'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```


jiridanek:
Again, fails in the expected manner

PR #1737, opened 2025-01-13
DISPATCH-2371: chore(gha): enable IPv6 inside Docker container to run ip6 tests

jiridanek:
This is a port from skupper-router of
* https://github.com/skupperproject/skupper-router/pull/401


jiridanek:
And with this, there is only one (and consistent) test fail with proton main

> 	 54 - system_tests_edge_router (Timeout)

PR #1739, opened 2025-01-13
DISPATCH-2368: chore(build): fix qd_log formatting placeholder

jiridanek:
It's unexpected that compilers can trace the string as it flows to printf, as there are no hints in the code, but apparently they can do this nowadays.

```
error: format specifies type 'long long' but the argument has type 'long' [-Werror,-Wformat]
```

This was fixed in skupper-router in
* https://github.com/skupperproject/skupper-router/pull/1248


jiridanek:
Still failing in the expected way.

PR #1741, opened 2025-01-13
DISPATCH-2370: Configure "Packit-as-a-Service" application for github.com/apache/qpid-dispatch

jiridanek:


packaging/qpid-dispatch.spec
```diff
@@ -0,0 +1,515 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Copyright Fedora Project Authors.
+# This code is licensed under MIT license
```
gemmellr, line 19: The header and following statements seem in conflict.

Things licenced to the foundation and using the ASF specific header, dont have copyright notices included, and certainly not ones attributing it to another project.

I thought your plan was to write a new file, but if this is just a lift of a Fedora file then it could simply remain MIT and be referenced as such in the LICENCE file.

packaging/qpid-dispatch.spec
```diff
@@ -0,0 +1,515 @@
...
+%global _use_systemd 1
+
+%global proton_minimum_version 0.34.0
+%global libwebsockets_minimum_version 3.2.0
+%global libnghttp2_minimum_version 1.33.0
+
+%undefine __brp_mangle_shebangs
+
+Name:          qpid-dispatch
+Version:       3.1.0
```
gemmellr, line 43: Hmm?

packaging/qpid-dispatch.spec
```diff
@@ -0,0 +1,515 @@
...
+Release:       9.20250113093119135559.jd_fix_build.65.g76e0c22d%{?dist}
+Summary:       Dispatch router for Qpid
+License:       ASL 2.0
+URL:           http://qpid.apache.org/
+Source0:       qpid-dispatch-3.1.0.tar.gz
```
gemmellr, line 48: Hmm again. More below.

packaging/dispatch.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/etc/qdrouterd.conf b/etc/qdrouterd.conf
+index 4e8e506c..add40c69 100644
+--- a/etc/qdrouterd.conf
++++ b/etc/qdrouterd.conf
+@@ -32,6 +32,13 @@ listener {
+     saslMechanisms: ANONYMOUS
+ }
+
++listener {
```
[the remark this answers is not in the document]
jiridanek: This is how dispatch has been always packaged for Fedora (and also as a Red Hat product). No idea where this came from.

It's true that I removed console from the .spec file. If I can't find a way to build it as part of the spec, I'll revisit this, but if I get the console working, then I'd like to leave this as is.

Previously, the console build has been performed by the .spec file author (on their own machine), and the finished console build was uploaded into fedpkg look aside cache as a "source" file.

packaging/qpid-dispatch.spec
jiridanek: The idea is to develop the Fedora .spec in the packaged project itself, and it's continuously kept working as the project evolves. Therefore, when project releases new version, the Fedora package for it is ready to be synced into Fedora and packagers don't have to scramble to adjust to all the changes in the delta from previous packaged release.

This is ofc not relevant for dispatch, but my initial thought was that dispatch will be a proving ground and proton will be where there may be actual benefits.

jiridanek: If MIT can be kept throughout, that would be best. Will do it that way.

.packit.yaml
```diff
@@ -0,0 +1,40 @@
...
+#
+
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+# https://packit.dev/docs/configuration/#top-level-keys
+upstream_project_url: https://qpid.apache.org/components/dispatch-router
+issue_repository: https://issues.apache.org/jira/projects/DISPATCH/issues
+
+specfile_path: packaging/qpid-dispatch.spec
```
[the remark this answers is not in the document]
jiridanek: That was thought out in skupper-router. We wanted to appear vendor neutral. If Ubuntu rolled out it's own Packit-like service, we wanted to have inclusively-named directory where their stuff could also live.

Admit that it would've worked well if there was a single Fedora file, not when there is multiple like this.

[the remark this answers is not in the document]
gemmellr: Thats fine, if its thought all such things can coexist and be easily maintained in one dir. I was just thinking to add a `packaging/subdir` for each.

